PR #5, opened 2020-10-27
Card marking write barrier for generational mode.

bernd-aws:
Adding a typical generational card marking write barrier, which is only active in generational mode and has no other effect yet than writing into a card table that is allocated at startup on the side and covers the whole heap as specified by -Xmx.

The code for card marking is copied verbatim from existing shared barrier code. A previous version inherited this code, by making ShenandoahBarrierSet a subclass of CardTableBarrierSet. Now the code pertinent to post write barriers is just copied down instead. This allows adding the new code without any changes to shared code (code that is not specific to Shenandoah).

Initializing the card table in proper sequence comes with slightly rearranging heap initialization. This can be seen near the bottom of the listing here.

Testing: ran the same SPECjvm2008 benchmark programs as before. Performance seems to be 1-1.5% less with card marking active, in average for this kind of benchmark. In non-generational mode, no change observed.

For now we only prepare x86 code. Support for the other architectures will provided later with additional commits.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/shenandoah pull/5/head:pull/5`
`$ git checkout pull/5`

### DIFF
--- a/src/hotspot/cpu/x86/gc/shenandoah/c1/shenandoahBarrierSetC1_x86.cpp
+++ b/src/hotspot/cpu/x86/gc/shenandoah/c1/shenandoahBarrierSetC1_x86.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019, Red Hat, Inc. All rights reserved.
+ * Copyright (c) 2018, 2020 Red Hat, Inc. All rights reserved.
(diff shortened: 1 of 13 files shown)


rkennke (approved):
It looks good to me, and surprisingly straightforward! Nice!

rkennke (approved):

PR #7, opened 2020-11-17
Generation affiliation transitions for heap regions.

bernd-aws:
Preparatory renaming generation->affiliation since we will want to have a class called ShenandoahGeneration that is not an enum. The next PR will build on this. It will be a big refactoring that pulls code out of the ShenandoahHeap class and puts it into generation-specific classes that are subclasses of ShenandoahGeneration.

The Affiliation declaration still sits in AllocRequest and not in ShenandoahHeapRegion or ShenandoahGeneration for the same reason as before: it is hard to get anything to compile in other combinations of what goes where. I tried.

The card table update code in the set_affiliation() will be revised and augmented later, but seems OK as a first step.

---------
### Progress

### Testing

|     | Linux x32 | Linux x64 | Windows x64 | macOS x64 |
| --- | ----- | ----- | ----- | ----- |
| Build | ✔️ (1/1 passed) | ✔️ (5/5 passed) | ✔️ (2/2 passed) | ✔️ (2/2 passed) |
| Test (tier1) |    |  ✔️ (9/9 passed) | ❌ (1/9 failed) | ✔️ (9/9 passed) |

**Failed test task**
- [Windows x64 (hs/tier1 compiler)](https://github.com/bernd-aws/shenandoah/runs/1412798172)

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/shenandoah pull/7/head:pull/7`
`$ git checkout pull/7`

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahCardTable.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahCardTable.cpp
@@ -23,7 +23,6 @@
  */
 
 #include "precompiled.hpp"
-#include "gc/shared/memset_with_concurrent_readers.hpp"
 #include "gc/shenandoah/shenandoahCardTable.hpp"
 
 void ShenandoahCardTable::initialize() {
--- a/src/hotspot/share/gc/shenandoah/shenandoahCardTable.hpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahCardTable.hpp
@@ -38,7 +38,7 @@ public:
 
   virtual void initialize();
 
-  inline bool is_in_young(oop obj) const {
+  virtual inline bool is_in_young(oop obj) const {
     ShouldNotReachHere();
     return false;
   }
--- a/src/hotspot/share/gc/shenandoah/shenandoahCardTable.hpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahCardTable.hpp
@@ -38,7 +38,7 @@ public:
 
   virtual void initialize();
 
-  virtual inline bool is_in_young(oop obj) const {
+  inline bool is_in_young(oop obj) const {
     ShouldNotReachHere();
     return false;
   }
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
@@ -547,6 +547,8 @@ public:
 
   bool is_in(const void* p) const;
 
+  bool is_in_young(const void* p) const;
+
   MemRegion reserved_region() const { return _reserved; }
   bool is_in_reserved(const void* addr) const { return _reserved.contains(addr); }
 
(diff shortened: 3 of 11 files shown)


src/hotspot/share/gc/shenandoah/shenandoahCardTable.cpp
```diff
@@ -23,6 +23,7 @@
  */
 
 #include "precompiled.hpp"
+#include "gc/shared/memset_with_concurrent_readers.hpp"
```
rkennke, line 26: What's this?

src/hotspot/share/gc/shenandoah/shenandoahCardTable.hpp
```diff
@@ -38,7 +38,7 @@ class ShenandoahCardTable: public CardTable {
 
   virtual void initialize();
 
-  inline bool is_in_young(oop obj) const {
+  virtual inline bool is_in_young(oop obj) const {
```
rkennke, line 41: Why do we need virtual here? Is it overridden anywhere?

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -547,6 +547,8 @@ class ShenandoahHeap : public CollectedHeap {
 
   bool is_in(const void* p) const;
 
+  bool is_in_young(const void* p) const;
```
rkennke, line 550: This overrides CollectedHeap method, right? So either put 'virtual' here to indicate this, or even better 'override'. I believe we can use that new C++11 specifier now...

rkennke (changes requested):
Nice changes. A few questions too...

BTW, I'd like to bring in latest upstream changes so that we don't diverge too much. In particular, this will bring in concurrent weak reference processing. I suggest that I do this after this PR is integrated, ok?

src/hotspot/share/gc/shenandoah/shenandoahCardTable.cpp
bernd-aws: Not needed. Removing it.

(This is a remnant from a former version of mine where more functionality was hosted in this class. Tsktsk.)

src/hotspot/share/gc/shenandoah/shenandoahCardTable.hpp
bernd-aws: Removing "virtual".

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
bernd-aws: This method does not exist in CollectedHeap. It does exist in GenCollectedHeap, but ShenandoahHeap is not a subclass of the latter. (GenCollectedHeap is rather specialized, not suitable for us. It assumes contiguous generation spaces. G1 does not use it either.)

bernd-aws:
Please feel free to bring in upstream changes any time. Meanwhile, I'll prepare the next PR, but I will be happy to rebase it before landing.

rkennke (approved):
Looks good to me! Thanks!

PR #9, opened 2020-12-09
Marking and evacuation constrained to young gen in generational mode.

bernd-aws:
These restrictions will change when we have multiple generations at work. For GenShen milestone 1 we want to operate entirely in the young generation, to prove our changes function thus far.

The second commit adds the missing generation affiliation for contiguous allocations, i.e. humongous object regions.

This step still runs normally when in any non-generational mode, and it runs at reduced heap capacity in generational mode. Only the young generation is populated in the latter case.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/shenandoah pull/9/head:pull/9`
`$ git checkout pull/9`


rkennke (approved):
Looks good. Thanks!

PR #10, opened 2020-12-31
Scan remembered

kdnilsen:
These commits represent the beginning of our efforts to implement remembered set scanning.  The implementation has known performance and functional limitations.  The implementation is structured to support two different representations of remembered set.

The initial remembered set implementation uses traditional card marking, where mutator threads mark cards as dirty whenever a pointer is written to heap.

We are contemplating a future change to the remembered set implementation, whereby the existing SATB buffer is augmented to hold heap addresses overwritten in addition to, when enabled, reference value overwritten.  During the post-processing of SATB buffer by GC threads, the remembered set can be updated from this information.

---------
### Progress
- [ ] Change must be properly reviewed

### Error
&nbsp;⚠️ OCA signatory status must be verified

### Download
`$ git fetch https://git.openjdk.java.net/shenandoah pull/10/head:pull/10`
`$ git checkout pull/10`


kdnilsen:
This pull request was accidentally issued against the openjdk:master branch.  I will re-issue against openjdk:genshen.

Before issuing the new pull request, I intend to coalesce multiple commits in order to simplify the review.

PR #11, opened 2021-01-08
Scan remembered

kdnilsen:
Add support for scanning remembered set
    
The code has support for two alternative implementations of the remembered set.  The current remembered set implementation uses traditional card marking, where the post writer barrier for pointer write operations sets the mark for every overwritten card.
    
A contemplated future remembered set implementation is represented in skeleton form within the shenandoahBufferWithSATBRememberedSet.hpp and shenandoahBufferWithSATBRememberedSet.inline.hpp files.  The idea of this alternative remembered set implementation is that the existing SATB buffers will be augmented to additionally remember the address of each overwritten reference field.  Subsequent processing of the SATB buffer contents by background GC threads will update the TBD remembered set representation.
    
There are known bugs and performance improvements in the remembered set scanning implementation that have been addressed in certain Amazon-internal commits.  These commits will be upstreamed at a later time after other commits not directly related to remembered set scanning are upstreamed.

---------
### Progress
- [ ] Change must be properly reviewed

### Error
&nbsp;⚠️ OCA signatory status must be verified

### Download
`$ git fetch https://git.openjdk.java.net/shenandoah pull/11/head:pull/11`
`$ git checkout pull/11`

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahBufferWithSATBRememberedSet.hpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahBufferWithSATBRememberedSet.hpp
@@ -1,0 +1,103 @@
+//
+// The anticipated benefits of an alternative SATB-based implementation of the remembered set are several fold:
+//
+//  1. The representation of remembered set is more concise (1 bit per card) and there is no remembered set maintenance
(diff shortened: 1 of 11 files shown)


kdnilsen:
It looks like I'll have to fix up some white space errors and resubmit.  FWIW, these errors were not detected by "git jcheck".

kdnilsen:
I'll issue another with white space fixed.

PR #12, opened 2021-01-08
Scan remembered

kdnilsen:
Add support for scanning remembered set

The code has support for two alternative implementations of the remembered set. The current remembered set implementation uses traditional card marking, where the post writer barrier for pointer write operations sets the mark for every overwritten card.

A contemplated future remembered set implementation is represented in skeleton form within the shenandoahBufferWithSATBRememberedSet.hpp and shenandoahBufferWithSATBRememberedSet.inline.hpp files. The idea of this alternative remembered set implementation is that the existing SATB buffers will be augmented to additionally remember the address of each overwritten reference field. Subsequent processing of the SATB buffer contents by background GC threads will update the TBD remembered set representation.

There are known bugs and performance improvements in the remembered set scanning implementation that have been addressed in certain Amazon-internal commits. These commits will be upstreamed at a later time after other commits not directly related to remembered set scanning are upstreamed.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/shenandoah pull/12/head:pull/12`
`$ git checkout pull/12`


src/hotspot/share/gc/shenandoah/shenandoahBufferWithSATBRememberedSet.hpp
```diff
@@ -0,0 +1,103 @@
+//
```
rkennke, line 1: New file lacks copyright notice.

src/hotspot/share/gc/shenandoah/shenandoahBufferWithSATBRememberedSet.hpp
```diff
@@ -0,0 +1,103 @@
...
+//  2. While the mutator overhead is similar between the modified SATB barrier mechanism and direct card marking, the
+//     SATB mechanism offers the potential of improved accuracy within the remembered set.  This is because direct card
+//     marking unconditionally marks every old-gen page that is overwritten with a pointer value, even when the new pointer
+//     value might refer to old-gen memory.  With the modified SATB mechanism, background GC threads process the addresses
+//     logged within the SATB buffer and mark cards as dirty only if the pointer found at an overwritten old-gen address
+//     refers to young-gen memory.  There are (at least) two options here:
+//
+//      a) Just blindly dirty each card that is overwritten with a pointer value, regardless of the written value, as with
+//         the implementation of traditional direct card marking.  When this card's memory region is eventually scanned, the
+//         the implementation of remembered set scanning will clear the page if it no longer holds references to young-gen
```
rkennke, line 17: Minor: double- 'the'

src/hotspot/share/gc/shenandoah/shenandoahBufferWithSATBRememberedSet.hpp
```diff
@@ -0,0 +1,103 @@
...
+//  =====
+//    2 bytes (rounded down) out of 512 bytes is 0.39% bookkeeping overhead
+
+
+// ShenandoahBufferWithSATBRememberedSet is not implemented correctly in its current form.
+//
+// This class is a placeholder to support a possible future implementation of remembered set that uses a generalization of the
+// existing SATB pre-write barrier to maintain remembered set information rather than using unconditional direct card marking in
+// a post-write barrier.
+class ShenandoahBufferWithSATBRememberedSet: public CHeapObj<mtGC> {
```
rkennke, line 61: So, if I understand correctly, you intend to piggy-back the field-address onto the SATB queue, is that right? It would mean that we need to double the size of the items in the buffer. And it would also mean that SATB needs to be turned on all the time, not just during marking. Is that correct? If that is so, I wonder if it may be better to do like G1 does and use a separate queue and barrier for this? It looks to me that it is doing exactly what we need here.

src/hotspot/share/gc/shenandoah/shenandoahBufferWithSATBRememberedSet.inline.hpp
```diff
@@ -0,0 +1,52 @@
+
```
rkennke, line 1: Missing copyright header too.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -1146,6 +1156,7 @@ class ShenandoahEvacuateUpdateRootsTask : public AbstractGangTask {
     ShenandoahEvacOOMScope oom_evac_scope;
     ShenandoahEvacuateUpdateRootsClosure<> cl;
     MarkingCodeBlobClosure blobsCl(&cl, CodeBlobToOopClosure::FixRelocations);
+
```
rkennke, line 1159: Extra whitespace

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -2680,21 +2691,68 @@ class ShenandoahUpdateHeapRefsTask : public AbstractGangTask {
...
+        // Kelvin believes (but has not confirmed) the following:
+        //   r->top() represents the upper end of memory that has been allocated within this region.
+        //       As new objects are allocated, the value of r->top() increases to accomodate each new
+        //       object.
+        //   r->get_update_watermark() represents the value that was held in r->top() at the start of
+        //       evacuation.  During evacuation, new objects may be allocated within this heap region
+        //       and this will cause r->top() to increase.  But any objects allocated during the evacuation
+        //       phase do not need to be scanned by update-refs because the to-space invariant is in force
+        //       during evacuation and this will assure that any objects residing between
+        //       r->get_update_watermark() and r->top() hold no pointers to from-space.
```
rkennke, line 2730: We do need to scan objects that have been allocated by evacuation, because they may hold from-space refs. For this reason, we bump-up UWM when allocating for evac (in ShFreeSet, iirc).

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -2680,21 +2691,68 @@ class ShenandoahUpdateHeapRefsTask : public AbstractGangTask {
...
+      // concurrently.  This can be done during concurrent evacuation of roots for example.
+      if (r->is_active() && (r->affiliation() == ShenandoahRegionAffiliation::OLD_GENERATION) && !r->is_cset()) {
+
+        // Note that we use this code even if we are doing an old-gen collection and we have a bitmap to
+        // represent marked objects within the heap region.
+        //
+        // It is necessary to process all objects rather than just the marked objects during update-refs of
+        // an old-gen region as part of an old-gen collection.  Otherwise, a subseqent update-refs scan of
+        // the same region will see stale pointers and crash.
+        //
```
rkennke, line 2720: This is probably a bad idea. Unreachable objects are not guaranteed to have their Klass* intact, and may crash when accessed. Why would it crash during subsequent scans?

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -2680,21 +2691,68 @@ class ShenandoahUpdateHeapRefsTask : public AbstractGangTask {
...
+        HeapWord *p = r->bottom();
+        ShenandoahObjectToOopBoundedClosure<T> objs(&cl, p, update_watermark);
+
+        // TODO: This code assumes every object ever allocated within this old-gen region is still live.  If we
+        // allow a sweep phase to turn garbage objects into free memory regions, we need to modify this code to
+        // skip over and/or synchronize access to these free memory regions.  There might be races, for example,
+        // if we are trying to scan one of these free memory regions while a different thread is trying to
+        // allocate from within a free region.
+        //
+        // Reality is that this code is likely to be replaced with JVM-292 code before we ever get around to
```
rkennke, line 2741: What is 'JVM-292' code? ;-)

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -65,6 +68,7 @@ class ShenandoahVerifier;
 class ShenandoahWorkGang;
 class VMStructs;
 
+
```
rkennke, line 71: Extra WS

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -35,7 +35,9 @@
 #include "gc/shenandoah/shenandoahPadding.hpp"
 #include "gc/shenandoah/shenandoahSharedVariables.hpp"
 #include "gc/shenandoah/shenandoahUnload.hpp"
+#include "gc/shenandoah/shenandoahScanRemembered.hpp"
 #include "memory/metaspace.hpp"
+#include "gc/shenandoah/shenandoahScanRemembered.hpp"
```
rkennke, line 40: You're including the same header 2x

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -44,6 +46,7 @@ class ConcurrentGCTimer;
 class ObjectIterateScanRootClosure;
 class ShenandoahCollectorPolicy;
 class ShenandoahControlThread;
+class ShenandoahDirectCardMarkRememberedSet;
```
rkennke, line 49: This looks unused.

src/hotspot/share/gc/shenandoah/shenandoahReferenceProcessor.hpp
```diff
@@ -27,6 +27,7 @@
 #define SHARE_VM_GC_SHENANDOAH_SHENANDOAHREFERENCEPROCESSOR_HPP
 
 #include "gc/shared/referenceDiscoverer.hpp"
+#include "gc/shared/referencePolicy.hpp"
```
rkennke, line 30: This looks not needed or, if it is actually needed, not related to this change.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.cpp
```diff
@@ -0,0 +1,146 @@
+/*
+ * Copyright (c) Amazon.com, Inc. or its affiliates.  All rights reserved.
```
rkennke, line 2: You do have the copyright header here, good! :-) But it's lacking the year, like so:
 * Copyright (c) 2021, Amazon.com, Inc. or its affiliates. All rights reserved.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
```diff
@@ -0,0 +1,945 @@
+/*
+ * Copyright (c) Amazon.com, Inc. or its affiliates.  All rights reserved.
```
rkennke, line 2: Missing copyright year, as above.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.cpp
```diff
@@ -0,0 +1,146 @@
...
+  _cluster_count = _card_count /
+      ShenandoahCardCluster<ShenandoahBufferWithSATBRememberedSet>::CardsPerCluster;
+  _card_shift = CardTable::card_shift;
+
+  _whole_heap_base = _heap->base();
+  _whole_heap_end = _whole_heap_base + _card_count *
+      ShenandoahCardCluster<ShenandoahBufferWithSATBRememberedSet>::CardsPerCluster;
+}
+
+// Implementation is not correct.  ShenandoahBufferWithSATBRememberedSet is a placeholder for future planned improvements.
```
rkennke, line 103: This should go into its own .cpp file I think. Actually, I am not sure that we should add empty&unused code at all. Maybe add it when you also fill the implementation and use?

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
```diff
@@ -0,0 +1,945 @@
...
+//     Span of each card entry is 512 B
+//     Each card table entry consumes 1 bit
+//     Assume one long word of card tables represents a cluster.
+//       This long word holds 64 card table entries, spanning a
+//       total of 32 KB
+//     The number of clusters per region is 128 MB / 32 KB = 4K
+//
+
+// The typical parallel remembered set scanning effort consists of the
+// following steps, all of which are performed during a JVM safetpoint:
```
rkennke, line 90: Typo: safetpoint -> safepoint.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
```diff
@@ -0,0 +1,945 @@
...
+// To initiate concurrent scanning, insert the above code sequences
+// into strong_roots_do and roots_do methods of ShenandoahRootScanner
+// (source file shenandoahRootProcessor.cpp) following the
+// construction of tc_cl and rm and before the call to
+// _serial_roots.oops_do().
+//
+// To initiate concurrent evacuation, insert the above code sequences
+// into ShenandoahRootEvacuator::roots_do() before the invocation of
+// _serial_roots.oops_do(oops, worker_id).
+
```
rkennke, line 212: This is not quite clear to me. Above you said that all of the work is done during safepoints. Here you mention 'concurrent scanning' ... what is true?

src/hotspot/share/gc/shenandoah/shenandoahBufferWithSATBRememberedSet.hpp
rkennke, line 61: It's also lacking the usual #ifndef - #define - #endif sequence that is needed by headers.

src/hotspot/share/gc/shenandoah/shenandoahBufferWithSATBRememberedSet.inline.hpp
rkennke: And header #ifndef-#define-#endif ;-)

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
```diff
@@ -0,0 +1,945 @@
...
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ *
+ */
+
+// Terminology used within this source file:
```
rkennke, line 26: We usually place big comments like this below #includes and just above class declarations.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
```diff
@@ -0,0 +1,945 @@
...
+
+    // Do nothing for now as we have a brute-force implementation
+    // of findSpanningObject().
+
+#endif  // IMPLEMENT_THIS_OPTIMIZATION_LATER
+  }
+
+  // What card number corresponds to old-gen heap addresss p.  (If p
+  // does not refer to old-gen memory, the returned value is undefined.)
+  uint32_t cardNoForAddr(HeapWord *p);
```
rkennke, line 787: Should be named card_no_for_addr(HeapWord* p) to match style of most other Hotspot code.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
```diff
@@ -0,0 +1,401 @@
+/*
+ * Copyright (c) Amazon.com, Inc. or its affiliates.  All rights reserved.
```
rkennke, line 2: Missing copyright year.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.cpp
```diff
@@ -0,0 +1,146 @@
...
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ *
+ */
+
+
```
rkennke, line 26: Make sure you're adding #include "precompiled.hpp" and make sure it builds with and without --disable-precompiled-headers

rkennke (changes requested):
Nice work, thanks! I like that it explains so many details in comments. I've found a few issues, see below.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -72,6 +73,8 @@
 #include "gc/shenandoah/mode/shenandoahIUMode.hpp"
 #include "gc/shenandoah/mode/shenandoahPassiveMode.hpp"
 #include "gc/shenandoah/mode/shenandoahSATBMode.hpp"
+#include "gc/shenandoah/shenandoahScanRemembered.inline.hpp"
```
rkennke, line 76: I noticed now, that you also include shenandoahScanRemembered.hpp above. You only need to include the .inline.hpp, and bonus points for keeping it in alphabethical order... ;-)

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -2680,21 +2690,73 @@ class ShenandoahUpdateHeapRefsTask : public AbstractGangTask {
...
+        //     skipping over objects that were not live as of the most recently completed old-gen GC effort.
+
+        // Anything beyond update_watermark does not need to be updated.
+        while (p < update_watermark) {
+          oop obj = oop(p);
+
+          // The invocation of do_object() is "borrowed" from the implementation of
+          // ShenandoahHeap::marked_object_iterate(), which is called by _heap->marked_object_oop_iterate().
+          objs.do_object(obj);
+          p += obj->size();
```
rkennke, line 2755: As noted before, this might crash, if the Klass* that an unreachable object points to has been unloaded. The problem, as far as I can tell, arises because we don't have up-to-date marking information anymore, because we clean the whole bitmap, and subsequent young-gen collection doesn't establish old-gen reachability. Could we solve this by preserving the most-recent old-gen marking in the bitmap, and only clean the young-gen-parts? For now, it can probably be worked around by turning off ClassUnloading altogether.

src/hotspot/share/gc/shenandoah/shenandoahReferenceProcessor.hpp
```diff
@@ -26,7 +26,8 @@
 #ifndef SHARE_VM_GC_SHENANDOAH_SHENANDOAHREFERENCEPROCESSOR_HPP
 #define SHARE_VM_GC_SHENANDOAH_SHENANDOAHREFERENCEPROCESSOR_HPP
 
-#include "gc/shared/referenceDiscoverer.hpp"
+// #include "gc/shared/referenceDiscoverer.hpp"
+// #include "gc/shared/referencePolicy.hpp"
 #include "memory/allocation.hpp"
```
rkennke, line 31: Please fix that properly in an upstream change. I leave it up to you to keep the parts here, next merge should override it.

rkennke (commented):
The copyright stuff needs to be sorted out, but other than that it looks good to me now. Please notice my other few comments too. Good work!

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
```diff
@@ -312,7 +312,13 @@ inline oop ShenandoahHeap::evacuate_object(oop p, Thread* thread) {
...
-    // Increment age in young copies
+    // Hey!  This code showed up in a merge conflict.  It has "nothing" to do with the patch that
+    // was merged, so kdnilsen is leaving it in place as is.  However, it looks to me like the object's
+    // age should be incremented before the copy is committed to avoid the need for synchronization here.
+    //
+    // kdnilsen believes the following code is replaced/relocated in a subsequent commit.
+
+    // Increment age in young copies.
     if (target_gen == YOUNG_GENERATION) {
       copy_val->incr_age();
```
zhengyu123, line 323: Now, age is increased during young gen copy. Please update ShenandoahStringDedup::enqueue_candidate() to remove age updating code.

zhengyu123 (changes requested):

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
earthling-amzn: We've experimented with a preserved bitmap on a branch. It works until we hit a full collection, at which point things go sideways. The compaction phase specifically needed to be modified to maintain the correctness of the bitmap we wanted to preserve. It needs a bit more debugging, but it's been tabled for now.

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
kdnilsen: We did discover the need to deal with displaced mark words and a subsequent commit replaces the call to incr_age() with a call to a newly introduced incrase_object_age() function which includes special handling for displaced mark words.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
kdnilsen: Thanks.  This fix will be reflected in my next push.

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
kdnilsen: I have also removed the code in ShenandoahStringDedup::enqueue_candidate() which increments the age of strings since we are now incrementing age during evacuation.

src/hotspot/share/gc/shenandoah/shenandoahReferenceProcessor.hpp
kdnilsen: I will figure out how to upstream that change.  Thanks.

src/hotspot/share/gc/shenandoah/shenandoahBufferWithSATBRememberedSet.inline.hpp
kdnilsen: As discussed in email message, this file has been removed from the pull request.

src/hotspot/share/gc/shenandoah/shenandoahBufferWithSATBRememberedSet.hpp
kdnilsen: As discussed in email messages exchanged on shenandoah-dev@openjdk.java.net, there are potential synergies between the SATB-based remembered set implementation and the existing SATB pre-write barrier.  Especially if the default Shenandoah configuration chooses not to use SATB barrier, that increases the appeal of borrowing the G1 barrier implementation.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
kdnilsen: This has been fixed.

src/hotspot/share/gc/shenandoah/shenandoahBufferWithSATBRememberedSet.hpp
kdnilsen: This file has been removed from the pull request.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
kdnilsen: Removed.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
kdnilsen: We've asked around inside Amazon.  We're good to add in the copyright dates.  So next commit will include these.

src/hotspot/share/gc/shenandoah/shenandoahBufferWithSATBRememberedSet.hpp
rkennke: If we were to choose the I-U barrier as default, that would still be possible/interesting to piggy-back the card-marking on. The I-U barrier is very similar to the SATB-barrier: it is also employed at reference-stores, and it also uses a reference-queue, pretty much like SATB. The only difference (barrier-side) is that, instead of enqueuing the previous value of a field before the store, it enqueues the *new* value, about to be stored (or even already-stored when done as post-barrier, it doesn't really matter).

rkennke (approved):
Looks good to me now.

zhengyu123 (approved):

src/hotspot/share/gc/shenandoah/shenandoahBufferWithSATBRememberedSet.hpp
kdnilsen: Thanks for the email clarifying how the I-U barrier works.  I hadn't yet studied that.  I agree that we could also piggy-back card marking on that barrier.  I will explore further.

PR #13, opened 2021-01-21
Milestone 1

kdnilsen:
The objective of Milestone-1 is to demonstrate that certain heap regions can be designated as representing old-gen, card marking and remembered set implementation do not crash. Since the current implementation does not promote objects into old-gen space, much of the implementation of card marking and remembered set scanning is not exercised.

The following simple test program was used to demonstrate execution:

public class hello {
  public static void main(String args[]) {
    Node n = null;
    for (int count = 10000; count > 0; count--) {
      n = Node.upheaval(n);
      System.out.print(count);
      System.out.print(": Hello world: ");
      for (int i = 0; i < args.length; i++) {
	System.out.print(args[i]);
	System.out.print(" ");
      }
      System.out.print("[node value is " + n.value() + "]");
      System.out.println("");
    }
  }
}

import java.util.Random;

public class Node {
  static private final int NeighborCount = 5;
  static private Random random = new Random(46);

  private int val;
  private Object field_o;

  private int[] field_ints;
  private Node [] neighbors;

  // Copy each neighbor of n into new node's neighbor array.  Then overwrite
  // arbitrarily selected neighbor with newly allocated leaf node.
  public static Node upheaval(Node n) {
    int first_val = random.nextInt();
    if (first_val < 0) first_val = -first_val;
    if (first_val < 0) first_val = 0;
    Node result = new Node(first_val);
    if (n != null) {
      for (int i = 0; i < NeighborCount; i++)
	result.neighbors[i] = n.neighbors[i];
    }
    int second_val = random.nextInt();
    if (second_val < 0) second_val = -second_val;
    if (second_val < 0) second_val = 0;
    int overwrite_index = first_val % NeighborCount;
    result.neighbors[overwrite_index] = new Node(second_val);
    return result;
  }
  
  public Node(int val) {
    this.val = val;
    this.field_o = new Object();
    this.field_ints = new int[8];
    this.field_ints[0] = 0xca;
    this.field_ints[1] = 0xfe;
    this.field_ints[2] = 0xba;
    this.field_ints[3] = 0xbe;
    this.field_ints[4] = 0xba;
    this.field_ints[5] = 0xad;
    this.field_ints[6] = 0xba;
    this.field_ints[7] = 0xbe;
    this.neighbors = new Node[NeighborCount];
  }

  public int value() {
    return val;
  }
}

---------
### Progress

### Reviewers
 * [Aditya Mandaleeka](https://openjdk.java.net/census#adityam) (@adityamandaleeka - Author)

### Download
`$ git fetch https://git.openjdk.java.net/shenandoah pull/13/head:pull/13`
`$ git checkout pull/13`

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahCardTable.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahCardTable.cpp
@@ -23,6 +23,7 @@
  */
 
 #include "precompiled.hpp"
+#include "gc/shenandoah/shenandoahHeap.inline.hpp"
 #include "gc/shenandoah/shenandoahCardTable.hpp"
 
 void ShenandoahCardTable::initialize() {
--- a/src/hotspot/share/gc/shenandoah/shenandoahCardTable.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahCardTable.cpp
@@ -23,7 +23,6 @@
  */
 
 #include "precompiled.hpp"
-#include "gc/shenandoah/shenandoahHeap.inline.hpp"
 #include "gc/shenandoah/shenandoahCardTable.hpp"
 
 void ShenandoahCardTable::initialize() {
--- a/genshen-docs/README.md
+++ b/genshen-docs/README.md
@@ -1,0 +1,9 @@
+## GenShen: the Generational Shenandoah project
+
+Gen Shen stands for Generational Shenandoah,
+and it also means "deep roots" in Mandarin.
+
+Documents in this directory:
+- [glossary.md](glossary.md): glossary of terms used in our documents,
+- [workplan.summary.md](workplan.summary.md): summary of the plan of record that we strive to follow,
+- [design.summary.md](design.summary.md): summary of design points that we rely upon as working hypotheses.
--- a/genshen-docs/design.summary.md
+++ b/genshen-docs/design.summary.md
@@ -1,0 +1,329 @@
+## Design Points
+
+This section discusses design decisions, as guided by the following tenets.
+
+### Tenets
+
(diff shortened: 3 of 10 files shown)


src/hotspot/share/gc/shenandoah/shenandoahCardTable.cpp
```diff
@@ -23,6 +23,7 @@
  */
 
 #include "precompiled.hpp"
+#include "gc/shenandoah/shenandoahHeap.inline.hpp"
```
rkennke, line 26: What is this include used for?

rkennke (changes requested):
Very nice! Only one question regarding what looks like a rogue include. Also, maybe the testprogram can be readily included as first jtreg test that exercises genshen when running make test TEST=hotspot_gc_shenandoah (even when it probably doesn't test anything except that it doesn't crash)?

src/hotspot/share/gc/shenandoah/shenandoahCardTable.cpp
kdnilsen: This include will be removed in the next commit.

adityamandaleeka:
This looks good to me. I like the idea to include the test into hs_gc_shen as the first "real" jtreg test for genshen. In time, perhaps tier1 genshen tests (like "does card marking crash") should be in the Shenandoah tier1 tests, but for now I'm fine with it where you put it.

rkennke (approved):
Looks good to me!

adityamandaleeka (approved):
Sorry, I wrote my approval in a comment earlier but forgot to actually submit the review.

PR #16, opened 2021-02-17
Fix builds that do not have Shenandoah or other GCs

shipilev:
Lots of builds are currently failing with:

```
/home/buildbot/worker/build-shenandoah-jdkX-linux/build/src/hotspot/share/gc/shared/gcConfiguration.cpp: In member function 'GCName GCConfiguration::young_collector() const':
/home/buildbot/worker/build-shenandoah-jdkX-linux/build/src/hotspot/share/gc/shared/gcConfiguration.cpp:46:16: error: 'ShenandoahGCMode' was not declared in this scope
     if (strcmp(ShenandoahGCMode, "generational") == 0) {
                ^~~~~~~~~~~~~~~~
```

The new code in `gcConfiguration.cpp` should actually sense which configurations are available.

Additional testing:
 - [x] Linux PPC64 build (as one failing example), now passing
 - [x] Linux x86_64 `gc/shenandoah/generational` tests

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/shenandoah pull/16/head:pull/16`
`$ git checkout pull/16`

### DIFF
--- a/src/hotspot/share/gc/shared/gcConfiguration.cpp
+++ b/src/hotspot/share/gc/shared/gcConfiguration.cpp
@@ -35,60 +35,46 @@
 #include "utilities/macros.hpp"
 
 GCName GCConfiguration::young_collector() const {
-#if INCLUDE_G1GC
   if (UseG1GC) {
     return G1New;
   }
-#endif
 
-#if INCLUDE_PARALLELGC
   if (UseParallelGC) {
     return ParallelScavenge;
   }
-#endif
 
-#if INCLUDE_SHENANDOAHGC
   if (UseShenandoahGC) {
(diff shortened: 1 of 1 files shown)


rkennke (approved):
Looks good. I wonder if this should go upstream?

shipilev:
> Looks good. I wonder if this should go upstream?

Oh wait, this file is in upstream! So we only fail Shenandoah builds, because we touch ShenandoahGCMode that is not available. Other GCs expose at least some symbols to make this code work. Let me do a more pointed fix.

shipilev:
Please look again.

rkennke (approved):
Looks good!

PR #19, opened 2021-02-19
Allow young collection to suspend marking in old generation

earthling-amzn:
## Summary of changes
The goal of these changes is to allow young cycles to suspend marking in the old generation. When the young cycle is complete, the concurrent old marking will resume.
* Evaluation of heuristics was pulled into a new `ShenandoahRegulatorThread`. Taking this evaluation out of line from the control thread allows the regulator to suspend (using the cancellation mechanism) old generation marking and start a young generation cycle.
* Task queues for marking have been moved from `ShenandoahMarkingContext` into `ShenandoahGeneration`. This allows the marking state for the old generation to persist across young generation cycles. The associated `is_complete` state has also been moved to `ShenandoahGeneration`.
* Old generation marking is bootstrapped by a complete young generation cycle. In this scenario, the mark closures for a young cycle are given reference to the young generation mark queues _and_ the old generation mark queues. Rather than ignore old references as is done for a normal young cycle, they are enqueued in the old generation mark queues. When the young cycle completes, the old generation marking continues using the task queues primed by the preceding young cycle.
* There is a new flag: `ShenandoahAllowOldMarkingPreemption` (defaults to true). Disabling this option will cause the regulator to schedule  `young` or `global` collects (according to heuristics), but will _not_ schedule `old` collects.
* The `global` generation is used to support satb and incremental-update modes. The `global` generation is also used for degenerated and implicit/explicit gc requests. Degenerated cycles are not working on this branch and the root cause is understood.
* The `global` generation is also used for `FullGC`, but this is also broken. The `FullGC` doesn't update the remembered set during compaction. We reckon there is a non-trivial amount of work to fix this.
* The `MARKING` gc state has been split into `YOUNG_MARKING` and `OLD_MARKING`.
* Immediate garbage collection is broken in generational mode. The update references phase is used to repair the remembered set, so when this phase is skipped the remembered set scan runs into trouble. A fix for this is in progress.
* The remembered set is scanned on a safepoint during initial mark. Work to make this concurrent is in progress.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/shenandoah pull/19/head:pull/19`
`$ git checkout pull/19`


rkennke:
First priority for me is that hotspot_gc_shenandoah should be passing. I posted some fixes here: https://github.com/earthling-amzn/shenandoah/pull/1 - integrate them and they should show up in this PR.

There is one remaining test failing: make run-test TEST=gc/shenandoah/TestObjItrWithHeapDump.java
It looks like the new GC regulator thread is not scheduling a full GC on System.gc(). This probably needs some work.

In general, I am very skeptical about introducing *another* thread to schedule GC stuff. The coordination between Java threads and the control thread is already quite complicated (we had a bunch of deadlock bugs in the past), and introducing another one here seems to make things even worse.

I haven't looked into the details yet, but suspending old marking for a young cycle also seems quite complicated in terms of coordination. This would require a sort of safepointing mechanism for GC threads, or maybe it is possible to use the SuspendibleThreadSet stuff for that. Can you explain which approach you have taken there?

earthling-amzn:
I'll pull the fixes into this PR shortly, I want to first address your comments on adding a thread and coordinating young and old cycles.

The control thread is still in charge of scheduling. It still handles `System.gc` and allocation failures. The only thing in the new regulator thread is evaluation of heuristics. If the heuristics want to start a cycle, they make a request to the control thread. In a way, this makes the heuristics a peer of other threads that may start a gc. The motivation here is to allow the regulator to request  a young gc while the control thread is busy with an old gc.

The coordination between young and old gc cycles is very coarse and I think the existing cancellation mechanism is largely sufficient. Old generation gc is not allowed to cancel a young gc. An old generation gc begins by first running a young gc cycle with the mark closures configured to enqueue (but not follow) old objects into the old generation mark queues. This bootstrap young gc cycle runs to completion before any of the old gen task queues are serviced.

Once the young cycle has completed, the old generation runs the rest of the marking phase from the old gen task queues primed during the bootstrap young cycle. At this point in development, the old generation does not create a collection set (this is planned). If the regulator thread discovers that the young generation heuristic wants to start a gc cycle, it will schedule it with the control thread. The code path here is similar to that of an allocation failure. The regulator thread will request _cancellation_ of the concurrent old generation mark except it will _not_ clear the old generation task queues and it will not deactivate the SATB barrier. Once cancelled, the control thread runs a normal (i.e., non-bootstrapping) young cycle to completion before the control thread resumes the concurrent old marking.

zhengyu123 (changes requested):
Quick scan

rkennke:
It looks much better now, hotspot_gc_shenandoah is passing. I'll give it a more in-depth review later. I've tried to run some programs with generational mode, and it survives a few cycles, until it crashes with:

```
#  Internal Error (/home/rkennke/src/openjdk/shenandoah/src/hotspot/share/gc/shenandoah/shenandoahCollectionSet.cpp:101), pid=6682, tid=6688
#  assert(!_heap->get_region(index)->is_cset()) failed: should have been cleared before

Stack: [0x00007fd9529cf000,0x00007fd952acf000],  sp=0x00007fd952acd530,  free space=1017k
Native frames: (J=compiled Java code, A=aot compiled Java code, j=interpreted, Vv=VM code, C=native code)
V  [libjvm.so+0x1592374]  ShenandoahCollectionSet::clear()+0x1a4
V  [libjvm.so+0x15cb5fa]  ShenandoahGeneration::prepare_regions_and_collection_set(bool)+0x20a
V  [libjvm.so+0x15ac5bb]  ShenandoahDegenGC::op_degenerated()+0x24b
V  [libjvm.so+0x15ad4e0]  ShenandoahDegenGC::entry_degenerated()+0xf0
V  [libjvm.so+0x16e391c]  VM_ShenandoahDegeneratedGC::doit()+0x2c
V  [libjvm.so+0x19126b7]  VM_Operation::evaluate()+0x187
V  [libjvm.so+0x19372fc]  VMThread::inner_execute(VM_Operation*)+0x30c
V  [libjvm.so+0x1937c75]  VMThread::loop()+0x255
V  [libjvm.so+0x1937f1c]  VMThread::run()+0xcc
V  [libjvm.so+0x183baf8]  Thread::call_run()+0xf8
V  [libjvm.so+0x13b860e]  thread_native_entry(Thread*)+0x10e

```
See also attached hs_err.

I'll look into this a bit.

[hs_err_pid6682.log](https://github.com/openjdk/shenandoah/files/6027426/hs_err_pid6682.log)

rkennke:
Hmm, let's step back a little.
Ideally, we would maintain a working state at each milestone or even each commit. Obviously, this is not really possible during (early) development of a large feature like generational. But, as far as I can tell, the code works to some degree, with the exception of certain features (e.g. ref-processing, class-unloading, full-gc, etc). And we can also define what sort of features or behaviours *should* work. Would it be possible to captures this in test programs (ideally in the form of jtreg tests)? This has numerous advantages:
- Ability to continuously verify that what used to work still works after changes
- Documentation: For example, run configuration (e.g. flags) document how to run programs with relevant features disabled
- Testcases can be much narrower in scope than full programs
- We can use WhiteBox API to have better control of the JVM (we newly added GC breakpoint support to be able to 'step through' GC phases in test cases, which can be very useful, see e.g. JDK-8262122)

BTW, the flag to disable ref processing is -XX:-RegisterReferences (somewhat unintuitively)

So, looking at the current state, both 'hotspot_gc_shenandoah' and 'hotspot_gc_shenandoah_generational' are passing, which means we're good in theory. However, I'd think that you added new features and behaviours that go beyond what is currently tested in hotspot_gc_shenandoah_generational, so maybe add a few testcases there?

Also, I am not quite sure which of the not-working-quite-well aspects you intend to fix for this PR (maybe degen and full-GC?), and which you intend to defer to later milestones?

zhengyu123:
> It looks much better now, hotspot_gc_shenandoah is passing. I'll give it a more in-depth review later. I've tried to run some programs with generational mode, and it survives a few cycles, until it crashes with:
> 
> ```
> #  Internal Error (/home/rkennke/src/openjdk/shenandoah/src/hotspot/share/gc/shenandoah/shenandoahCollectionSet.cpp:101), pid=6682, tid=6688
> #  assert(!_heap->get_region(index)->is_cset()) failed: should have been cleared before
> 
> Stack: [0x00007fd9529cf000,0x00007fd952acf000],  sp=0x00007fd952acd530,  free space=1017k
> Native frames: (J=compiled Java code, A=aot compiled Java code, j=interpreted, Vv=VM code, C=native code)
> V  [libjvm.so+0x1592374]  ShenandoahCollectionSet::clear()+0x1a4
> V  [libjvm.so+0x15cb5fa]  ShenandoahGeneration::prepare_regions_and_collection_set(bool)+0x20a
> V  [libjvm.so+0x15ac5bb]  ShenandoahDegenGC::op_degenerated()+0x24b
> V  [libjvm.so+0x15ad4e0]  ShenandoahDegenGC::entry_degenerated()+0xf0
> V  [libjvm.so+0x16e391c]  VM_ShenandoahDegeneratedGC::doit()+0x2c
> V  [libjvm.so+0x19126b7]  VM_Operation::evaluate()+0x187
> V  [libjvm.so+0x19372fc]  VMThread::inner_execute(VM_Operation*)+0x30c
> V  [libjvm.so+0x1937c75]  VMThread::loop()+0x255
> V  [libjvm.so+0x1937f1c]  VMThread::run()+0xcc
> V  [libjvm.so+0x183baf8]  Thread::call_run()+0xf8
> V  [libjvm.so+0x13b860e]  thread_native_entry(Thread*)+0x10e
> ```
> 
> See also attached hs_err.
> 
> I'll look into this a bit.
> 
> [hs_err_pid6682.log](https://github.com/openjdk/shenandoah/files/6027426/hs_err_pid6682.log)

Looks like that degen into a wrong point, should be _degenerated_updaterefs, instead of _degenerated_outside_cycle.

Please consider adding generational GC info in event log:

http://cr.openjdk.java.net/~zgu/shenandoah/gen_event_log.patch


Thanks,

earthling-amzn:
I'm not sure what a generational specific jtreg test would look like. Would we extend the Whitebox API to tell us the age of objects or regions? I'd like to run the entire hotspot_gc_shenandoah test suite in generational mode while minimizing the number of flags required to maximize the number of passing test (for now, we need `-XX:-RegisterReferences` and `-XX:-UnloadClasses`). I'll figure out what that number of passing tests is today and that will be our proverbial line in the sand. From there we'll ratchet it up. Does that make sense?

Regarding this PR.

Known Issues:

* Full GC's are broken. We understand some of the reasons why (compaction does not update the remembered set). We do not plan to fix this until after the next milestone.
* The `ShenandoahImmediateThreshold` shortcut is disabled (it bypasses some remembered set maintenance during update references). We have a plan and some work in progress to fix this, but it hasn't formally been committed to a milestone yet.
* Degenerated cycles _should_ work. We know they're broken right now and we're working on it. This will be fixed in this PR.
 
Expected Behavior:

With `-XX:-ShenandoahAllowOldMarkingPreemption`:  
  * This _should_ be stable. In this configuration it will run either _young_ or _global_ cycles.
  * Class unloading only happens in global cycles.
  * Concurrent reference processing hasn't been tested in this configuration.

With: `-XX:+ShenandoahAllowOldMarkingPreemption`:
  * It _should_ run complete young cycles and finish marking of the old generation.
  * It _should_ be able to pause old generation marking and run a complete young cycle (and then resume old marking).
  * Class unloading and reference processing must be disabled (will do this when the mode is initialized).
  * Eventually the application will exhaust the heap because it doesn't actually reclaim old regions. 
 
Reclaiming old regions is planned for the next milestone. We will deliver this in smaller incremental changes. 

I will gratefully pull in @zhengyu123 's patch. We have deferred some of the operational improvements while we work on the core features. We expect the JMX beans will need some consideration and we need to add a timing phase for the remembered set scan (at the least).

I hope all that makes sense. Thanks for taking the time to read all that.

zhengyu123 (changes requested):

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
```diff
@@ -0,0 +1,245 @@
...
+  ShenandoahHeap* const heap = ShenandoahHeap::heap();
+  // TODO: Add a phase for rset scan.
+  // ShenandoahGCPhase phase(ShenandoahPhaseTimings::finish_mark);
+  uint nworkers = heap->workers()->active_workers();
+  reserve_task_queues(nworkers);
+
+  ShenandoahReferenceProcessor* rp = heap->ref_processor();
+  ShenandoahRegionIterator regions;
+  ShenandoahScanRememberedTask task(task_queues(), old_gen_task_queues(), rp, &regions);
+  heap->safepoint_workers()->run_task(&task);
```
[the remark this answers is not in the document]
earthling-amzn: Yep, that's wrong. We used to have this code in init mark, which was removed upstream. I fumbled the integration.

rkennke:
> I'm not sure what a generational specific jtreg test would look like. Would we extend the Whitebox API to tell us the age of objects or regions? I'd like to run the entire hotspot_gc_shenandoah test suite in generational mode while minimizing the number of flags required to maximize the number of passing test (for now, we need `-XX:-RegisterReferences` and `-XX:-UnloadClasses`). I'll figure out what that number of passing tests is today and that will be our proverbial line in the sand. From there we'll ratchet it up. Does that make sense?

Yes, that sounds sensible. It might even be too much. Pick a few sensible test programs that are known to work with a certain combination of flags, and add those to hotspot_gc_shenandoah_generational configuration. That should be good enough.

> Regarding this PR.
> 
> Known Issues:
> 
>     * Full GC's are broken. We understand some of the reasons why (compaction does not update the remembered set). We do not plan to fix this until after the next milestone.

Ok, good.

>     * The `ShenandoahImmediateThreshold` shortcut is disabled (it bypasses some remembered set maintenance during update references). We have a plan and some work in progress to fix this, but it hasn't formally been committed to a milestone yet.

Ok.

>     * Degenerated cycles _should_ work. We know they're broken right now and we're working on it. This will be fixed in this PR.

Alright. IIRC, you mentioned that degen is problematic when GC is cancelled during e.g. young-gen collection, and we basically need to start from 0 to get the full marking info, or something. If this is the case (I might be mis-remembering it), you could just as well jump into full GC. If, otoh, there is a reasonable way that STW GC could pick up a cancelled young or old GC, then you'll probably need to extend degen machinery to finish the generational young or old cycle. Does that make sense?

> Expected Behavior:
> 
> With `-XX:-ShenandoahAllowOldMarkingPreemption`:
> 
>     * This _should_ be stable. In this configuration it will run either _young_ or _global_ cycles.

Ok, so please test for that with a test program.

>     * Class unloading only happens in global cycles.

This seems reasonable.

>     * Concurrent reference processing hasn't been tested in this configuration.

Sure, disable it for now, and address it in a later milestone.

> With: `-XX:+ShenandoahAllowOldMarkingPreemption`:
> 
>     * It _should_ run complete young cycles and finish marking of the old generation.
> 
>     * It _should_ be able to pause old generation marking and run a complete young cycle (and then resume old marking).

Ok, can you make a test program, that tests with this flag? Not sure if you can test the exact behaviour that you describe, though.

>     * Class unloading and reference processing must be disabled (will do this when the mode is initialized).
> 
>     * Eventually the application will exhaust the heap because it doesn't actually reclaim old regions.

Ok, sure.

> Reclaiming old regions is planned for the next milestone. We will deliver this in smaller incremental changes.

Great!
 
> I will gratefully pull in @zhengyu123 's patch. We have deferred some of the operational improvements while we work on the core features. We expect the JMX beans will need some consideration and we need to add a timing phase for the remembered set scan (at the least).
> 
> I hope all that makes sense. Thanks for taking the time to read all that.

Thanks for explaining that much! That was exactly the information that I needed :-)

earthling-amzn:
We had some discussion on how to handle degenerated cycles. Having the degenerated cycle complete the generation in-progress would probably shorten the pause times, but running the degenerated cycle on the entire heap would reclaim the most memory. At any rate, it's not an irreversible decision and it's something to experiment with.

I'll add JTreg tests to exercise these new flags and create a new configuration to run the tests we expect to pass in generational mode.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
```diff
@@ -0,0 +1,244 @@
...
+  ShenandoahHeap* const heap = ShenandoahHeap::heap();
+  // TODO: Add a phase for rset scan.
+  // ShenandoahGCPhase phase(ShenandoahPhaseTimings::finish_mark);
+  uint nworkers = heap->safepoint_workers()->active_workers();
+  reserve_task_queues(nworkers);
+
+  ShenandoahReferenceProcessor* rp = heap->ref_processor();
+  ShenandoahRegionIterator regions;
+  ShenandoahScanRememberedTask task(task_queues(), old_gen_task_queues(), rp, &regions);
+  heap->safepoint_workers()->run_task(&task);
```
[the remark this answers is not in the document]
earthling-amzn: I figured `ParallelGCThreads` would be greater than `ConcGCThreads`, looking closer I see that the regular workers get `_max_workers`. I could put a `ShenandoahWorkerScope` here and have it get the number of workers from `ShenandoahWorkerPolicy`?. We plan to move this code off the safepoint anyway.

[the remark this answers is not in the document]
zhengyu123: Yep. In this case, should just use regular workers, then.

rkennke:
This looks much better! My tests now die when attempting full-GC, which is expected (and not going to be addressed in this PR). hotspot_gc_shenandoah also seems stable now. I'll give it an in-depth review next week.

rkennke:
You should enable GitHub actions in your fork under Settings->Actions->Allow all actions. Then you'll get presubmit-tests which runs a whole bunch of stuff on various platforms. A great sanity check.

earthling-amzn:
Yes, it's looking much more stable now. Hmm, I have actions enabled and I see it running builds on different platforms. I've been abusing this branch/pull request to fix windows and aarch64 builds.

rkennke:
It seems that 'RegisterReferences' is a develop flag and does not exist in release build, which means it cannot be changed by the generational mode (it results in crash). If reference processing really is a concern, then the way out would be to introduce our own flag, e.g. ShenandoahReferenceProcessing, and use that (and set RegisterReferences accordingly in non-release builds). OR, maybe even better, put a method should_process_references() in ShenandoahHeurisitics and override that to return false. OTOH, I just tried a bunch of SPECjvm workloads, and they seem to run fine without disabling reference processing (until they attempt a full-GC which is not yet supported).

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -436,16 +501,29 @@ ShenandoahHeapRegion* ShenandoahHeapRegion::humongous_start_region() const {
...
-  ShenandoahHeap::heap()->marking_context()->reset_top_at_mark_start(this);
+  heap->marking_context()->reset_top_at_mark_start(this);
   set_update_watermark(bottom());
 
   make_empty();
-  _affiliation = ShenandoahRegionAffiliation::FREE;
+#ifdef DEBUG_TRACE
+  printf("SHR::recycle(), setting region (%llx, %llx, %llx) to FREE\n",
+         (unsigned long long) bottom(), (unsigned long long) top(), (unsigned long long) end());
+  fflush(stdout);
```
rkennke, line 524: In Hotspot, it's simpler to use tty->print_cr("...") for this, or even better use logging e.g. log_trace(gc)("...") or so.

src/hotspot/share/gc/shenandoah/shenandoahMarkClosures.cpp
```diff
@@ -0,0 +1,54 @@
+#include "precompiled.hpp"
```
rkennke, line 1: It's lacking a copyright header here.

src/hotspot/share/gc/shenandoah/shenandoahMarkClosures.hpp
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (c) 2020, Red Hat, Inc. All rights reserved.
```
rkennke, line 2: And this copyright header has the wrong year in it :-)

src/hotspot/share/gc/shenandoah/shenandoahMarkClosures.hpp
```diff
@@ -0,0 +1,49 @@
...
+#ifndef SHARE_GC_SHENANDOAH_SHENANDOAHMARKCLOSURES_HPP
+#define SHARE_GC_SHENANDOAH_SHENANDOAHMARKCLOSURES_HPP
+
+#include "gc/shenandoah/shenandoahHeap.inline.hpp"
+#include "gc/shenandoah/shenandoahHeap.hpp"
+#include "gc/shenandoah/shenandoahHeapRegion.hpp"
+#include "gc/shenandoah/shenandoahSharedVariables.hpp"
+#include "gc/shenandoah/shenandoahMarkingContext.hpp"
+#include "gc/shenandoah/shenandoahMarkClosures.hpp"
+
```
rkennke, line 35: Instead of including everything, we can put forward declaration for any type that is only used as pointer in the header.

src/hotspot/share/gc/shenandoah/shenandoahOldGC.cpp
```diff
@@ -0,0 +1,59 @@
+#include "precompiled.hpp"
```
rkennke, line 1: Missing copyright header here.

src/hotspot/share/gc/shenandoah/shenandoahOldGC.hpp
```diff
@@ -0,0 +1,18 @@
+#ifndef SHARE_GC_SHENANDOAH_SHENANDOAHOLDGC_HPP
```
rkennke, line 1: And here too.

src/hotspot/share/gc/shenandoah/shenandoahOldGC.hpp
```diff
@@ -0,0 +1,18 @@
...
+#define SHARE_GC_SHENANDOAH_SHENANDOAHOLDGC_HPP
+
+#include "gc/shared/gcCause.hpp"
+#include "gc/shenandoah/shenandoahConcurrentMark.hpp"
+#include "gc/shenandoah/shenandoahConcurrentGC.hpp"
+#include "gc/shenandoah/shenandoahHeap.hpp"
```
rkennke, line 7: I see no use of ShenandoahHeap or ShenandoahConcurrentMark in this header. I believe those two includes could be removed.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
```diff
@@ -0,0 +1,30 @@
+#include "precompiled.hpp"
```
rkennke, line 1: Missing header.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.hpp
```diff
@@ -0,0 +1,22 @@
+
```
rkennke, line 1: Missing header

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.hpp
```diff
@@ -0,0 +1,22 @@
...
+#ifndef SHARE_VM_GC_SHENANDOAH_SHENANDOAHOLDGENERATION_HPP
+#define SHARE_VM_GC_SHENANDOAH_SHENANDOAHOLDGENERATION_HPP
+
+#include "gc/shenandoah/shenandoahGeneration.hpp"
+
```
rkennke, line 6: You should add a forward declaration of ShenandoahHeapRegion and ShenandoahHeapRegionClosure.

src/hotspot/share/gc/shenandoah/shenandoahPhaseTimings.hpp
```diff
@@ -99,8 +99,8 @@ class outputStream;
   f(conc_rendezvous_roots,                          "Rendezvous")                      \
   f(conc_evac,                                      "Concurrent Evacuation")           \
                                                                                        \
-  f(init_update_refs_gross,                         "Pause Init  Update Refs (G)")     \
-  f(init_update_refs,                               "Pause Init  Update Refs (N)")     \
+  f(init_update_refs_gross,                         "Pause Init Update Refs (G)")      \
+  f(init_update_refs,                               "Pause Init Update Refs (N)")      \
```
rkennke, line 103: Looks like unrelated spacing changes

src/hotspot/share/gc/shenandoah/shenandoahRegulatorThread.cpp
```diff
@@ -0,0 +1,158 @@
+/*
+ * Copyright (c), Amazon.com, Inc. and/or its affiliates. All rights reserved.
```
rkennke, line 2: You should add the copyright year (2021) here.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
```diff
@@ -23,14 +23,9 @@
...
-#ifndef SHARE_GC_SHENANDOAH_SHENANDOAHSCANREMEMBERED_HPP
-#define SHARE_GC_SHENANDOAH_SHENANDOAHSCANREMEMBERED_HPP
-
-#include <stdint.h>
-#include "memory/iterator.hpp"
-#include "gc/shenandoah/shenandoahCardTable.hpp"
-
+// During development of this new feature, we want the option to test
+// with and without, and to compare performance before and after.
+#define FAST_REMEMBERED_SET_SCANNING
```
rkennke, line 28: You'll still want the normal #ifndef SHARE_GC_SHENANDOAH_SHENANDOAHSCANREMEMBERED_HPP protection, don't you?

rkennke (changes requested):
That is a huge change! Good work!
First-pass review follows.

(I need to remember to scan for 'HEY!' in addition to 'TODO' ;-) )

src/hotspot/share/gc/shenandoah/mode/shenandoahGenerationalMode.cpp
```diff
@@ -42,28 +41,14 @@ void ShenandoahGenerationalMode::initialize_flags() const {
   SHENANDOAH_ERGO_ENABLE_FLAG(ExplicitGCInvokesConcurrent);
   SHENANDOAH_ERGO_ENABLE_FLAG(ShenandoahImplicitGCInvokesConcurrent);
 
+  // HEY! Disabled while M7 work is in progress.
+  FLAG_SET_ERGO(ShenandoahUnloadClassesFrequency, 0);
+  FLAG_SET_ERGO(RegisterReferences, false);
```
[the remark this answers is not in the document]
earthling-amzn: I also tried some runs with reference processing enabled and they succeeded. As I think about it, the reference processing itself will probably be fine with these changes. My concern would be that finalizers in the old generation might not run. I disabled it out of caution, but we can try leaving it enabled.

src/hotspot/cpu/aarch64/gc/shenandoah/c1/shenandoahBarrierSetC1_aarch64.cpp
```diff
@@ -29,6 +29,7 @@
 #include "gc/shenandoah/shenandoahBarrierSet.hpp"
 #include "gc/shenandoah/shenandoahBarrierSetAssembler.hpp"
 #include "gc/shenandoah/c1/shenandoahBarrierSetC1.hpp"
+#include "gc/shenandoah/mode/shenandoahMode.hpp"
```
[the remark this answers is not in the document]
earthling-amzn: Yes, removed.

src/hotspot/cpu/aarch64/gc/shenandoah/shenandoahBarrierSetAssembler_aarch64.cpp
```diff
@@ -60,7 +61,7 @@ void ShenandoahBarrierSetAssembler::arraycopy_prologue(MacroAssembler* masm, Dec
       if (ShenandoahSATBBarrier && dest_uninitialized) {
         __ tbz(rscratch1, ShenandoahHeap::HAS_FORWARDED_BITPOS, done);
       } else {
-        __ mov(rscratch2, ShenandoahHeap::HAS_FORWARDED | ShenandoahHeap::MARKING);
+        __ mov(rscratch2, ShenandoahHeap::HAS_FORWARDED | ShenandoahHeap::OLD_MARKING);
```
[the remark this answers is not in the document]
earthling-amzn: Good catch, was too focused on fixing the build.

src/hotspot/share/gc/shenandoah/mode/shenandoahMode.hpp
```diff
@@ -25,8 +25,10 @@
 #ifndef SHARE_GC_SHENANDOAH_MODE_SHENANDOAHMODE_HPP
 #define SHARE_GC_SHENANDOAH_MODE_SHENANDOAHMODE_HPP
 
+#include "runtime/java.hpp"
```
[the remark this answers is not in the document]
earthling-amzn: Not needed, removed.

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
```diff
@@ -182,9 +183,11 @@ inline bool ShenandoahHeap::check_cancelled_gc_and_yield(bool sts_active) {
   }
 }
 
-inline void ShenandoahHeap::clear_cancelled_gc() {
+inline void ShenandoahHeap::clear_cancelled_gc(bool clear_oom_handler) {
   _cancelled_gc.set(CANCELLABLE);
-  _oom_evac_handler.clear();
+  if (clear_oom_handler) {
+    _oom_evac_handler.clear();
```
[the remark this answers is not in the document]
earthling-amzn: Prior to these changes, the gc cancellation flag was always set because of an allocation failure, so the flag was cleared on a safepoint (final update refs). Now we have a case where we've "cancelled" a gc, but not because we've run out of memory so we don't need to clear the oom evac handler.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -323,6 +337,9 @@ size_t ShenandoahHeapRegion::get_gclab_allocs() const {
 void ShenandoahHeapRegion::set_live_data(size_t s) {
   assert(Thread::current()->is_VM_thread(), "by VM thread");
   _live_data = (s >> LogHeapWordSize);
+  if (_affiliation == FREE) {
+    //assert(_live_data == 0, "Setting non-zero live data (%zu) on FREE region", s);
+  }
```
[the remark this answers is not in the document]
earthling-amzn: Just left over debugging, removed.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
earthling-amzn: I agree, I've taken out all the debug scaffolding.

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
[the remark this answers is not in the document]
rkennke: Ok. But I think the OOM handler still needs to be cleared at the safepoint.

src/hotspot/share/gc/shenandoah/shenandoahOldGC.cpp
earthling-amzn: Fixed.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
earthling-amzn: Wow, I sure did miss a lot of copyright headers. Can `git jcheck` be made to check for this?

src/hotspot/share/gc/shenandoah/shenandoahPhaseTimings.hpp
earthling-amzn: Do you want me to revert this and put it in a separate PR? Maybe when we add a phase for remembered set scanning?

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
earthling-amzn: Guards were further down in the file, I pulled them up closer to where you'd expect them.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -391,23 +408,71 @@ void ShenandoahHeapRegion::print_on(outputStream* st) const {
...
     oop_iterate_humongous(blk);
   } else {
-    oop_iterate_objects(blk);
+    oop_iterate_objects(blk, fill_dead_objects, reregister_coalesced_objects);
   }
 }
 
-void ShenandoahHeapRegion::oop_iterate_objects(OopIterateClosure* blk) {
-  assert(! is_humongous(), "no humongous region here");
+void ShenandoahHeapRegion::oop_iterate_objects(OopIterateClosure* blk, bool fill_dead_objects, bool reregister_coalesced_objects) {
```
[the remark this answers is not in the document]
earthling-amzn: We have some work in progress (scheduled for completion these next two weeks) that will remove the need to fill dead objects. I'd rather clean this up as part of those changes.

src/hotspot/share/gc/shenandoah/shenandoahPhaseTimings.hpp
rkennke: You could propose it upstream. However, looking at it again, it may have been done on purpose, to line up "Init Update Refs" with "Final Update Refs". Not sure how reasonable that is, though. I'd revert that part of the change here, though.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
earthling-amzn: `grep -r -L -i copyright src/hotspot/share/gc/shenandoah` also works.

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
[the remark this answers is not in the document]
earthling-amzn: Ah yes, sorry. The default value for the boolean is true (to clear the oom handler). I've updated the other callers to pass the argument explicitly.

rkennke:
Do you plan to enable reference processing again on old-gen marking? Because I think it's a bad idea to not do it. Ideally, we'd do reference processing on both young-gen and old-gen markings, if that is not possible, then we should at least do it on old-gen markings (I believe other GCs don't do it in young-gen).

bernd-aws:
Yes, we plan to make reference processing function again and at a minimum this will hold for non-young collections.

rkennke:
Hey, you shout out when you think it's ready to go, yes? Because I think instead of piling more stuff into this PR, we should instead integrate it ASAP, and then progress with smaller PRs. Ok?

earthling-amzn:
Ha - was just about to email you. Yes, I think it's good to go. It passes all 71 of the tier1 Shenandoah tests.

rkennke (approved):
Looks good!

rkennke:
Can you please merge latest from shenandoah master? "This branch cannot be rebased due to conflicts" *sigh*

earthling-amzn:
Hmm, shenandoah:master is up to date (179c167a184 Fix Zero builds) and merged. If I try to `merge --squash` locally into shenandoah:master it goes through without any problems.

earthling-amzn:
I can rebase on the tip of shenandoah:master and (force) push that to this branch. Reckon that would help?

PR #21, opened 2021-03-11
Update periodic gc test for changed heuristic log message

earthling-amzn:
This also includes a change to hold down regulator requests when the GC is already running in non-generational modes.

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/shenandoah pull/21/head:pull/21`
`$ git checkout pull/21`


rkennke (approved):
Looks ok.
I was just thinking that GCMode kinda clashes with the ShenandoahGCMode (satb, iu, generational) distinction, and is thus a little confusing. It *did* confuse me for a bit :-)

earthling-amzn:
What about `GCPhase` or `GCStage`?

Another enumeration that warrants some discussion: `GenerationMode` - which is confusingly similar to the class name `ShenandoahGenerationalMode` (and the class `ShenandoahGeneration`). I think at the least it should have the Shenandoah prefix, or perhaps declare it inside `ShenandoahGenerationalMode` as just `Generation`?

PR #22, opened 2021-03-12
Only scan dirty cards, not entire regions, in update refs remset scan.

bernd-aws:
So far in update refs, when iterating over regions to process the remset, we scan an entire region for references into young gen if we find any dirty card in it. With this update, we only process dirty cards in each region and leave out clean ones. To do so, we refactor the scan code from the marking phase into a method that is then reused for update refs, passing down a different closure. The JVM switch `ShenandoahUseSimpleCardScanning` provides a means to fall back on the previous code.

(In the future, we may want to replace region stealing with card stealing.)

---------
### Progress

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/shenandoah pull/22/head:pull/22`
`$ git checkout pull/22`


rkennke (approved):
Looks good! Thanks!

PR #23, opened 2021-03-18
Fix ShenandoahPurgeSATBTask

zhengyu123:
ShenandoahPurgeSATBTask looks wrong.

1) Thread's token should be claimed by task, not by closure
2) Threads::threads_do() is single thread version.

Note: It can be improved by only iterating JavaThread with ShenandoahJavaThreadsIterator, but it needs phase time ...

---------
### Progress

### Reviewers
 * [Roman Kennke](https://openjdk.java.net/census#rkennke) (@rkennke - **Reviewer**) ⚠️ Review applies to b9c27ee8bd7ccbc989ad1c864ccbbd20e69079d8

### Download
To checkout this PR locally:
`$ git fetch https://git.openjdk.java.net/shenandoah pull/23/head:pull/23`
`$ git checkout pull/23`

To update a local copy of the PR:
`$ git checkout pull/23`
`$ git pull https://git.openjdk.java.net/shenandoah pull/23/head`

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
@@ -32,24 +32,21 @@
 #include "gc/shenandoah/shenandoahMark.inline.hpp"
 #include "gc/shenandoah/shenandoahOldGeneration.hpp"
 #include "gc/shenandoah/shenandoahOopClosures.inline.hpp"
+#include "gc/shenandoah/shenandoanRootProcessor.hpp"
 #include "gc/shenandoah/shenandoahStringDedup.hpp"
 #include "gc/shenandoah/shenandoahUtils.hpp"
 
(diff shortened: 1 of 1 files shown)


rkennke (approved):
Ugh. Right. Looks good!

zhengyu123:
> shenandoahConcurrentMark

Thanks for pointing out. I will fix it in jdk/jdk

rkennke:
Actually, now that I think about it a little more, I believe you should leave the claiming code where it was. It was one closure per task, and that should correctly claim the threads. Now you let all tasks see all the threads, which is wrong. Or I'm missing something.

zhengyu123:
> Actually, now that I think about it a little more, I believe you should leave the claiming code where it was. It was one closure per task, and that should correctly claim the threads. Now you let all tasks see all the threads, which is wrong. Or I'm missing something.

The threads are claimed inside Threads::possibly_parallel_threads_do()

PR #24, opened 2021-03-18
Restore ShenandoahVerify functionality for generational mode

earthling-amzn:
### Summary
1. Add remembered set to verifier roots for young generation collections
2. Don't follow pointers outside the generation during verify traversals
3. Verify generation's `used` value
4. Correct errors in computing a generation's `used` values

### Testing
1. Dacapo suite completes without verification errors
2. hotspot_gc_shenandoah completes without verification errors

---------
### Progress

### Reviewers
 * [Roman Kennke](https://openjdk.java.net/census#rkennke) (@rkennke - **Reviewer**) ⚠️ Review applies to b4d1ea782a46d96b955fec7e6ea698224d102dc3

### Download
To checkout this PR locally:
`$ git fetch https://git.openjdk.java.net/shenandoah pull/24/head:pull/24`
`$ git checkout pull/24`

To update a local copy of the PR:
`$ git checkout pull/24`
`$ git pull https://git.openjdk.java.net/shenandoah pull/24/head`

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
@@ -816,7 +816,7 @@ public:
   }
 };
 
-void ShenandoahHeapRegion::promote() {
+size_t ShenandoahHeapRegion::promote() {
   assert(SafepointSynchronize::is_at_safepoint(), "must be at a safepoint");
 
   ShenandoahHeap* heap = ShenandoahHeap::heap();
@@ -825,6 +825,9 @@ void ShenandoahHeapRegion::promote() {
   assert(affiliation() == YOUNG_GENERATION, "Only young regions can be promoted");
 
   UpdateCardValuesClosure update_card_values;
+  ShenandoahGeneration *old_generation = heap->old_generation();
+  ShenandoahGeneration *young_generation = heap->young_generation();
+
   if (is_humongous_start()) {
     oop obj = oop(bottom());
     assert(marking_context->is_marked(obj), "promoted humongous object should be alive");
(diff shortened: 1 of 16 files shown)


src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -825,6 +825,9 @@ void ShenandoahHeapRegion::promote() {
   assert(affiliation() == YOUNG_GENERATION, "Only young regions can be promoted");
 
   UpdateCardValuesClosure update_card_values;
+  ShenandoahGeneration *old_generation = heap->old_generation();
```
rkennke, line 828: We usually use Type* var instead of Type *var notation.

src/hotspot/share/gc/shenandoah/shenandoahRootVerifier.cpp
```diff
@@ -109,6 +110,12 @@ void ShenandoahRootVerifier::oops_do(OopClosure* oops) {
     ShenandoahStringDedup::oops_do_slow(oops);
   }
 
+  ShenandoahHeap *heap = ShenandoahHeap::heap();
```
rkennke, line 113: Same as above. Use ShenandoahHeap* heap. Occurs a few times in this file.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
```diff
@@ -573,4 +573,40 @@ ShenandoahScanRemembered<RememberedSet>::cluster_for_addr(HeapWordImpl **addr) {
...
+    _cl->do_oop(o);
+  }
+};
+
+template<typename RememberedSet>
+inline void ShenandoahScanRemembered<RememberedSet>::oops_do(OopClosure *cl) {
+  ShenandoahOopIterateAdapter adapter(cl);
+  ShenandoahHeap* heap = ShenandoahHeap::heap();
+  for (size_t i = 0, n = heap->num_regions(); i < n; ++i) {
+    ShenandoahHeapRegion* region = heap->get_region(i);
```
rkennke, line 596: And in this file too. Please check the whole change for this pattern ;-)

earthling-amzn:
@rkennke - to answer your question about testing. The generational mode has failures if the test causes a FullGC to run (this expected). It also looks like there may be an issue with `UseStringDeduplication`, but we haven't investigated this yet. Apart from those failures we pass ~200 out of 234 hotspot_gc_shenandoah tests.

rkennke:
> @rkennke - to answer your question about testing. The generational mode has failures if the test causes a FullGC to run (this is expected right now). It also looks like there may be an issue with `UseStringDeduplication`, but we haven't investigated this yet. Apart from those failures we pass ~200 out of 234 hotspot_gc_shenandoah tests.

Very nice! Maybe it would be worth to add run configurations to the tests that are expected to pass under hotspot_gc_shenandoah_generational?

PR #25, opened 2021-03-23
Refix ShenandoahPurgeSATBTask

zhengyu123:
My early fix was incomplete and incorrect.

1) ShenandoahPurgeSATBTask needs to change thread claim token for iterating threads in parallel, this was missing in early change.
2) Threads::threads_do() -> Threads::possibly_parallel_threads_do() was incorrect, the later version only iterates Java and VM threads. This part of changes are reverted.
3) A minor change: declaring _trashed_oops as volatile as it is a shared variable.

---------
### Progress

### Reviewers
 * @earthling-amzn (no known github.com user name / role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/25/head:pull/25` \
`$ git checkout pull/25`

`$ git checkout pull/25` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/25/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 25`

`$ git pr show -t 25`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/25.diff">https://git.openjdk.java.net/shenandoah/pull/25.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
@@ -38,14 +38,18 @@
 class ShenandoahFlushAllSATB : public ThreadClosure {
  private:
   SATBMarkQueueSet& _satb_qset;
+  uintx _claim_token;
 
  public:
   explicit ShenandoahFlushAllSATB(SATBMarkQueueSet& satb_qset) :
-    _satb_qset(satb_qset) { }
+    _satb_qset(satb_qset),
+    _claim_token(Threads::thread_claim_token()) { }
 
   void do_thread(Thread* thread) {
-    // Transfer any partial buffer to the qset for completed buffer processing.
-    _satb_qset.flush_queue(ShenandoahThreadLocalData::satb_mark_queue(thread));
+    if (thread->claim_threads_do(true, _claim_token)) {
+      // Transfer any partial buffer to the qset for completed buffer processing.
+      _satb_qset.flush_queue(ShenandoahThreadLocalData::satb_mark_queue(thread));
+    }
   }
 };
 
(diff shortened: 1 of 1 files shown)


earthling-amzn (approved):
Thanks again for setting this straight.

zhengyu123:
> Thanks again for setting this straight.

@earthling-amzn Thanks for reviewing and sorry for the noise.

earthling-amzn:
Can we move this PR forward?

zhengyu123:
> Can we move this PR forward?

Roman is on PTO this week, I will ping him next week.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
```diff
@@ -92,12 +96,14 @@ class ShenandoahPurgeSATBTask : public AbstractGangTask {
   ShenandoahObjToScanQueueSet* _mark_queues;
 
 public:
-  size_t _trashed_oops;
+  volatile size_t _trashed_oops;
```
rkennke, line 99: If it the intention is to make shared access to the field safe, then C++ volatile does not (fully) achieve this. You also need to use Atomic::inc() and similar to change/read it.

rkennke (approved):
Looks good, only a minor thing about shared access to _trashed_oops.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
zhengyu123: Adding volatile keyword, mainly just for coding convention on shared variable. It uses Atomic::add() to update variable, so it is correct.

PR #26, opened 2021-03-31
Clean up CondCardMark code: remove excess barriers, enable it by default

shipilev:
The `StoreLoad`, `StoreStore` barriers in that code are only needed for CMS, and does not relate to Shenandoah. Also, `UseCondCardMark` is a recommended option for most machines today, and therefore we better just enable it ergonomically.

Additional testing:
 - [x] hotspot_gc_shenandoah, x86_64, {fastdebug,release}

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/26/head:pull/26` \
`$ git checkout pull/26`

`$ git checkout pull/26` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/26/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 26`

`$ git pr show -t 26`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/26.diff">https://git.openjdk.java.net/shenandoah/pull/26.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/gc/shenandoah/shenandoahBarrierSetAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/gc/shenandoah/shenandoahBarrierSetAssembler_aarch64.cpp
@@ -643,9 +643,6 @@ void ShenandoahBarrierSetAssembler::gen_write_ref_array_post_barrier(MacroAssemb
 
   __ load_byte_map_base(scratch);
   __ add(start, start, scratch);
-  // if (ct->scanned_concurrently()) {
-  //   __ membar(__ StoreStore);
-  // }
   __ bind(L_loop);
   __ strb(zr, Address(start, count));
   __ subs(count, count, 1);
--- a/src/hotspot/cpu/aarch64/gc/shenandoah/shenandoahBarrierSetAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/gc/shenandoah/shenandoahBarrierSetAssembler_aarch64.cpp
@@ -390,15 +390,11 @@ void ShenandoahBarrierSetAssembler::store_check(MacroAssembler* masm, Register o
 
   if (UseCondCardMark) {
     Label L_already_dirty;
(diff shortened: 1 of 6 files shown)


src/hotspot/cpu/aarch64/gc/shenandoah/shenandoahBarrierSetAssembler_aarch64.cpp
```diff
@@ -390,15 +390,11 @@ void ShenandoahBarrierSetAssembler::store_check(MacroAssembler* masm, Register o
 
   if (UseCondCardMark) {
     Label L_already_dirty;
-    __ membar(Assembler::StoreLoad);
```
zhengyu123, line 393: I believe this StoreStore barrier is for preventing race against concurrent refinement. Question is that, generational Shenandoah going to implement concurrent refinement in the future?

earthling-amzn: We plan to start work on concurrent remembered set scanning very soon. I'm not far enough into it yet to know if we'll need these barriers, but it looks like we might.

zhengyu123 (approved):
Looks goodf

src/hotspot/cpu/aarch64/gc/shenandoah/shenandoahBarrierSetAssembler_aarch64.cpp
shipilev: Either way, this thing is gone with https://bugs.openjdk.java.net/browse/JDK-8261309 -- and it was needed for CMS. I would prefer to drop it to match upstream, and then reintroduce the barriers iff they are needed.

earthling-amzn: I agree, thanks for the link and the background on the change.

PR #28, opened 2021-04-06
Use timed wait to sleep control thread

earthling-amzn:
Using a timed `wait` rather than a naked sleep allows the control thread to be more responsive to requests from mutators and the regulator thread to start GC cycles. The sleep time is also changed from `ShenandoahControlIntervalMin` to `ShenandoahControlIntervalMax` to reduce unnecessary polling cycles. We could use a plain `wait`, but the control thread is responsible for periodically sending allocation metrics to the `pacer`.

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
@@ -52,6 +52,7 @@ ShenandoahControlThread::ShenandoahControlThread() :
   ConcurrentGCThread(),
   _alloc_failure_waiters_lock(Mutex::leaf, "ShenandoahAllocFailureGC_lock", true, Monitor::_safepoint_check_always),
   _gc_waiters_lock(Mutex::leaf, "ShenandoahRequestedGC_lock", true, Monitor::_safepoint_check_always),
+  _control_lock(Mutex::leaf, "ShenandoahControlGC_lock", true, Monitor::_safepoint_check_never),
   _periodic_task(this),
   _requested_gc_cause(GCCause::_no_cause_specified),
   _requested_generation(GenerationMode::GLOBAL),
@@ -367,11 +368,12 @@ void ShenandoahControlThread::run_service() {
       last_shrink_time = current;
     }
 
-    // HEY! kemperw would like to have this thread sleep on a timed wait so it
-    // could be explicitly woken when there is something to do. The timed wait
-    // is necessary because this thread has a responsibility to send
-    // 'alloc_words' to the pacer when it does not perform a GC.
-    os::naked_short_sleep(ShenandoahControlIntervalMin);
+    {
+      // The timed wait is necessary because this thread has a responsibility to send
+      // 'alloc_words' to the pacer when it does not perform a GC.
+      MonitorLocker lock(&_control_lock, Mutex::SafepointCheckFlag::_no_safepoint_check_flag);
+      lock.wait(ShenandoahControlIntervalMax);
+    }
   }
 
   // Wait for the actual stop(), can't leave run_service() earlier.
@@ -721,6 +723,7 @@ bool ShenandoahControlThread::request_concurrent_gc(GenerationMode generation) {
   if (_mode == none) {
     _requested_gc_cause = GCCause::_shenandoah_concurrent_gc;
     _requested_generation = generation;
+    notify_control_thread();
     return true;
   }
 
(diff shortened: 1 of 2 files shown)


rkennke:
This looks like something that @shipilev has touched a few times before.

shipilev (approved):
I don't mind this going into sandbox, but honestly, we need to fix it in the mainline. The awkwardness of control loop sleep-looping gets very tedious.

shipilev:
I shall try and do this in mainline with this: https://bugs.openjdk.java.net/browse/JDK-8264851 -- probably next week, though.

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
```diff
@@ -367,11 +368,12 @@ void ShenandoahControlThread::run_service() {
...
 
-    // HEY! kemperw would like to have this thread sleep on a timed wait so it
-    // could be explicitly woken when there is something to do. The timed wait
-    // is necessary because this thread has a responsibility to send
-    // 'alloc_words' to the pacer when it does not perform a GC.
-    os::naked_short_sleep(ShenandoahControlIntervalMin);
+    {
+      // The timed wait is necessary because this thread has a responsibility to send
+      // 'alloc_words' to the pacer when it does not perform a GC.
+      MonitorLocker lock(&_control_lock, Mutex::SafepointCheckFlag::_no_safepoint_check_flag);
```
shipilev, line 374: This is just `Mutex::_no_safepoint_check_flag`, Hotspot uses that everywhere. Another instance in also in the same file.

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
```diff
@@ -756,6 +765,7 @@ void ShenandoahControlThread::handle_requested_gc(GCCause::Cause cause) {
   while (current_gc_id < required_gc_id) {
     _gc_requested.set();
     _requested_gc_cause = cause;
+    notify_control_thread();
```
shipilev, line 768: So this thing holds `_gc_waiters_lock`. I noticed it when trying to implement JDK-8264851. Fastdebug builds would fail with: `#  assert(false) failed: Attempting to acquire lock ShenandoahControlWaitGC_lock/11 out of order with lock ShenandoahRequestedGC_lock/11 -- possible deadlock`. How is that handled here?

shipilev (commented):
Oh wait, questions...

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
earthling-amzn: Sorry, my IDE does this. Fixed.

earthling-amzn:
I'm not sure we can fix this so easily on mainline. The control thread on mainline needs to periodically evaluate the heuristics. To support the generational changes, heuristic evaluation was moved to a separate thread (so that it can initiate a young collection while the control thread is busy handling an old collection).

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
earthling-amzn: Looking into this.

earthling-amzn: I think this is a false positive. There's no deadlock here because the control thread never tries to hold both these locks at the same time. I changed the rank of the `_gc_waiters_lock` to allow a thread to hold them both.

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
```diff
@@ -51,7 +51,7 @@
 ShenandoahControlThread::ShenandoahControlThread() :
   ConcurrentGCThread(),
   _alloc_failure_waiters_lock(Mutex::leaf, "ShenandoahAllocFailureGC_lock", true, Monitor::_safepoint_check_always),
-  _gc_waiters_lock(Mutex::leaf, "ShenandoahRequestedGC_lock", true, Monitor::_safepoint_check_always),
+  _gc_waiters_lock(Mutex::safepoint, "ShenandoahRequestedGC_lock", true, Monitor::_safepoint_check_always),
```
shipilev, line 54: Look at `src/hotspot/share/runtime/mutexLocker.cpp`: if we need to elevate the lock above/below the `leaf`, the usual thing to do is to say `leaf + 1` or `leaf - 1`. I would prefer to make `_control_lock` into `leaf - 1`, as that one is special, not all other locks.

shipilev (approved):
OK then, with one minor stylistic nit.

shipilev:
Good to go.

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
earthling-amzn: Yes, good call.

shipilev:
OpenJDK infra seems to be at maintenance. Bots would hopefully integrate as infra is online again.

earthling-amzn:
> OpenJDK infra seems to be at maintenance. Bots would hopefully integrate as infra is online again.

Yay bots!

PR #29, opened 2021-04-07
Mixed evacuation

kdnilsen:
This code adds to generational Shenandoah the ability to perform concurrent garbage collection of young-gen and old-gen memory. Following completion of an old-gen concurrent marking effort, we select certain old-gen heap regions to serve as candidates for future collection sets. All dead objects within the old-gen heap regions that are not part of this candidate set are coalesced and filled so that remembered-set scanning of these old-gen heap regions will not be confused by "zombie objects" (objects that old-gen has decided are dead, which reside in regions that have not yet been collected). After concurrently coalescing and filling these dead objects, each subsequent young-gen evacuation pass includes a subset of the old-gen candidates until all candidates have been collected. This code passes TIER1 and hotspot-gc-shenandoah jtreg tests without regressions. A new jtreg test has been added to exercise concurrent old/young GC.

---------
### Progress

### Reviewers
 * [Roman Kennke](https://openjdk.java.net/census#rkennke) (@rkennke - **Reviewer**) ⚠️ Review applies to d6fa428bef00403ab88c92c6d9bac9ab093b5944

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/29/head:pull/29` \
`$ git checkout pull/29`

`$ git checkout pull/29` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/29/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 29`

`$ git pr show -t 29`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/29.diff">https://git.openjdk.java.net/shenandoah/pull/29.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
@@ -54,8 +54,8 @@ const double ShenandoahAdaptiveHeuristics::HIGHEST_EXPECTED_AVAILABLE_AT_END = 0
 const double ShenandoahAdaptiveHeuristics::MINIMUM_CONFIDENCE = 0.319; // 25%
 const double ShenandoahAdaptiveHeuristics::MAXIMUM_CONFIDENCE = 3.291; // 99.9%
 
-ShenandoahAdaptiveHeuristics::ShenandoahAdaptiveHeuristics(ShenandoahGeneration* generation) :
-  ShenandoahHeuristics(generation),
+ShenandoahAdaptiveHeuristics::ShenandoahAdaptiveHeuristics(ShenandoahGeneration* generation, ShenandoahHeuristics* old_heuristics) :
+  ShenandoahHeuristics(generation, old_heuristics),
   _margin_of_error_sd(ShenandoahAdaptiveInitialConfidence),
   _spike_threshold_sd(ShenandoahAdaptiveInitialSpikeThreshold),
   _last_trigger(OTHER) { }
@@ -95,6 +95,7 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
                      byte_size_in_proper_unit(actual_free), proper_unit_for_byte_size(actual_free),
                      byte_size_in_proper_unit(max_cset),    proper_unit_for_byte_size(max_cset),
                      byte_size_in_proper_unit(min_garbage), proper_unit_for_byte_size(min_garbage));
+
 
   // Better select garbage-first regions
   QuickSort::sort<RegionData>(data, (int)size, compare_by_garbage, false);
@@ -200,6 +201,10 @@ bool ShenandoahAdaptiveHeuristics::should_start_gc() {
   size_t capacity = _generation->soft_max_capacity();
   size_t available = _generation->available();
   size_t allocated = _generation->bytes_allocated_since_gc_start();
+
+  if (ShenandoahHeuristics::should_defer_gc()) {
+    return false;
+  }
 
   // Make sure the code below treats available without the soft tail.
   size_t soft_tail = max_capacity - capacity;
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
@@ -96,7 +96,6 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
                      byte_size_in_proper_unit(max_cset),    proper_unit_for_byte_size(max_cset),
                      byte_size_in_proper_unit(min_garbage), proper_unit_for_byte_size(min_garbage));
 
-
   // Better select garbage-first regions
   QuickSort::sort<RegionData>(data, (int)size, compare_by_garbage, false);
 
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveOldHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveOldHeuristics.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019, Red Hat, Inc. All rights reserved.
+ * Copyright (c) 2018, 2019, 2021, Red Hat, Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -24,7 +24,7 @@
 
 #include "precompiled.hpp"
 
-#include "gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.hpp"
+#include "gc/shenandoah/heuristics/shenandoahAdaptiveOldHeuristics.hpp"
 #include "gc/shenandoah/shenandoahCollectionSet.hpp"
 #include "gc/shenandoah/shenandoahFreeSet.hpp"
 #include "gc/shenandoah/shenandoahGeneration.hpp"
@@ -36,13 +36,13 @@
 // These constants are used to adjust the margin of error for the moving
 // average of the allocation rate and cycle time. The units are standard
 // deviations.
-const double ShenandoahAdaptiveHeuristics::FULL_PENALTY_SD = 0.2;
-const double ShenandoahAdaptiveHeuristics::DEGENERATE_PENALTY_SD = 0.1;
+const double ShenandoahAdaptiveOldHeuristics::FULL_PENALTY_SD = 0.2;
+const double ShenandoahAdaptiveOldHeuristics::DEGENERATE_PENALTY_SD = 0.1;
 
 // These are used to decide if we want to make any adjustments at all
 // at the end of a successful concurrent cycle.
-const double ShenandoahAdaptiveHeuristics::LOWEST_EXPECTED_AVAILABLE_AT_END = -0.5;
-const double ShenandoahAdaptiveHeuristics::HIGHEST_EXPECTED_AVAILABLE_AT_END = 0.5;
+const double ShenandoahAdaptiveOldHeuristics::LOWEST_EXPECTED_AVAILABLE_AT_END = -0.5;
+const double ShenandoahAdaptiveOldHeuristics::HIGHEST_EXPECTED_AVAILABLE_AT_END = 0.5;
 
 // These values are the confidence interval expressed as standard deviations.
 // At the minimum confidence level, there is a 25% chance that the true value of
@@ -51,20 +51,20 @@ const double ShenandoahAdaptiveHeuristics::HIGHEST_EXPECTED_AVAILABLE_AT_END = 0
 // MAXIMUM_CONFIDENCE interval here means there is a one in a thousand chance
 // that the true value of our estimate is outside the interval. These are used
 // as bounds on the adjustments applied at the outcome of a GC cycle.
-const double ShenandoahAdaptiveHeuristics::MINIMUM_CONFIDENCE = 0.319; // 25%
-const double ShenandoahAdaptiveHeuristics::MAXIMUM_CONFIDENCE = 3.291; // 99.9%
+const double ShenandoahAdaptiveOldHeuristics::MINIMUM_CONFIDENCE = 0.319; // 25%
+const double ShenandoahAdaptiveOldHeuristics::MAXIMUM_CONFIDENCE = 3.291; // 99.9%
 
-ShenandoahAdaptiveHeuristics::ShenandoahAdaptiveHeuristics(ShenandoahGeneration* generation) :
-  ShenandoahHeuristics(generation),
+ShenandoahAdaptiveOldHeuristics::ShenandoahAdaptiveOldHeuristics(ShenandoahGeneration* generation) :
+  ShenandoahOldHeuristics(generation),
   _margin_of_error_sd(ShenandoahAdaptiveInitialConfidence),
   _spike_threshold_sd(ShenandoahAdaptiveInitialSpikeThreshold),
   _last_trigger(OTHER) { }
 
-ShenandoahAdaptiveHeuristics::~ShenandoahAdaptiveHeuristics() {}
+ShenandoahAdaptiveOldHeuristics::~ShenandoahAdaptiveOldHeuristics() {}
 
-void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(ShenandoahCollectionSet* cset,
-                                                                         RegionData* data, size_t size,
-                                                                         size_t actual_free) {
+void ShenandoahAdaptiveOldHeuristics::choose_collection_set_from_regiondata(ShenandoahCollectionSet* cset,
+                                                                            RegionData* data, size_t size,
+                                                                            size_t actual_free) {
   size_t garbage_threshold = ShenandoahHeapRegion::region_size_bytes() * ShenandoahGarbageThreshold / 100;
 
   // The logic for cset selection in adaptive is as follows:
@@ -121,12 +121,12 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
   }
 }
 
-void ShenandoahAdaptiveHeuristics::record_cycle_start() {
+void ShenandoahAdaptiveOldHeuristics::record_cycle_start() {
   ShenandoahHeuristics::record_cycle_start();
   _allocation_rate.allocation_counter_reset();
 }
 
-void ShenandoahAdaptiveHeuristics::record_success_concurrent() {
+void ShenandoahAdaptiveOldHeuristics::record_success_concurrent() {
   ShenandoahHeuristics::record_success_concurrent();
 
   size_t available = ShenandoahHeap::heap()->free_set()->available();
@@ -176,7 +176,7 @@ void ShenandoahAdaptiveHeuristics::record_success_concurrent() {
   }
 }
 
-void ShenandoahAdaptiveHeuristics::record_success_degenerated() {
+void ShenandoahAdaptiveOldHeuristics::record_success_degenerated() {
   ShenandoahHeuristics::record_success_degenerated();
   // Adjust both trigger's parameters in the case of a degenerated GC because
   // either of them should have triggered earlier to avoid this case.
@@ -184,7 +184,7 @@ void ShenandoahAdaptiveHeuristics::record_success_degenerated() {
   adjust_spike_threshold(DEGENERATE_PENALTY_SD);
 }
 
-void ShenandoahAdaptiveHeuristics::record_success_full() {
+void ShenandoahAdaptiveOldHeuristics::record_success_full() {
   ShenandoahHeuristics::record_success_full();
   // Adjust both trigger's parameters in the case of a full GC because
   // either of them should have triggered earlier to avoid this case.
@@ -196,7 +196,7 @@ static double saturate(double value, double min, double max) {
   return MAX2(MIN2(value, max), min);
 }
 
-bool ShenandoahAdaptiveHeuristics::should_start_gc() {
+bool ShenandoahAdaptiveOldHeuristics::should_start_gc() {
   size_t max_capacity = _generation->max_capacity();
   size_t capacity = _generation->soft_max_capacity();
   size_t available = _generation->available();
@@ -276,7 +276,7 @@ bool ShenandoahAdaptiveHeuristics::should_start_gc() {
   return ShenandoahHeuristics::should_start_gc();
 }
 
-void ShenandoahAdaptiveHeuristics::adjust_last_trigger_parameters(double amount) {
+void ShenandoahAdaptiveOldHeuristics::adjust_last_trigger_parameters(double amount) {
   switch (_last_trigger) {
     case RATE:
       adjust_margin_of_error(amount);
@@ -292,78 +292,13 @@ void ShenandoahAdaptiveHeuristics::adjust_last_trigger_parameters(double amount)
   }
 }
 
-void ShenandoahAdaptiveHeuristics::adjust_margin_of_error(double amount) {
+void ShenandoahAdaptiveOldHeuristics::adjust_margin_of_error(double amount) {
   _margin_of_error_sd = saturate(_margin_of_error_sd + amount, MINIMUM_CONFIDENCE, MAXIMUM_CONFIDENCE);
   log_debug(gc, ergo)("Margin of error now %.2f", _margin_of_error_sd);
 }
 
-void ShenandoahAdaptiveHeuristics::adjust_spike_threshold(double amount) {
+void ShenandoahAdaptiveOldHeuristics::adjust_spike_threshold(double amount) {
   _spike_threshold_sd = saturate(_spike_threshold_sd - amount, MINIMUM_CONFIDENCE, MAXIMUM_CONFIDENCE);
   log_debug(gc, ergo)("Spike threshold now: %.2f", _spike_threshold_sd);
 }
 
-ShenandoahAllocationRate::ShenandoahAllocationRate() :
-  _last_sample_time(os::elapsedTime()),
-  _last_sample_value(0),
-  _interval_sec(1.0 / ShenandoahAdaptiveSampleFrequencyHz),
-  _rate(int(ShenandoahAdaptiveSampleSizeSeconds * ShenandoahAdaptiveSampleFrequencyHz), ShenandoahAdaptiveDecayFactor),
-  _rate_avg(int(ShenandoahAdaptiveSampleSizeSeconds * ShenandoahAdaptiveSampleFrequencyHz), ShenandoahAdaptiveDecayFactor) {
-}
-
-double ShenandoahAllocationRate::sample(size_t allocated) {
-  double now = os::elapsedTime();
-  double rate = 0.0;
-  if (now - _last_sample_time > _interval_sec) {
-    if (allocated >= _last_sample_value) {
-      rate = instantaneous_rate(now, allocated);
-      _rate.add(rate);
-      _rate_avg.add(_rate.avg());
-    }
-
-    _last_sample_time = now;
-    _last_sample_value = allocated;
-  }
-  return rate;
-}
-
-double ShenandoahAllocationRate::upper_bound(double sds) const {
-  // Here we are using the standard deviation of the computed running
-  // average, rather than the standard deviation of the samples that went
-  // into the moving average. This is a much more stable value and is tied
-  // to the actual statistic in use (moving average over samples of averages).
-  return _rate.davg() + (sds * _rate_avg.dsd());
-}
-
-void ShenandoahAllocationRate::allocation_counter_reset() {
-  _last_sample_time = os::elapsedTime();
-  _last_sample_value = 0;
-}
-
-bool ShenandoahAllocationRate::is_spiking(double rate, double threshold) const {
-  if (rate <= 0.0) {
-    return false;
-  }
-
-  double sd = _rate.sd();
-  if (sd > 0) {
-    // There is a small chance that that rate has already been sampled, but it
-    // seems not to matter in practice.
-    double z_score = (rate - _rate.avg()) / sd;
-    if (z_score > threshold) {
-      return true;
-    }
-  }
-  return false;
-}
-
-double ShenandoahAllocationRate::instantaneous_rate(size_t allocated) const {
-  return instantaneous_rate(os::elapsedTime(), allocated);
-}
-
-double ShenandoahAllocationRate::instantaneous_rate(double time, size_t allocated) const {
-  size_t last_value = _last_sample_value;
-  double last_time = _last_sample_time;
-  size_t allocation_delta = (allocated > last_value) ? (allocated - last_value) : 0;
-  double time_delta_sec = time - last_time;
-  return (time_delta_sec > 0)  ? (allocation_delta / time_delta_sec) : 0;
-}
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveOldHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveOldHeuristics.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019, 2021, Red Hat, Inc. All rights reserved.
+ * Copyright (c) 2018, 2021, Red Hat, Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -24,6 +24,7 @@
 
 #include "precompiled.hpp"
 
+#include "gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.hpp"
 #include "gc/shenandoah/heuristics/shenandoahAdaptiveOldHeuristics.hpp"
 #include "gc/shenandoah/shenandoahCollectionSet.hpp"
 #include "gc/shenandoah/shenandoahFreeSet.hpp"
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveOldHeuristics.hpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveOldHeuristics.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019, Red Hat, Inc. All rights reserved.
+ * Copyright (c) 2018, 2019, 2021, Red Hat, Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -22,40 +22,19 @@
  *
  */
 
-#ifndef SHARE_GC_SHENANDOAH_HEURISTICS_SHENANDOAHADAPTIVEHEURISTICS_HPP
-#define SHARE_GC_SHENANDOAH_HEURISTICS_SHENANDOAHADAPTIVEHEURISTICS_HPP
+#ifndef SHARE_GC_SHENANDOAH_HEURISTICS_SHENANDOAHADAPTIVEOLDHEURISTICS_HPP
+#define SHARE_GC_SHENANDOAH_HEURISTICS_SHENANDOAHADAPTIVEOLDHEURISTICS_HPP
 
-#include "gc/shenandoah/heuristics/shenandoahHeuristics.hpp"
+#include "gc/shenandoah/heuristics/shenandoahOldHeuristics.hpp"
+#include "gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.hpp"
 #include "gc/shenandoah/shenandoahPhaseTimings.hpp"
 #include "utilities/numberSeq.hpp"
 
-class ShenandoahAllocationRate : public CHeapObj<mtGC> {
- public:
-  explicit ShenandoahAllocationRate();
-  void allocation_counter_reset();
-
-  double sample(size_t allocated);
-
-  double instantaneous_rate(size_t allocated) const;
-  double upper_bound(double sds) const;
-  bool is_spiking(double rate, double threshold) const;
-
- private:
-
-  double instantaneous_rate(double time, size_t allocated) const;
-
-  double _last_sample_time;
-  size_t _last_sample_value;
-  double _interval_sec;
-  TruncatedSeq _rate;
-  TruncatedSeq _rate_avg;
-};
-
-class ShenandoahAdaptiveHeuristics : public ShenandoahHeuristics {
+class ShenandoahAdaptiveOldHeuristics : public ShenandoahOldHeuristics {
 public:
-  ShenandoahAdaptiveHeuristics(ShenandoahGeneration* generation);
+  ShenandoahAdaptiveOldHeuristics(ShenandoahGeneration* generation);
 
-  virtual ~ShenandoahAdaptiveHeuristics();
+  virtual ~ShenandoahAdaptiveOldHeuristics();
 
   virtual void choose_collection_set_from_regiondata(ShenandoahCollectionSet* cset,
                                                      RegionData* data, size_t size,
@@ -68,7 +47,7 @@ public:
 
   virtual bool should_start_gc();
 
-  virtual const char* name()     { return "Adaptive"; }
+  virtual const char* name()     { return "AdaptiveOld"; }
   virtual bool is_diagnostic()   { return false; }
   virtual bool is_experimental() { return false; }
 
@@ -128,4 +107,4 @@ public:
   TruncatedSeq _available;
 };
 
-#endif // SHARE_GC_SHENANDOAH_HEURISTICS_SHENANDOAHADAPTIVEHEURISTICS_HPP
+#endif // SHARE_GC_SHENANDOAH_HEURISTICS_SHENANDOAHADAPTIVEOLDHEURISTICS_HPP
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveOldHeuristics.hpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveOldHeuristics.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019, 2021, Red Hat, Inc. All rights reserved.
+ * Copyright (c) 2018, 2021, Red Hat, Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -26,9 +26,10 @@
 #define SHARE_GC_SHENANDOAH_HEURISTICS_SHENANDOAHADAPTIVEOLDHEURISTICS_HPP
 
 #include "gc/shenandoah/heuristics/shenandoahOldHeuristics.hpp"
-#include "gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.hpp"
 #include "gc/shenandoah/shenandoahPhaseTimings.hpp"
 #include "utilities/numberSeq.hpp"
+
+class ShenandoahAllocationRate;
 
 class ShenandoahAdaptiveOldHeuristics : public ShenandoahOldHeuristics {
 public:
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAggressiveHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAggressiveHeuristics.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019 Red Hat, Inc. All rights reserved.
+ * Copyright (c) 2018, 2019, Red Hat, Inc. All rights reserved.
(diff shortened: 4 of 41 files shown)


earthling-amzn:
Think it wants a description of the changes in the body of the PR.

kdnilsen:
I'm not sure how to address the "empty pull request body" problem.  The purpose of this pull request is described in the commit comments:
This code adds to generational Shenandoah the ability to perform concurrent garbage collection of young-gen and old-gen memory.  Following completion of an old-gen concurrent marking effort, we select certain old-gen hap regions to serve as candidates for future collection sets.  All dead objects within the old-gen heap regions that are not part of this candidate set are coalesced and filled so that remembered-set scanning of these old-gen heap regions will not be confused by "zombie objects" (objects that old-gen has decided are dead, which reside in regions that have not yet been collected).  After concurrently coalescing and filling these dead objects, each subsequent young-gen evacuation pass includes a subset of the old-gen candidates until all candidates have been collected.  This code passes TIER1 and hotspot_gc_shenandoah jtreg tests without regressions.  A new jtreg test has been added to exercises concurrent old/young GC.

rkennke:
> I'm not sure how to address the "empty pull request body" problem. The purpose of this pull request is described in the commit comments:
> This code adds to generational Shenandoah the ability to perform concurrent garbage collection of young-gen and old-gen memory. Following completion of an old-gen concurrent marking effort, we select certain old-gen heap regions to serve as candidates for future collection sets. All dead objects within the old-gen heap regions that are not part of this candidate set are coalesced and filled so that remembered-set scanning of these old-gen heap regions will not be confused by "zombie objects" (objects that old-gen has decided are dead, which reside in regions that have not yet been collected). After concurrently coalescing and filling these dead objects, each subsequent young-gen evacuation pass includes a subset of the old-gen candidates until all candidates have been collected. This code passes TIER1 and hotspot_gc_shenandoah jtreg tests without regressions. A new jtreg test has been added to exercises concurrent old/young GC.

Please add this text in the PR description above. Klick '...' -> 'Edit' and enter the text there.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -450,6 +502,11 @@ void ShenandoahHeapRegion::oop_iterate_objects(OopIterateClosure* blk, bool fill
           heap->card_scan()->register_object(obj_addr);
         obj_addr += obj->oop_iterate_size(blk);
       } else {
+        // Object is not marked, accumulate it into span of unmarked objects to be coalesced and filled
+
+        // TODO:  We can optimize this.  Don't visit each of possibly many consecutive unmarked objects.
+        // Instead, skip to the following marked object, as indicated by the first following mark bit in
+        // marking_context.
         int size = obj->size();
```
zhengyu123, line 510: Same as above

src/hotspot/share/gc/shenandoah/shenandoahOldGC.cpp
```diff
@@ -84,3 +124,50 @@ bool ShenandoahOldGC::collect(GCCause::Cause cause) {
...
+  char msg[1024];
+  ShenandoahHeap* const heap = ShenandoahHeap::heap();
+
+  entry_coalesce_and_fill_message(msg, sizeof(msg));
+  ShenandoahConcurrentPhase gc_phase(msg, ShenandoahPhaseTimings::coalesce_and_fill);
+
+  // Not sure if I want/need any of the following here:
+  //
+  // TraceCollectorStats tcs(heap->monitoring_support()->concurrent_collection_counters());
+  // EventMark em("%s", msg);
```
zhengyu123, line 167: Yes, I think we want them.

src/hotspot/share/gc/shenandoah/shenandoahPhaseTimings.hpp
```diff
@@ -64,6 +64,8 @@ class outputStream;
   f(final_mark,                                     "Pause Final Mark (N)")            \
   f(finish_mark,                                    "  Finish Mark")                   \
   SHENANDOAH_PAR_PHASE_DO(finish_mark_,             "    FM: ", f)                     \
+  f(coalesce_and_fill,                              "Coalesce and Fill Old Dead")      \
```
zhengyu123, line 67: Does not seem to be used anywhere.

zhengyu123 (changes requested):

rkennke:
Please, check out jcheck complaints under the 'Checks' tab and fix them. I'll start a review in the meantime.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
```diff
@@ -75,116 +84,221 @@ ShenandoahHeuristics::~ShenandoahHeuristics() {
 void ShenandoahHeuristics::choose_collection_set(ShenandoahCollectionSet* collection_set) {
   assert(collection_set->count() == 0, "Must be empty");
 
-  ShenandoahHeap* heap = ShenandoahHeap::heap();
+  if (_generation->generation_mode() == OLD) {
+    // Old-gen doesn't actually choose a collection set to be evacuated by its own gang of worker tasks.
+    // Instead, it computes the set of regions to be evacuated by subsequent young-gen evacuation passes.
+    prepare_for_old_collections();
+  } else {
+    ShenandoahHeap* heap = ShenandoahHeap::heap();
```
rkennke, line 92: I'd factor out the whole else block here to call prepare_for_young_collection() or whatever is fitting there.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
```diff
@@ -304,6 +436,155 @@ double ShenandoahHeuristics::time_since_last_gc() const {
 
 bool ShenandoahHeuristics::in_generation(ShenandoahHeapRegion* region) {
   return (_generation->generation_mode() == GLOBAL)
-      || (_generation->generation_mode() == YOUNG && region->affiliation() == YOUNG_GENERATION)
-      || (_generation->generation_mode() == OLD && region->affiliation() == OLD_GENERATION);
+    || (_generation->generation_mode() == YOUNG && region->affiliation() == YOUNG_GENERATION)
+    || (_generation->generation_mode() == OLD && region->affiliation() == OLD_GENERATION);
```
rkennke, line 440: I wouldn't change indentation here, or if it needs changing, then line up the || with the _generation above. Iow, shift it 4 spaces to the right instead of 2 spaces to the left.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
```diff
@@ -304,6 +436,155 @@ double ShenandoahHeuristics::time_since_last_gc() const {
...
+  ShenandoahHeap* heap = ShenandoahHeap::heap();
+  uint free_regions = 0;
+  size_t cand_idx = 0;
+  size_t total_garbage = 0;
+  size_t num_regions = heap->num_regions();
+
+  RegionData* candidates = _region_data;
+  for (size_t i = 0; i < num_regions; i++) {
+    ShenandoahHeapRegion* region = heap->get_region(i);
+    if (!in_generation(region))
```
rkennke, line 454: Please always use { and } in if (and else) constructs.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
```diff
@@ -304,6 +436,155 @@ double ShenandoahHeuristics::time_since_last_gc() const {
...
+  // Give special treatment to humongous regions.  Assume humongous regions is entirely
+  // garbage or entirely non-garbage.  Assume that a head humongous region and the associated
+  // humongous continuous regions are uniformly entirely garbage or entirely non-garbage.
+  //
+  // Sift garbage humongous regions to front, non-garbage humongous regions to end of array.
+  size_t first_non_humongous_empty = 0;
+  size_t first_humongous_non_empty = cand_idx;
+
+  // This loop is written as while rather than for because of
+  // suspected gcc error in translating/optimizing for-loop
```
rkennke, line 475: WTF, really? Do we have any indication of this?

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
```diff
@@ -304,6 +436,155 @@ double ShenandoahHeuristics::time_since_last_gc() const {
...
+
+
+  // Prioritize regions to select garbage-first regions
+  QuickSort::sort<RegionData>(candidates + first_non_humongous_empty, (int)(first_humongous_non_empty - first_non_humongous_empty),
+                              compare_by_garbage, false);
+
+  // Any old-gen region that contains 50% garbage or more is to be
+  // evacuated.  In the future, this threshold percentage may be specified on
+  // the command line or preferrably determined by dynamic heuristics.
+#define CollectionThresholdGarbagePercent 50
```
rkennke, line 515: We already have ShenandoahGarbageThreshold (defaults to 60 iirc) maybe we want to use this? Or use separate global/old/young thresholds?

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.hpp
```diff
@@ -58,6 +58,26 @@ class ShenandoahCollectionSet;
...
+//   which OLD regions need their garbage to be coalesced and filled.
+//   OLD also needs to know when all of its candidate evacuation regions
+//   have been evacuated because it cannot trigger a new OLD
+//   concurrent mark until that has happened.
+//  GLOBAL: not sure what this needs, or how it interacts with OLD, YOUNG
+//   heuristics.
+//
+// For expediency, we'll have a single class that is a "union" of
+// necessary functionality.
+
```
rkennke, line 80: Hmmhmm, this seems a suboptimal design. My intuition would have been to not have each heuristic need to know about any old heuristic, but keep them separate, and compose functionality elsewhere, but I leave that to you. The way it looks now it may end up recursive (young-heuristics->old-heuristics->old-heuristics->...-> etc)

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
```diff
@@ -515,6 +515,7 @@ void ShenandoahControlThread::resume_concurrent_old_cycle(ShenandoahGeneration*
   // marking. This flag here (passed by reference) is used to control
   // precisely where the regulator is allowed to cancel a GC.
   ShenandoahOldGC gc(generation, _allow_old_preemption);
+
```
rkennke, line 518: Lone whitespace change here.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
```diff
@@ -126,7 +127,6 @@ void ShenandoahGeneration::reset_mark_bitmap() {
 
 void ShenandoahGeneration::prepare_gc() {
   reset_mark_bitmap();
-
```
rkennke, line 129: Lone whitespace change here.

test/hotspot/jtreg/gc/shenandoah/generational/TestConcurrentEvac.java
```diff
@@ -0,0 +1,209 @@
...
+
+import sun.hotspot.WhiteBox;
+import java.util.Random;
+import java.util.HashMap;
+
+/* WARNING!
+ *  As of the date on which this test was added into the jtreg suite, heuristic
+ *  of old-gen GC passes is very simplistic.  A further shortcoming of the
+ *  Generational Shenandoah as of introduction of this test is that it does
+ *  currently support full GC.  If garbage collection falls behind mutator
```
rkennke, line 35: You probably meant to say that it "does currently *not* support full GC"

rkennke (changes requested):
Nice progress!
First-pass review follows.

src/hotspot/share/gc/shenandoah/shenandoahPhaseTimings.hpp
kdnilsen: Is used in shenandoahOldGC.cpp:entry_coalesce_and_fill().

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -403,9 +403,61 @@ void ShenandoahHeapRegion::print_on(outputStream* st) const {
...
+      }
+      assert(obj->klass() != NULL, "klass should not be NULL");
+      obj_addr += obj->size();
+    } else {
+      // Object is not marked, accumulate it into span of unmarked objects to be coalesced and filled
+
+      // TODO: We can optimize this.  Don't visit each of possibly many consecutive unmarked objects.
+      // Instead, skip to the following marked object, as indicated by the first following mark bit in
+      // marking_context.
+      int size = obj->size();
```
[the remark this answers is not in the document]
kdnilsen: I've fixed this now.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
kdnilsen: I've also fixed this now.

src/hotspot/share/gc/shenandoah/shenandoahOldGC.cpp
kdnilsen: I've uncommented this so it is now included.

rkennke:
I am thinking about how to better design the heuristics so to avoid pollution of every single heuristic with a reference to (possibly non-existing) old-gen heuristic. How about this:
- In the base ShenandoahHeuristic add a virtual and empty prime_collection_set() that is called where the block if (_old_heuristics != NULL) { prime_collection_set_with_old_candidates(); } is now.
- Leave all heuristics untouched and also don't introduce a new reference _old_heuristics. Instead:
- Add a new heuristic e.g. ShenandoahGenerationalHeuristic that has pointers to two other heuristics: one for each new-gen and old-gen. Override all relevant methods to delegate to new-gen-heuristics, and implement prime_collection_set() to implement what you're proposing here already. In other words, this heuristic would be mostly composite of old and new gen heuristic.

kdnilsen:
Was actually in the middle of doing something similar to this.  I can redirect towards your solution or continue with the one I am working on.   Here's how mine differs from yours:

1. I would introduce a subclass of ShenandoahHeuristics to be called ShenandoahOldHeuristics.  This would hold all the new methods that are specific to old-generation heuristics, currently added to ShenandoahHeuristics: prime_collection_set()  (renamed from prime_collection_set_with_old_candidates), prepare_for_old_collection(), start_old_evacuations(), unprocessed_old_collection_candidates(), next_old_collection_candidates(), consume_old_collection_candidate(), old_coalesce_and_fill_candidates(), get_coalesce_and_fill_candidates(), should_defer_gc()
2. I would pass a new ShenandoahOldHeuristics* old_heuristics argument to choose_collection_set() (possibly null in the case that we are doing GLOBAL - non-generation GC) and this argument would be passed down into prepare_for_other_collection().
3. I would hoist the check for should_defer_gc() into ShenandoahRegulatorThread::start_old_cycle() so that we don't have to "pollute" all the existing heuristics with this test.

As with your proposal, there would no longer be a _old_heuristics field within ShenandoahHeuristics.  This is why I have to pass in the new old_heuristics argument to the invocation of choose_collection_set().

Are you ok with this approach?

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
earthling-amzn: We will remove this comment. We saw some pretty strange behavior while debugging, but it turned out to be a bug in `rr`.

rkennke:
> Was actually in the middle of doing something similar to this. I can redirect towards your solution or continue with the one I am working on. Here's how mine differs from yours:
> 
>     1. I would introduce a subclass of ShenandoahHeuristics to be called ShenandoahOldHeuristics.  This would hold all the new methods that are specific to old-generation heuristics, currently added to ShenandoahHeuristics: prime_collection_set()  (renamed from prime_collection_set_with_old_candidates), prepare_for_old_collection(), start_old_evacuations(), unprocessed_old_collection_candidates(), next_old_collection_candidates(), consume_old_collection_candidate(), old_coalesce_and_fill_candidates(), get_coalesce_and_fill_candidates(), should_defer_gc()
> 
>     2. I would pass a new ShenandoahOldHeuristics* old_heuristics argument to choose_collection_set() (possibly null in the case that we are doing GLOBAL - non-generation GC) and this argument would be passed down into prepare_for_other_collection().
> 
>     3. I would hoist the check for should_defer_gc() into ShenandoahRegulatorThread::start_old_cycle() so that we don't have to "pollute" all the existing heuristics with this test.
> 
> 
> As with your proposal, there would no longer be a _old_heuristics field within ShenandoahHeuristics. This is why I have to pass in the new old_heuristics argument to the invocation of choose_collection_set().
> 
> Are you ok with this approach?

That sounds reasonable. Go for it.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
kdnilsen: I've refactored this code.  This comment is no longer relevant.

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
kdnilsen: To be fixed in next commit

test/hotspot/jtreg/gc/shenandoah/generational/TestConcurrentEvac.java
kdnilsen: You are right.  Will be corrected in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.hpp
kdnilsen: This has been addressed with recently committed refactoring of ShenandoahOldHeuristics and subclasses.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
kdnilsen: I think I've fixed this now.  Auto-indent of my editor got carried away and I apparently lost track of original spacing.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
kdnilsen: Ok.  Fixed.  (This code moved to shenadoahOldHeuristics.cpp)

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
kdnilsen: Comment will be gone in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.hpp
kdnilsen: This has been addressed with recently pushed refactoring of ShenandoahOldHeuristics and subclasses.

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
kdnilsen: Removed this blank line.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
kdnilsen: Added this blank line back in.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
kdnilsen: Comment has been removed.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
kdnilsen: I've made and tested this change now.  Appears to me that default of ShenandoahGarbageThreshold is 25.  On established stress workloads, this change has the effect of choosing larger collection sets for old-gen.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahCompactOldHeuristics.hpp
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright (c) 2018, 2019, 2021, Red Hat, Inc. All rights reserved.
```
rkennke, line 2: Same as above.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2020, Red Hat, Inc. All rights reserved.
+ * Copyright (c) 2018, 2020, 2021, Red Hat, Inc. All rights reserved.
```
rkennke, line 2: Same as above.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
```diff
@@ -162,14 +160,21 @@ void ShenandoahHeuristics::choose_collection_set(ShenandoahCollectionSet* collec
   size_t immediate_percent = (total_garbage == 0) ? 0 : (immediate_garbage * 100 / total_garbage);
 
   if (immediate_percent <= ShenandoahImmediateThreshold) {
+
+    if (old_heuristics != NULL) {
+      old_heuristics->prime_collection_set(collection_set);
+    }
+    // else, thisi s global collection and doesn't need to prime_collection_set()
```
rkennke, line 167: Typo here 'thisi s'

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
```diff
@@ -162,14 +160,21 @@ void ShenandoahHeuristics::choose_collection_set(ShenandoahCollectionSet* collec
...
+    // Add young-gen regions into the collection set.  This is a virtual call, implemented differently by each
+    // of the heuristics subclasses.
     choose_collection_set_from_regiondata(collection_set, candidates, cand_idx, immediate_garbage + free);
   }
 
   size_t cset_percent = (total_garbage == 0) ? 0 : (collection_set->garbage() * 100 / total_garbage);
 
   size_t collectable_garbage = collection_set->garbage() + immediate_garbage;
   size_t collectable_garbage_percent = (total_garbage == 0) ? 0 : (collectable_garbage * 100 / total_garbage);
-
```
rkennke, line 172: Stray newline change.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
```diff
@@ -196,6 +201,7 @@ void ShenandoahHeuristics::record_cycle_end() {
 }
 
 bool ShenandoahHeuristics::should_start_gc() {
+
```
rkennke, line 204: Stray newline change.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.hpp
```diff
@@ -134,6 +151,8 @@ class ShenandoahHeuristics : public CHeapObj<mtGC> {
   virtual void initialize();
 
   double time_since_last_gc() const;
+
+
```
rkennke, line 155: Stray newlines.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -0,0 +1,309 @@
...
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ *
+ */
+
+#include "gc/shenandoah/heuristics/shenandoahOldHeuristics.hpp"
+#include "precompiled.hpp"
```
rkennke, line 26: Convention is to always put include for precompiled.hpp first.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -0,0 +1,309 @@
...
+  // TODO We should probably enforce this, but there is no enforcement currently.  Key idea: if there is not
+  // sufficient memory within old-gen to hold an object that wants to be promoted, defer promotion until a
+  // subsequent evacuation pass.  Since enforcement may be expensive, requiring frequent synchronization
+  // between mutator and GC threads, here's an alternative "greedy" mitigation strategy: Set the parameter's
+  // value so overflow is "very rare".  In the case that we experience overflow, evacuate what we can from
+  // within the old collection set, but don't evacuate everything.  At the end of evacuation, any collection
+  // set region that was not fully evacuated cannot be recycled.  It becomes a prime candidate for the next
+  // collection set selection.  Here, we'd rather fall back to this contingent behavior than force a full STW
+  // collection.
+#define PROMOTION_BUDGET_BYTES (ShenandoahHeapRegion::region_size_bytes() / 2)
```
rkennke, line 67: I'd put macro definitions at the top of the file, right below the includes.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+  // Prioritize regions to select garbage-first regions
+  QuickSort::sort<RegionData>(candidates + first_non_humongous_empty, (int)(first_humongous_non_empty - first_non_humongous_empty),
+                              compare_by_garbage, false);
+
+  // Any old-gen region that contains (ShenandoahGarbageThreshold
+  // (default value 25))% garbage or more is to be evacuated.  In the
+  // future, this threshold percentage may be specified on the command
+  // line or preferrably determined by dynamic heuristics.
+#define CollectionThresholdGarbagePercent ShenandoahGarbageThreshold
```
rkennke, line 220: Same as above: please place define at top of file.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -0,0 +1,309 @@
...
+    }
+  }
+
+  // If we reach here, all of non-humogous old-gen regions are candidates for collection set.
+  _hidden_next_old_collection_candidate = 0;
+  _hidden_old_collection_candidates = first_humongous_non_empty;
+  _first_coalesce_and_fill_candidate = 0;
+  _old_coalesce_and_fill_candidates = 0;
+
+#undef CollectionThresholdGarbagePercent
```
rkennke, line 249: I don't think the constants need to be undef-ed when defined at top of file. Alternatively, you may consider making them true C++ constants.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahPassiveOldHeuristics.cpp
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright (c) 2018, 2019, 2021, Red Hat, Inc. All rights reserved.
```
rkennke, line 2: Same as above, only use first and last years.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahPassiveOldHeuristics.hpp
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright (c) 2018, 2019, 2021, Red Hat, Inc. All rights reserved.
```
rkennke, line 2: Same as above.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahStaticHeuristics.hpp
```diff
@@ -27,6 +27,9 @@
 
 #include "gc/shenandoah/heuristics/shenandoahHeuristics.hpp"
 
+// TODO: Make this a subclass of ShenandoahHeuristics and create a "tailored" copy of this class to be subclass of
+// ShenandoahOldHeuristics.  Fix the constructor implementation to invoke ShenandoahHeuristics() superclass constructor.
+// Change ShenandoahMode::initialize_old_heuristics() to return the subclass of ShenandoahOldHeuristics.
```
rkennke, line 32: Is this comment still valid? You seem to propose a ShenandoahStaticOldHeuristics too.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahStaticOldHeuristics.cpp
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright (c) 2018, 2019, 2021, Red Hat, Inc. All rights reserved.
```
rkennke, line 2: Same as above.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahStaticOldHeuristics.hpp
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (c) 2018, 2019, 2021, Red Hat, Inc. All rights reserved.
```
rkennke, line 2: Same as above.

rkennke (changes requested):
Nice. I believe I only have complaints about cosmetic issues.
One design-related question remains: why is it necessary to duplicate all heuristics as old variants? Is there any chance the existing heuristics can be reused? Or does it make sense to have different variants for old? Asking out of curiosity, I'm leaving the design to you.

kdnilsen:
Thanks much for the quick feedback.  I believe I can get through all of these later today.

Regarding the question of whether it makes sense to duplicate the heuristics, I believe the answer is yes.  Here is some of the rationale:

1. In general, we expect old-gen to behave very differently than young-gen.  If all is properly configured, young-gen will experience high allocation rates and very high infant mortality.  While old-gen will experience relatively low allocation (promotion) rates and no infant mortality.  If properly configured, we do not promote to old-gen until we have confirmed that an object is no more likely to die in infancy than to die in old age.

2. The old-gen heuristics need to maintain additional information that is not required by the young-gen heuristics, so all of the old-gen subclass heuristics need to derive from a different base class.  Among information required by old-gen is the candidate collection set, the coalesce-and-fill set, and awareness of whether previously selected old-gen candidate collection set has been fully processed.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
```diff
@@ -96,6 +96,7 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
                      byte_size_in_proper_unit(max_cset),    proper_unit_for_byte_size(max_cset),
                      byte_size_in_proper_unit(min_garbage), proper_unit_for_byte_size(min_garbage));
 
+
```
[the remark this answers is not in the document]
kdnilsen: Fixed in next commit

src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveOldHeuristics.cpp
```diff
@@ -0,0 +1,304 @@
+/*
+ * Copyright (c) 2018, 2019, 2021, Red Hat, Inc. All rights reserved.
```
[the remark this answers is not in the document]
kdnilsen: Fixed in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveOldHeuristics.hpp
```diff
@@ -0,0 +1,110 @@
+/*
+ * Copyright (c) 2018, 2019, 2021, Red Hat, Inc. All rights reserved.
```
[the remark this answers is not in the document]
kdnilsen: Fixed in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveOldHeuristics.hpp
```diff
@@ -0,0 +1,110 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ *
+ */
+
+#ifndef SHARE_GC_SHENANDOAH_HEURISTICS_SHENANDOAHADAPTIVEOLDHEURISTICS_HPP
+#define SHARE_GC_SHENANDOAH_HEURISTICS_SHENANDOAHADAPTIVEOLDHEURISTICS_HPP
+
+#include "gc/shenandoah/heuristics/shenandoahOldHeuristics.hpp"
+#include "gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.hpp"
```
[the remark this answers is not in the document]
kdnilsen: The reason for including is to provide definition of ShenandoahAllocationRate, which is used by both ShenandoahAdatpvieHeuristics and ShenandoahAdaptiveOldHeuristics.  The code will compile if I remove the include here, but insert it into shenandoahAdaptiveOldHeuristics.cpp.  That's how the new commit does it.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahAggressiveHeuristics.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019, Red Hat, Inc. All rights reserved.
+ * Copyright (c) 2018, 2019 Red Hat, Inc. All rights reserved.
```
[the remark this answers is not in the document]
kdnilsen: Fixed in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahAggressiveOldHeuristics.cpp
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright (c) 2018, 2019, 2021, Red Hat, Inc. All rights reserved.
```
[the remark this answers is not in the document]
kdnilsen: Fixed in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahAggressiveOldHeuristics.hpp
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (c) 2018, 2019, 2021, Red Hat, Inc. All rights reserved.
```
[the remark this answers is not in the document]
kdnilsen: Fixed in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahCompactOldHeuristics.cpp
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright (c) 2018, 2019, 2021, Red Hat, Inc. All rights reserved.
```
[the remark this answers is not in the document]
kdnilsen: Fixed in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahCompactOldHeuristics.hpp
kdnilsen: Fixed in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
kdnilsen: Fixed in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
kdnilsen: Fixed in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
kdnilsen: Fixed in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
kdnilsen: Fixed in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.hpp
kdnilsen: Fixed in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
kdnilsen: Fixed in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
kdnilsen: replaced with C++ comment in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
kdnilsen: replaced with C++ constant in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
kdnilsen: Replaced with C++ constant in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahPassiveOldHeuristics.cpp
kdnilsen: Fixed in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahPassiveOldHeuristics.hpp
kdnilsen: Fixed in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahStaticHeuristics.hpp
kdnilsen: Oops.  Forgot to remove that.  Removed in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahStaticOldHeuristics.cpp
kdnilsen: Fixed in next commit.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahStaticOldHeuristics.hpp
kdnilsen: Fixed in next commit.

rkennke (approved):
Looks good to me, thank you!

src/hotspot/share/gc/shenandoah/shenandoahOldGC.cpp
```diff
@@ -44,6 +81,12 @@ bool ShenandoahOldGC::collect(GCCause::Cause cause) {
   // Complete marking under STW
   vmop_entry_final_mark();
 
+  ShenandoahOldHeuristics* old_heuristics = heap->old_heuristics();
+
+  entry_coalesce_and_fill();
+
+  old_heuristics->start_old_evacuations();
```
zhengyu123, line 88: Could you encapsulate in entry_old_evacuations() with proper timing/logging?  it really helps future debugging the crashes with hs_err file.

zhengyu123 (changes requested):

src/hotspot/share/gc/shenandoah/mode/shenandoahPassiveMode.cpp
```diff
@@ -59,3 +60,11 @@ ShenandoahHeuristics* ShenandoahPassiveMode::initialize_heuristics(ShenandoahGen
   ShouldNotReachHere();
   return NULL;
 }
+
+ShenandoahOldHeuristics* ShenandoahPassiveMode::initialize_old_heuristics(ShenandoahGeneration* generation) const {
+  if (ShenandoahGCHeuristics != NULL) {
```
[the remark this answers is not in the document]
kdnilsen: Thanks for careful second review of the code.  This is changed to assert in next commit.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.hpp
```diff
@@ -124,6 +126,9 @@ class ShenandoahGeneration : public CHeapObj<mtGC> {
  protected:
 
   virtual bool is_concurrent_mark_in_progress() = 0;
+
+ private:
```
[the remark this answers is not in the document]
kdnilsen: Thanks.  I've fixed private.  Also fixed protected above.  They both had spaces.  Will see these changes in next commit.

src/hotspot/share/gc/shenandoah/mode/shenandoahMode.cpp
```diff
@@ -48,3 +52,22 @@ ShenandoahHeuristics* ShenandoahMode::initialize_heuristics(ShenandoahGeneration
   ShouldNotReachHere();
   return NULL;
 }
+
+ShenandoahOldHeuristics* ShenandoahMode::initialize_old_heuristics(ShenandoahGeneration* generation) const {
+
+  if (ShenandoahGCHeuristics != NULL) {
```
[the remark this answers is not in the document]
kdnilsen: Thanks.  Replaced with assert.  Will see in next commit.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -492,6 +493,7 @@ void ShenandoahHeap::initialize_heuristics() {
 ShenandoahHeap::ShenandoahHeap(ShenandoahCollectorPolicy* policy) :
   CollectedHeap(),
   _gc_generation(NULL),
+  _old_heuristics(NULL),
```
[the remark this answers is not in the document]
kdnilsen: Thanks.  Fixed in next commit.

src/hotspot/share/gc/shenandoah/shenandoahOldGC.cpp
kdnilsen: Thanks.  I've created new function entry_old_evacuations(), which calls entry_coalesce_and_fill() followed by old_heuristics->start_old_evacuations().  There is already timing/logging in entry_coalesce_and_fill().  old_heuristics->start_old_evacuations() just "flips a switch" so that subsequent evacuations will fold in old-gen collection sets so I don't think we need any timing there.

This will show up in next commit.

zhengyu123 (approved):

kdnilsen:
Are we looking for further reviews or is this ok to be merged now?

kdnilsen:
Thanks

PR #30, opened 2021-04-12
Use PLABs for old gen allocations, including promotions.

bernd-aws:
Introducing a 3rd kind of LAB for generational mode: "PLAB".

Shenandoah already has TLABs and GCLABs. The latter pertain firmly to young gen and allocations in old gen are so far always shared allocations, without LAB. With this PR, old gen allocations go through LABs as well and we separate those from the other two kinds. In other words, each generation has its separate GC-decicated LAB. 

When not in generational mode, GCLABs work as before and PLABs are never used.

The new diagnostic flag -XX:-ShenandoahUsePLAB can be used to turn off PLAB use, so that all old gen allocations are shared as before this PR. By default PLABs will be used.

Note that In source code, the type of both GCLABs and PLABs is "PLAB*", sonce GCLABs already had that type, but variable references and other value-level identifiers are consistently named "gclab" and "plab".

In ShenandoahFreeSet::allocate_single() there is a notable additional change that can easily be reverted if needed. Instead of fitting the intended allocation into any free or occupied region that can hold it, we now search for an occupied region in the same generation first, before considering free regions. The idea is to pack regions more densely as long as they have space left. This provides more opportunity for region borrowing between generations, should the need arise. For single-generational Shenandoah, it should not make a huge difference, right? Or is it crucial to find any free slot hit in that search loop ASAP?

Full collections need to set empty/trash regions that they compact into to non-FREE. For now, they will be made "YOUNG". (See the second commit.)

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/30/head:pull/30` \
`$ git checkout pull/30`

`$ git checkout pull/30` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/30/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 30`

`$ git pr show -t 30`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/30.diff">https://git.openjdk.java.net/shenandoah/pull/30.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
@@ -423,11 +423,9 @@ void ShenandoahFreeSet::flip_to_gc(ShenandoahHeapRegion* r) {
   }
   assert_bounds();
 
-  shenandoah_assert_heaplocked();
-  try_recycle_trashed(r);
-
-  assert(r->is_empty(), "Region must be empty after flipping from mutator to GC.");
-  r->set_affiliation(FREE);
+  // We do not ensure that the region is no longer trash,
+  // relying on try_allocate_in(), which always comes next,
+  // to recycle trash before attempting to allocate anything in the region.
 }
 
 void ShenandoahFreeSet::clear() {
--- a/src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
@@ -35,7 +35,6 @@
 #include "memory/resourceArea.hpp"
 #include "runtime/orderAccess.hpp"
 
-
 ShenandoahFreeSet::ShenandoahFreeSet(ShenandoahHeap* heap, size_t max_regions) :
   _heap(heap),
   _mutator_free_bitmap(max_regions, mtGC),
@@ -63,6 +62,23 @@ bool ShenandoahFreeSet::is_collector_free(size_t idx) const {
   assert (idx < _max, "index is sane: " SIZE_FORMAT " < " SIZE_FORMAT " (left: " SIZE_FORMAT ", right: " SIZE_FORMAT ")",
           idx, _max, _collector_leftmost, _collector_rightmost);
   return _collector_free_bitmap.at(idx);
+}
+
+HeapWord* ShenandoahFreeSet::allocate_with_affiliation(ShenandoahRegionAffiliation affiliation, ShenandoahAllocRequest& req, bool& in_new_region) {
+  for (size_t c = _collector_rightmost + 1; c > _collector_leftmost; c--) {
+    // size_t is unsigned, need to dodge underflow when _leftmost = 0
+    size_t idx = c - 1;
+    if (is_collector_free(idx)) {
+      ShenandoahHeapRegion* r = _heap->get_region(idx);
+      if (r->affiliation() == affiliation) {
+        HeapWord* result = try_allocate_in(r, req, in_new_region);
+        if (result != NULL) {
+          return result;
+        }
+      }
+    }
+  }
+  return NULL;
 }
 
 HeapWord* ShenandoahFreeSet::allocate_single(ShenandoahAllocRequest& req, bool& in_new_region) {
@@ -96,34 +112,17 @@ HeapWord* ShenandoahFreeSet::allocate_single(ShenandoahAllocRequest& req, bool& 
       break;
     }
     case ShenandoahAllocRequest::_alloc_gclab:
+    case ShenandoahAllocRequest::_alloc_plab:
     case ShenandoahAllocRequest::_alloc_shared_gc: {
-      // size_t is unsigned, need to dodge underflow when _leftmost = 0
-
-      // Fast-path: try to allocate in the collector view first
-      for (size_t c = _collector_rightmost + 1; c > _collector_leftmost; c--) {
-        size_t idx = c - 1;
-        if (is_collector_free(idx)) {
-          ShenandoahHeapRegion* r = _heap->get_region(idx);
-          if (r->is_young() && req.is_old()) {
-            // We don't want to cannibalize a young region to satisfy
-            // an evacuation from an old region.
-            continue;
-          }
-          HeapWord* result = try_allocate_in(r, req, in_new_region);
-          if (result != NULL) {
-            if (r->is_old()) {
-              // HEY! This is a very coarse card marking. We hope to repair
-              // such cards during remembered set scanning.
-
-              // HEY! To support full generality with alternative remembered set implementations,
-              // is preferable to not make direct access to the current card_table implementation.
-              //  Try ShenandoahHeap::heap()->card_scan()->mark_range_as_dirty(result, req.actual_size());
-
-              ShenandoahBarrierSet::barrier_set()->card_table()->dirty_MemRegion(MemRegion(result, req.actual_size()));
-            }
-            return result;
-          }
-        }
+      // First try to fit into a region that is already in use in the same generation.
+      HeapWord* result = allocate_with_affiliation(req.affiliation(), req, in_new_region);
+      if (result != NULL) {
+        return result;
+      }
+      // Then try a free region that is dedicated to GC allocations.
+      result = allocate_with_affiliation(FREE, req, in_new_region);
+      if (result != NULL) {
+        return result;
       }
 
       // No dice. Can we borrow space from mutator view?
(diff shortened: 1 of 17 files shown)


src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -98,6 +99,7 @@ void ShenandoahHeapRegion::make_regular_allocation() {
     case _empty_uncommitted:
       do_commit();
     case _empty_committed:
+      set_affiliation(YOUNG_GENERATION);
```
earthling-amzn, line 102: It's not clear to me why we would set the affiliation here. Do we know this region isn't going to be used for evacuations in old gen?

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
```diff
@@ -240,13 +257,6 @@ inline oop ShenandoahHeap::evacuate_object(oop p, Thread* thread) {
     } else if (mark.age() >= InitialTenuringThreshold) {
       oop result = try_evacuate_object(p, thread, r, OLD_GENERATION);
       if (result != NULL) {
-        // TODO: Just marking the cards covering this object dirty
-        // may overall be less efficient than scanning it now for references to young gen
-        // or other alternatives like deferred card marking or scanning.
-        // We should revisit this.
-        // Furthermore, the object start should be registered for remset scanning.
-        MemRegion mr(cast_from_oop<HeapWord*>(result), result->size());
```
earthling-amzn, line 248: Don't we need this? The object going into the old region may have pointers to young that need to be scanned.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -96,64 +112,33 @@ HeapWord* ShenandoahFreeSet::allocate_single(ShenandoahAllocRequest& req, bool&
...
           ShenandoahHeapRegion* r = _heap->get_region(idx);
           if (can_allocate_from(r)) {
-            if (r->is_young() && req.is_old()) {
-              continue;
-            }
-
             flip_to_gc(r);
             HeapWord *result = try_allocate_in(r, req, in_new_region);
             if (result != NULL) {
-              if (r->is_old()) {
```
earthling-amzn, line 147: Sorry, it's not clear to me why we aren't concerned about pointers to young anymore?

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
bernd-aws: We do need this, but not necessarily right here. I moved this close to where the object transitions, in try_evacuate_object. When a thread succeeds in moving the object there, its card is dirtied and it is registered right away. See line ~340, where it says "// Successfully evacuated. Our copy is now the public one!".

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
bernd-aws: Not as long as no content has been written into such objects. Here we are merely allocating an object that does not have references yet. If the mutator gets to it afterwards to plant references into it, this will still be in young gen, so no remset action is required. The only way to make this object hold old->young references is by means of promotion. That's covered in try_evacuate_object() for normal-sized objects (see above). For humongous objects, it's supposed to be covered in the course of region promotion.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
bernd-aws: This is to fix full single-generational collections. But yes, not good enough in the long run. => I need to put a TODO note to remind us of later repairing proper full collections in generational mode.

bernd-aws: The TODO is in the next commit I filed for this PR. The text for it is taken from a similar place that needs to be fixed in a very similar way.

bernd-aws:
Every time I edit the description text to remove the part about being draft and testing, the GitHub bot edits it back. So, I declare here that the bugs in question at the time of opening the PR as draft have been fixed and that it is now ready for review, not in draft.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -96,64 +112,33 @@ HeapWord* ShenandoahFreeSet::allocate_single(ShenandoahAllocRequest& req, bool&
...
-        }
+      // First try to fit into a region that is already in use in the same generation.
+      HeapWord* result = allocate_with_affiliation(req.affiliation(), req, in_new_region);
+      if (result != NULL) {
+        return result;
+      }
+      // Then try a free region that is dedicated to GC allocations.
+      result = allocate_with_affiliation(FREE, req, in_new_region);
+      if (result != NULL) {
+        return result;
```
rkennke, line 125: I am not sure about this. What must not happen here is that we mix new allocations and evacuations into the same region. That is because we need to update references in evacuated objects later, but not for new objects (we're guaranteed to only store to-space refs into new objects). I guess it's similar for promotions: we do need to update refs in promoted objects. That means we can mix promotions and evacuations, but cannot mix either of them with new allocations. Can you clarify this?

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -439,6 +422,12 @@ void ShenandoahFreeSet::flip_to_gc(ShenandoahHeapRegion* r) {
     adjust_bounds();
   }
   assert_bounds();
+
+  shenandoah_assert_heaplocked();
+  try_recycle_trashed(r);
+
+  assert(r->is_empty(), "Region must be empty after flipping from mutator to GC.");
+  r->set_affiliation(FREE);
```
rkennke, line 430: Why is this block needed now?

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -831,6 +831,55 @@ HeapWord* ShenandoahHeap::allocate_from_gclab_slow(Thread* thread, size_t size)
...
+    // ensure that the returned space is not considered parsable by
+    // any concurrent GC thread.
+    size_t hdr_size = oopDesc::header_size();
+    Copy::fill_to_words(plab_buf + hdr_size, actual_size - hdr_size, badHeapWordVal);
+#endif // ASSERT
+  }
+  plab->set_buf(plab_buf, actual_size);
+  return plab->allocate(size);
+}
+
```
rkennke, line 882: Is this a full (or almost full) copy of allocate_from_gclab_slow() ? Maybe this warrants some refactoring/sharing?

rkennke (commented):
Nice progress! I have a few questions/comments.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
earthling-amzn: I think this could be an assert. The region should already be `FREE` at this point.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
bmathiske: Yes, this is a full copy. Will refactor.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
bernd-aws: There is no mixing between new allocations and evacuations or promotions, because new allocations are handled in the separate switch cases _alloc_tlab and _alloc_shared, only using the mutator view. Here we use the GC view (except for overflow into trash/empty mutator regions that get "flipped").

Even though from an old gen perspective promotions are "incoming allocations", they bypass the to-space invariant by copying object content without passing barriers. In that they are in the same boat/region/PLAB with evacuations from old gen to old gen. Yes, they also need to be processed by update-refs. And we mix them in PLABs, just as we already did in case _alloc_shared_gc.

allocate_with_affiliation() is introduced for separating the hunt for a suitable region in the same generation from the hunt for a free region. Instead of taking whichever comes first, we now attempt to fit into a non-free region first. I can undo this if there is a good reason to not try to pack existing non-free regions first.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
bernd-aws: When debugging, I came across regions that were "trash" but not "FREE".

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
bernd-aws: I tried to refactor: it quickly gets quite ugly. A lot of gclab values are hardwired into identifier names. For example, check out the upper part of allocate_from_gclab_slow()!  How can we abstract over ShenandoahThreadLocalData::gclab_size() vs. ShenandoahThreadLocalData::plab_size()? This issues repeats with other tags/labels/identifers, e.g., when creating a ShenandoahAllocRequest(). I'd have to come up with a whole new way of addressing LABs generically and that would touch a lot of code.

bernd-aws: A template with 4 parameters might work, but I have doubts that methods as template parameters would increase readability.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
bernd-aws: Turns out that was OK, because the code at hand here is actually always followed by code that does the same (in try_allocate_in()). My code insertion here is thus redundant. I will remove it.

rkennke (approved):
Looks good to me. Thank you!

PR #32, opened 2021-04-29
Fix region promotion

bernd-aws:
There are bugs in region promotion. 
- Humongous object continuation region remainder object starts are not registered for remembered set scanning purposes. 
- Also, iterating over humongous objects at promotion time to set card table values is inefficient. This is better delayed until card table scanning. 
- Regions that are not either regular or humongous should not get promoted, in particular the cset region state needs to be excluded.
- We need to prevent regions that contain TLABs or GCLABs from getting promoted with these remaining active. One could retire those LABs once again at final update refs, but that would be inefficient. An alternative defensive and robust fix is to exclude regions that contain any TLAB or GCLAB from promotion. This should not restrict promotions much, because regions tend to fill up as they age and eventually they won't have LABs.

All of the above is fixed by the commits proposed here.
This PR is best applied after PR #30, the outcome of which it is rebased to at the moment.


earthling-amzn:
I've created a rebased version of this PR: https://github.com/openjdk/shenandoah/pull/36. Please close this PR.

rkennke:
I believe you can simply close this PR now, right?

rkennke:
Closing this, PR#36 rebased and fix this.

PR #36, opened 2021-05-05
Fix Region Promotion

earthling-amzn:
This is rebased pull request of https://github.com/openjdk/shenandoah/pull/32.

There are bugs in region promotion.

* Humongous object continuation region remainder object starts are not registered for remembered set scanning purposes.
* Also, iterating over humongous objects at promotion time to set card table values is inefficient. This is better delayed until card table scanning.
* Regions that are not either regular or humongous should not get promoted, in particular the cset region state needs to be excluded.
* We need to prevent regions that contain TLABs or GCLABs from getting promoted with these remaining active. One could retire those LABs once again at final update refs, but that would be inefficient. An alternative defensive and robust fix is to exclude regions that contain any TLAB or GCLAB from promotion. This should not restrict promotions much, because regions tend to fill up as they age and eventually they won't have LABs.

All of the above is fixed by the commits proposed here.
This PR is best applied after PR #30, the outcome of which it is rebased to at the moment.

---------
### Progress

### Reviewers

### Contributors
 * Bernd Mathiske `<bmathiske@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/36/head:pull/36` \
`$ git checkout pull/36`

`$ git checkout pull/36` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/36/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 36`

`$ git pr show -t 36`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/36.diff">https://git.openjdk.java.net/shenandoah/pull/36.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -1012,43 +1012,6 @@ HeapWord* ShenandoahHeap::allocate_memory(ShenandoahAllocRequest& req) {
 HeapWord* ShenandoahHeap::allocate_memory_under_lock(ShenandoahAllocRequest& req, bool& in_new_region) {
   ShenandoahHeapLocker locker(lock());
   HeapWord* result = _free_set->allocate(req, in_new_region);
-  // Register the newly allocated object while we're holding the global lock since there's no synchronization
-  // built in to the implementation of register_object().  There are potential races when multiple independent
-  // threads are allocating objects, some of which might span the same card region.  For example, consider
-  // a card table's memory region within which three objects are being allocated by three different threads:
-  //
-  // objects being "concurrently" allocated:
-  //    [-----a------][-----b-----][--------------c------------------]
-  //            [---- card table memory range --------------]
-  //
-  // Before any objects are allocated, this card's memory range holds no objects.  Note that:
-  //   allocation of object a wants to set the has-object, first-start, and last-start attributes of the preceding card region.
-  //   allocation of object b wants to set the has-object, first-start, and last-start attributes of this card region.
-  //   allocation of object c also wants to set the has-object, first-start, and last-start attributes of this card region.
-  //
-  // The thread allocating b and the thread allocating c can "race" in various ways, resulting in confusion, such as last-start
-  // representing object b while first-start represents object c.  This is why we need to require all register_object()
-  // invocations to be "mutually exclusive".  Later, when we use GCLABs and PLABs to allocate memory for promotions and evacuations,
-  // the protocol may work something like the following:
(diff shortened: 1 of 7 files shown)


src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -1299,6 +1262,37 @@ void ShenandoahHeap::gclabs_retire(bool resize) {
...
+class ShenandoahTagGCLABClosure : public ThreadClosure {
+public:
+  void do_thread(Thread* thread) {
+    PLAB* gclab = ShenandoahThreadLocalData::gclab(thread);
+    assert(gclab != NULL, "GCLAB should be initialized for %s", thread->name());
+    if (gclab->words_remaining() > 0) {
+      ShenandoahHeapRegion* r = ShenandoahHeap::heap()->heap_region_containing(gclab->allocate(0));
+      r->set_young_lab_flag();
+    }
+  }
```
rkennke, line 1274: Are you not interested in TLABs and PLABs here?

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -1299,6 +1262,37 @@ void ShenandoahHeap::gclabs_retire(bool resize) {
...
+  ShenandoahTagGCLABClosure cl;
+  workers()->threads_do(&cl);
+  for (JavaThreadIteratorWithHandle jtiwh; JavaThread *t = jtiwh.next(); ) {
+    cl.do_thread(t);
+    ThreadLocalAllocBuffer& tlab = t->tlab();
+    if (tlab.end() != NULL) {
+      ShenandoahHeapRegion* r = heap_region_containing(tlab.start());
+      r->set_young_lab_flag();
+    }
+  }
```
rkennke, line 1293: Ah ok I see you handle TLABs here. Maybe move this into the ShenandoahTagGCLABClosure? What about PLABs?

rkennke (commented):
Looks good. Only minor comment/question.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
earthling-amzn: I think we aren't worried about PLABs here because they should only exist in old regions (i.e., ones that have already promoted). I'll refactor this to not loop over the threads twice and add an assert that the thread's plab exists in an old region.

earthling-amzn: On second look, the `GCLABClosure` is only visiting the worker threads (they won't have TLABs).

rkennke (approved):
Right. Ok then.

PR #50, opened 2021-06-21
Introduce second mark bitmap to support remembered set scan

earthling-amzn:
## Motivation
After final mark, Shenandoah chooses the collection set (regions that will be reclaimed). If there is a sufficient amount of reclaimable memory in regions that are 100% garbage, it recycles them immediately and skips the evacuation and update reference phases (there are no pointers into these 100% garbage regions). Note that in this case there are still dead objects in regions outside of the collection set. During subsequent remembered set scans, Shenandoah must safely iterate objects in regions of memory indicated by marked cards. Specifically, it must not attempt to iterate over oops in the dead objects outside the collection set. Prior to the changes in this review, Generational Shenandoah disarmed these dead objects by coalescing them and then overwriting them with filler objects during the update references phase. There are some drawbacks to this approach:

* Originally, Generational Shenandoah used the oopDesc::size API to compute the size of filler objects. This API uses the klass for the oop to provide the size. Since we were using the size method on dead oops, if there were no instances of the klass still alive it could be unloaded during weak root processing and the size method would fail. To prevent this issue, we disabled ClassUnloading. 
* We now use the mark bitmap to compute the size of dead objects, but there is still a risk here: when young collections coalesce and fill dead objects during concurrent old generation marking, the mark bitmap is incomplete.
* It forced us to go through evacuation and update reference phases for every cycle. Our tests (Dacapo) show that approximately 33% of collections (with the default settings) find enough pure garbage regions to complete without evacuations.
* It forces the collector to perform work proportional to the amount of garbage outside the collection set (in the old generation). For performance reasons, the collector should avoid any work which is proportional to the amount of garbage.

## Overview of Changes
Rather than filling in dead objects to provide safety to the remembered set scan, the scan now uses the marking bit map to skip over dead objects. This leverages the work already completed during the marking phase and no longer requires any additional maintenance work during the update reference phase. There are, however, a few wrinkles to consider:

* The evacuation phase needs to update the mark bitmap. This essentially mimics the work currently done to register objects with the remembered set.
* During concurrent old marking, the remembered set scan cannot rely on the mark bitmap for the old generation. For this reason, a second bitmap has been added. At the beginning of a concurrent old mark, the bitmaps are switched. During concurrent old marking, the remembered set scan will use the previous (i.e., stable) bitmap for determining object liveness.
* Though not directly related to these changes, a fix for reference processing has been implemented in terms of these changes (i.e., the fix uses the secondary bitmap). Please note, the "fix" for reference processing is incorrect. It does not preserve the atomicity required when null'ing out references. Work on reference processing is proceeding on a separate branch, unrelated to these changes.

## Remaining Work

* Object registration should no longer be necessary. We can use the mark bitmap to tell us if a card has any objects.
* Shenandoah has an optimization to avoid clearing bitmap regions that are known to contain no marked bits. This optimization is currently disabled. It needs to be debugged (I think it is confused by extra marks made during evacuation).
* Region promotions that happen during a concurrent old mark must copy the bitmap of the promoted region - this is currently done in a most inefficient way.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/50/head:pull/50` \
`$ git checkout pull/50`

`$ git checkout pull/50` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/50/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 50`

`$ git pr show -t 50`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/50.diff">https://git.openjdk.java.net/shenandoah/pull/50.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahBitmapRegion.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahBitmapRegion.cpp
@@ -1,3 +1,26 @@
+/*
+ * Copyright (c) 2013, 2021, Red Hat, Inc. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ *
+ */
 
 #include "gc/shared/gc_globals.hpp"
 #include "gc/shenandoah/shenandoahAsserts.hpp"
--- a/src/hotspot/share/gc/shenandoah/shenandoahBitmapRegion.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahBitmapRegion.cpp
@@ -22,6 +22,8 @@
  *
  */
 
+#include "precompiled.hpp"
+
 #include "gc/shared/gc_globals.hpp"
 #include "gc/shenandoah/shenandoahAsserts.hpp"
 #include "gc/shenandoah/shenandoahBitmapRegion.hpp"
--- a/src/hotspot/share/gc/shenandoah/shenandoahBitmapRegion.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahBitmapRegion.cpp
@@ -1,0 +1,165 @@
+
+#include "gc/shared/gc_globals.hpp"
+#include "gc/shenandoah/shenandoahAsserts.hpp"
+#include "gc/shenandoah/shenandoahBitmapRegion.hpp"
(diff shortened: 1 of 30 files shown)


src/hotspot/share/gc/shenandoah/shenandoahBitmapRegion.hpp
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+#ifndef SHARE_GC_SHENANDOAH_SHENANDOAHBITMAPREGION_HPP
+#define SHARE_GC_SHENANDOAH_SHENANDOAHBITMAPREGION_HPP
+
+#include "memory/memRegion.hpp"
+
+class ShenandoahHeapRegion;
+
+class ShenandoahBitmapRegion {
```
rkennke, line 32: It would be nice to have some comments here. What is this class? What's its purpose? Etc.

src/hotspot/share/gc/shenandoah/shenandoahBitmapRegion.cpp
```diff
@@ -0,0 +1,188 @@
+/*
+ * Copyright (c) 2013, 2021, Red Hat, Inc. All rights reserved.
```
rkennke, line 2: I believe a new file should have only 2021 here.

src/hotspot/share/gc/shenandoah/shenandoahBitmapRegion.hpp
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (c) 2013, 2021, Red Hat, Inc. All rights reserved.
```
rkennke, line 2: Should only have 2021 here.

rkennke (changes requested):
Looks good overall, only some cosmetic requests.

earthling-amzn:
Thanks for taking a look at this @rkennke . I apologize for the late reply (was on vacation and missed your comments). Are you concerned about the additional memory used for the second bitmap (at least two additional bits for each live object)? I think it would be possible to _only_ have the secondary bitmap slices committed during old generation marking.

rkennke (approved):
Looks good to me. Thank you!

rkennke:
Please resolve merge conflicts, and integrate.

earthling-amzn:
Closing this for now.

PR #53, opened 2021-07-22
Generational support for weak roots and references

earthling-amzn:
### Summary
The LRB for non-strong references is modified to permit resurrection of objects outside the generation being collected. In other words, resurrection is only blocked for unmarked objects in the generation being collected.

Each `ShenandoahGeneration` has its own reference processor instance. In some cases, a reference from the old generation may end up on the young generation discovered list if the reference points to a young referent (this would happen if the old reference is in the remembered set). However, young references that point to referents in the old generation are _not_ discovered. This has the effect of strongly marking the old generation referent. This also avoids the case of having young references on the old generation discovered list being evacuated/relocated while they wait for old generation reference processing (although we believe this case would be handled correctly by the existing update references code).

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/53/head:pull/53` \
`$ git checkout pull/53`

`$ git checkout pull/53` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/53/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 53`

`$ git pr show -t 53`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/53.diff">https://git.openjdk.java.net/shenandoah/pull/53.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/mode/shenandoahGenerationalMode.hpp
+++ b/src/hotspot/share/gc/shenandoah/mode/shenandoahGenerationalMode.hpp
@@ -26,7 +26,7 @@
 #define SHARE_GC_SHENANDOAH_MODE_SHENANDOAHGENERATIONALMODE_HPP
 
 #include "gc/shenandoah/mode/shenandoahMode.hpp"
-#include "oops/oop.hpp"
+#include "oops/oopsHierarchy.hpp"
 
 enum GenerationMode {
   YOUNG,
--- a/src/hotspot/share/gc/shenandoah/mode/shenandoahGenerationalMode.hpp
+++ b/src/hotspot/share/gc/shenandoah/mode/shenandoahGenerationalMode.hpp
@@ -26,6 +26,7 @@
 #define SHARE_GC_SHENANDOAH_MODE_SHENANDOAHGENERATIONALMODE_HPP
 
 #include "gc/shenandoah/mode/shenandoahMode.hpp"
+#include "oops/oop.hpp"
 
 enum GenerationMode {
   YOUNG,
@@ -39,7 +40,7 @@ enum ShenandoahRegionAffiliation {
   OLD_GENERATION
 };
 
-const char* affiliation_name(const void* ptr);
+const char* affiliation_name(oop ptr);
 const char* affiliation_name(ShenandoahRegionAffiliation type);
 
 class ShenandoahGenerationalMode : public ShenandoahMode {
--- a/src/hotspot/share/gc/shenandoah/mode/shenandoahGenerationalMode.hpp
+++ b/src/hotspot/share/gc/shenandoah/mode/shenandoahGenerationalMode.hpp
@@ -39,7 +39,8 @@ enum ShenandoahRegionAffiliation {
   OLD_GENERATION
 };
 
-const char *affiliation_name(ShenandoahRegionAffiliation type);
+const char* affiliation_name(const void* ptr);
+const char* affiliation_name(ShenandoahRegionAffiliation type);
 
 class ShenandoahGenerationalMode : public ShenandoahMode {
 public:
--- a/src/hotspot/share/gc/shenandoah/shenandoahOldGC.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahOldGC.cpp
@@ -32,6 +32,7 @@
 #include "gc/shenandoah/shenandoahGeneration.hpp"
 #include "gc/shenandoah/heuristics/shenandoahHeuristics.hpp"
 #include "gc/shenandoah/shenandoahWorkerPolicy.hpp"
+#include "prims/jvmtiTagMap.hpp"
 #include "utilities/events.hpp"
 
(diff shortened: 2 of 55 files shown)


rkennke:
Before I review this, can you give me more details of your thinking? In particular:
- why do we need one reference processor per generation? Why is a single global reference processor not enough?
- why would we want to allow resurrection of objects in the not-collected generation?

earthling-amzn:
Yes of course, there are many scenarios to consider here. We did consider a solution that would have the collector process references only during a _young_ and _global_ collection, but our ultimate goal is to eliminate the _global_ collections from generational mode. 

The need to have a reference processor per generation comes from the requirement for all references to a referent to be cleared _atomically_ by the collector. If we have only one reference processor, it would necessarily be shared by the young and old collection cycles, however this would break the _atomically_ cleared requirement. Young collection cycles are allowed to interrupt old collections. A young cycle which interrupts old marking would not be allowed to process references because marking of old (discovering of all references) would not yet be complete[1]. So we have two separate discovered lists, separated by the generation of their _referents_. Since a young collection will only collect young _referents_, we put old and young _references_ alike on the same _young_ discovered list.

This also motivates the changes in the LRB - I probably did not explain that well in the summary. The LRB code is such that it blocks resurrection only when `concurrent_weak_root_in_progress`, with the intention that weak references/roots pointing to _unmarked_ objects are destined to be cleared and so can no longer be accessed (resurrected). With two generations, it now matters which one is being collected. The weak roots are off heap[2] and so themselves are not distinguishable by generation, however, they do hold references to objects in the heap. At the beginning of old marking, the mark bitmaps are cleared. If a young collection occurs _before_ the old marking is complete it will block resurrection while `concurrent_weak_root_in_progress` is true - however, the mark bitmap for the old generation is incomplete so it should _not_ block resurrection for old objects. In other words, this change in the LRB effectively gives us `old_concurrent_weak_root_in_progress` and `young_concurrent_weak_root_in_progress` without having to introduce the complexity of additional states to the collector's state machine.


[1] As I think on this, we could consider a design that would have young cycles which interrupt old marking _defer_ reference processing to the old generation cycle, but we would need to be careful of dangling referents (that is, referents which were _not_ strongly reachable at the end of the young cycle would be collected, but the reference to them would not yet be cleared). It doesn't seem like this would be safe.
[2] This is certainly true for _some_ weak roots, but I'm not sure it's true for _all_ weak roots.

zhengyu123:
Would it be simpler to encapsulate generational info in MarkingContext for is_marked_xxx tests? e.g. for young marking context, everything in old gen returns is_marked_strong = true?

earthling-amzn:
There are `is_marked_or_old` methods in `MarkingContext` for use with the verifier (because it visits all objects, even when it's run during a young collect). I think so long as we want to process weak references in young collections _and_ old collections we need to maintain two discovered lists. I can see a way through this where we process references in young and _global_ collections (i.e., entire heap) using one discovered list (since young and global collections cannot overlap).

src/hotspot/share/gc/shenandoah/mode/shenandoahGenerationalMode.hpp
```diff
@@ -26,6 +26,7 @@
 #define SHARE_GC_SHENANDOAH_MODE_SHENANDOAHGENERATIONALMODE_HPP
 
 #include "gc/shenandoah/mode/shenandoahMode.hpp"
+#include "oops/oop.hpp"
```
rkennke, line 29: I believe the proper way to get 'oop' into scope is to include oopsHierarchy.hpp

src/hotspot/share/gc/shenandoah/shenandoahOldGC.cpp
```diff
@@ -129,21 +131,15 @@ bool ShenandoahOldGC::collect(GCCause::Cause cause) {
...
   vmop_entry_final_mark();
 
-  entry_old_evacuations();
-
   // We aren't dealing with old generation evacuation yet. Our heuristic
   // should not have built a cset in final mark.
   assert(!heap->is_evacuation_in_progress(), "Old gen evacuations are not supported");
 
-  // Concurrent stack processing
-  if (heap->is_evacuation_in_progress()) {
```
rkennke, line 139: Where has this gone?

rkennke (commented):
Looks good. I only have a few remarks/questions.

src/hotspot/share/gc/shenandoah/mode/shenandoahGenerationalMode.hpp
earthling-amzn: Sure, will make this change.

src/hotspot/share/gc/shenandoah/shenandoahOldGC.cpp
earthling-amzn: This was unreachable code (see assert just above on line 136). The old generation defers all evacuation to _mixed_ collections.

rkennke (approved):

earthling-amzn:
/integrate

Thank you!

PR #64, opened 2021-09-08
Enable remembered set verification during global collections

earthling-amzn:
There are two sets of changes here:
 * Fixes to the verifier itself to enable remembered set verification during global collection
 * Fixes to prevent the remembered set scan from parsing unmarked objects after a global collection

These changes pass the dacapo suite with `-XX:+ShenandoahVerify` and pass the shenandoah tier1 tests.

---------
### Progress

### Reviewers
 * [Kelvin Nilsen](https://openjdk.java.net/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to 7136de8c8162787eb58b902877dde35c95d27f55

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/64/head:pull/64` \
`$ git checkout pull/64`

`$ git checkout pull/64` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/64/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 64`

`$ git pr show -t 64`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/64.diff">https://git.openjdk.java.net/shenandoah/pull/64.diff</a>

</details>


src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -2634,34 +2628,36 @@ void ShenandoahHeap::verify_rem_set_at_mark() {
...
-          if (ctx) {
-            // TAMS not relevant here
-            obj_addr = ctx->get_next_marked_addr(obj_addr, t);
           } else {
-            obj_addr += obj->size();
+            // This object is not live so we don't verify dirty cards contained therein
+            ShenandoahHeapRegion* r = heap_region_containing(obj_addr);
+            HeapWord* tams = ctx->top_at_mark_start(r);
+            if (obj_addr >= tams) {
+              obj_addr += obj->size();
```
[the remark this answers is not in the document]
kdnilsen: If we verify remset while old marking is "ongoing", ctx is not "reliable".  In that case, I had been arranging for ctx to equal NULL with the understanding that obj->size() would get me to the next (possibly coalesced and filled) object within old-gen.  I don't yet follow the logic for how you can always rely on ctx being not NULL here.

src/hotspot/share/gc/shenandoah/shenandoahMark.inline.hpp
```diff
@@ -278,27 +278,24 @@ inline void ShenandoahMark::mark_through_ref(T *p, ShenandoahObjToScanQueue* q,
...
+      if (heap->mode()->is_generational()) {
         // TODO: As implemented herein, GLOBAL collections reconstruct the card table during GLOBAL concurrent
         // marking. Note that the card table is cleaned at init_mark time so it needs to be reconstructed to support
         // future young-gen collections.  It might be better to reconstruct card table in
         // ShenandoahHeapRegion::global_oop_iterate_and_fill_dead.  We could either mark all live memory as dirty, or could
         // use the GLOBAL update-refs scanning of pointers to determine precisely which cards to flag as dirty.
-        //
-        if ((GENERATION == YOUNG) && ShenandoahHeap::heap()->is_in(p) && ShenandoahHeap::heap()->is_in_old(p)) {
-          RememberedScanner* scanner = ShenandoahHeap::heap()->card_scan();
+        if (GENERATION == YOUNG && heap->is_in_old(p)) {
```
kdnilsen, line 293: I think I had added the heap->is_in(p) test because some program was sending in a non-heap value which was causing an assertion failure in is_in_old(p) during DEBUG builds (that have assertions enabled).  Maybe need to test this patch with FAST_DEBUG just to be sure we're ok with code as now written.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
```diff
@@ -553,15 +547,23 @@ ShenandoahScanRemembered<RememberedSet>::verify_registration(HeapWord* address,
...
-    HeapWord* end_of_interest = base_addr + max_offset;
+    // should represent this object.  Otherwise, last_offset is a don't care.
     do {
-      HeapWord* obj_addr = base_addr + offset;
-      oop obj = oop(base_addr + offset);
       prev_offset = offset;
-      offset = ctx->get_next_marked_addr(base_addr + offset, end_of_interest) - base_addr;
+      oop obj = cast_to_oop(base_addr + offset);
+      ShenandoahHeapRegion* region = heap->heap_region_containing(obj);
+      HeapWord* tams = ctx->top_at_mark_start(region);
```
kdnilsen, line 555: Think you can move both these lines outside the loop, because we're not crossing region boundaries here.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
```diff
@@ -553,15 +547,23 @@ ShenandoahScanRemembered<RememberedSet>::verify_registration(HeapWord* address,
     // This is a mixed evacuation: rely on mark bits to identify which objects need to be properly registered
```
kdnilsen, line 547: Maybe this comment is too restrictive.  I think the new idea is that we use ctx != nullptr for both mixed evac and for GLOBAL collections.  Am I right about this?

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
```diff
@@ -952,7 +953,7 @@ class ShenandoahScanRemembered: public CHeapObj<mtGC> {
   void coalesce_objects(HeapWord *addr, size_t length_in_words);
 
   // Return true iff this object is "properly" registered.
-  bool verify_registration(HeapWord* address, size_t size_in_words);
+  bool verify_registration(HeapWord* address, ShenandoahMarkingContext* ctx);
```
kdnilsen, line 956: There's some trickery with the value of the ctx argument that should probably be described in the API comment.  In particular, ctx is non-null iff the the marking context is "valid" for the purposes of finding live objects within old-gen memory.  It is not valid if we are doing a young collection while old-gen is concurrently marking.  When ctx is NULL, it is required that the old-gen memory has been coalesced and filled.  During mixed evacuations, we rely upon ctx within old-gen because we know that a new old-gen concurrent mark cannot begin until the previously selected evacuation candidates have all been collected.  We don't bother to coalesce-and-fill the collection candidates, but they may be scanned multiple times by young-gen (mixed) remembered set scanning while we are waiting for these candidate regions to be collected.  It's not clear to me that this patch honors all of these behaviors.

src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
```diff
@@ -1004,6 +1039,12 @@ void ShenandoahConcurrentGC::op_cleanup_complete() {
   ShenandoahHeap::heap()->free_set()->recycle_trash();
 }
 
+void ShenandoahConcurrentGC::op_global_coalesce_and_fill() {
+  ShenandoahHeap* const heap = ShenandoahHeap::heap();
+  ShenandoahGlobalCoalesceAndFill coalesce;
```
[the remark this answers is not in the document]
earthling-amzn: The `ShenandoahGlobalCoalesceAndFill` closure only acts on old active, non-humongous regions.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -2199,12 +2199,7 @@ class ShenandoahUpdateHeapRefsTask : public AbstractGangTask {
           _heap->marked_object_oop_iterate(r, &cl, update_watermark);
         } else if (r->affiliation() == ShenandoahRegionAffiliation::OLD_GENERATION) {
           if (_heap->active_generation()->generation_mode() == GLOBAL) {
-            // This code is only relevant to GLOBAL GC.  With OLD GC, all coalescing and filling is done before any relevant
-            // evacuations.
-
-            // This is an old region in a global cycle.  Make sure that the next cycle does not iterate over dead objects
-            // which haven't had their references updated.  This is not a promotion.
-            r->global_oop_iterate_and_fill_dead(&cl);
+            _heap->marked_object_oop_iterate(r, &cl, update_watermark);
```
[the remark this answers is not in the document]
earthling-amzn: I'll try putting the 'fill-after-final-mark' on the code path that skips updates references and restoring this code. That way, we'll do one or the other, but not both.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -2634,34 +2628,36 @@ void ShenandoahHeap::verify_rem_set_at_mark() {
...
+        if (!scanner->verify_registration(obj_addr, ctx)) {
           ShenandoahAsserts::print_failure(ShenandoahAsserts::_safe_all, obj, obj_addr, NULL,
                                           "Verify init-mark remembered set violation", "object not properly registered", __FILE__, __LINE__);
         }
       } else if (!r->is_humongous()) {
         HeapWord* t = r->top();
         while (obj_addr < t) {
           oop obj = oop(obj_addr);
-          // ctx->is_marked() returns true if mark bit set (TAMS not relevant here)
+          // ctx->is_marked() returns true if mark bit set or if obj above TAMS.
```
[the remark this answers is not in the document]
earthling-amzn: Will restore original comment.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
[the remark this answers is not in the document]
earthling-amzn: If `ctx` is NULL, we'll have gone into the other path and used `obj->size()` to advance `obj_addr`. I'll add an assert that `obj_addr` is below TAMS during the init-mark verification.  We only verify the remembered set during a mixed evacuation or global collect (as of this change). In both cases there should be no concurrent marking of old (I'll add an assert for this also).

src/hotspot/share/gc/shenandoah/shenandoahMark.inline.hpp
earthling-amzn: I changed the implementation of `is_in_old` to first test for `is_in` in an earlier patch.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
earthling-amzn: This method is called from the other rset verify methods (init-mark, update-references, after-full-gc). Here I've just made it take the `ctx` computed in the caller (rather than re-selecting it in the called method). The size parameter was unused.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
earthling-amzn: Yes, I'll update the comment to include global collections.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
earthling-amzn: Okay, wasn't sure if cards would always be aligned with regions.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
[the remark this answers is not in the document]
earthling-amzn: I was able to simplify this (and similar code in a few other places) because we've previously checked if the object is marked (which would return true for an object above TAMS).

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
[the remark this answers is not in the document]
earthling-amzn: After thinking about this, I'd rather do it on a separate PR. We'd still want to fill before update references when `ShenandoahVerify` is enabled.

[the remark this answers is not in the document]
kdnilsen: Thanks.  This sounds good.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -2642,14 +2642,9 @@ void ShenandoahHeap::verify_rem_set_at_mark() {
...
           } else {
-            // This object is not live so we don't verify dirty cards contained therein
-            ShenandoahHeapRegion* r = heap_region_containing(obj_addr);
-            HeapWord* tams = ctx->top_at_mark_start(r);
-            if (obj_addr >= tams) {
-              obj_addr += obj->size();
-            } else {
-              obj_addr = ctx->get_next_marked_addr(obj_addr, tams);
-            }
+            // This object is not live so we don'top verify dirty cards contained therein
```
kdnilsen, line 2645: typo here in the comment.

kdnilsen (approved):

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
earthling-amzn: Fixed, thanks.

PR #75, opened 2021-09-29
Full region promotion

kdnilsen:
Prior to this patch, full-region-promotion occurred during the final-evacuation safepoint.  When heap regions are very large (e.g. 16 MB) and a large number of heap regions need to be promoted, this was measured to require over 500 ms in certain scenarios.  Another disadvantage of the prior implementation is that entire heap regions are promoted as-is without any compaction.  A problem with this is that all fragments of garbage within the young-region are promoted into old-gen, where the target heap-region utilization is generally much higher than heap-utilization target within young-gen memory.  Once promoted, it is much more costly to compact this memory because an old-gen collection is much more expensive than a young-gen collection.

This patch moves region promotion out of the safepoint.  Regions that have reached the tenure age are biased for inclusion within the collection set and are processed during concurrent evacuation using the established conventions.  Humongous heap regions are also promoted during concurrent evacuation rather than during the safepoint.

---------
### Progress

### Reviewers
 * [William Kemper](https://openjdk.java.net/census#wkemper) (@earthling-amzn - Committer) ⚠️ Review applies to 5921401e811f4fef9aba573748e16acbbd624997

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/75/head:pull/75` \
`$ git checkout pull/75`

`$ git checkout pull/75` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/75/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 75`

`$ git pr show -t 75`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/75.diff">https://git.openjdk.java.net/shenandoah/pull/75.diff</a>

</details>


src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -916,94 +867,52 @@ void ShenandoahHeapRegion::set_affiliation(ShenandoahRegionAffiliation new_affil
...
+  oop obj = cast_to_oop(bottom());
+  assert(marking_context->is_marked(obj), "promoted humongous object should be alive");
+
+  size_t spanned_regions = ShenandoahHeapRegion::required_regions(obj->size() * HeapWordSize);
+  size_t index_limit = index() + spanned_regions;
+
+  log_debug(gc)("promoting humongous region " SIZE_FORMAT ", spanning " SIZE_FORMAT, index(), spanned_regions);
+
+  // Since this region may have served previously as OLD, it may hold obsolete object range info.
+  heap->card_scan()->reset_object_range(bottom(), bottom() + spanned_regions * ShenandoahHeapRegion::region_size_words());
```
earthling-amzn, line 890: Do we write this memory again on line 911 or 915?

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
```diff
@@ -798,8 +801,10 @@ ShenandoahScanRemembered<RememberedSet>::process_region(ShenandoahHeapRegion *re
   unsigned int cluster_size = CardTable::card_size_in_words * ShenandoahCardCluster<ShenandoahDirectCardMarkRememberedSet>::CardsPerCluster;
   size_t num_clusters = (size_t) ((num_heapwords - 1 + cluster_size) / cluster_size);
 
-  // Remembered set scanner
-  process_clusters(start_cluster_no, num_clusters, end_of_range, cl, use_write_table);
+  if (!region->is_humongous_continuation()) {
```
earthling-amzn, line 804: Is this safe? why do we no longer need to scan humongous continuations?

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
kdnilsen: The code at lines 911 and 915 clears or sets the card dirty bits.  The code here at line 890 changes the remembered set's knowledge of where objects begin.  These are two different activities.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
kdnilsen: humongous continuations are scanned when the humongous start that "governs" the continuation is scanned.  I'll add a comment here.

earthling-amzn: Wouldn't we need to change how `end_of_range` is computed for humongous starts?

kdnilsen: That's a good catch.  I'm a bit inconsistent with this.  I think this is "harmless" because end_of_range does not limit the scanning within a single humongous object.  It only limits the search for the "next object" following an initially scanned object.  I agree I should fix this up to make the code less confusing.  (Do you agree that the error is "harmless as is"?   Would appreciate your careful eyes on this.)

earthling-amzn: `end_of_range` is used earlier to compute the `count` parameter (which is named `num_clusters` in the calling function). As I read it, this _does_ limit the number of cards scanned in the `process_clusters` call.

kdnilsen: The way I read it, if you call process_clusters with an address that represents the start of a humongous region, it will go through its object-processing loop at least once (p < endp first time through) and will completely scan the object found at p, which will be a humongous object that spans multiple following regions (which may exceed count clusters scanned in total and may reach reach well beyond endp in total scan).  Do you agree with this?

earthling-amzn: Ah, okay - the `obj->oop_iterate` (or more likely, `array->oop_iterate_range`) will cover the entire humongous object if there is a dirty card in the start region. If the humongous object has no dirty cards in the start region, the last card for the object will be past the end card for this cluster and we'll `oop_iterate` it anyway.

earthling-amzn (approved):

PR #76, opened 2021-09-29
Decouple generational heuristics

earthling-amzn:
### Summary
This commit is part of an effort to improve the performance of the old generation heuristic.

* added `ShenandoahOldGCHeuristics` - choose the heuristic for old generation (adaptive is default, aggressive is not allowed, only used to trigger old marking, has no influence on cset selection for old generation).
* added `ShenandoahOldGarbageThreshold` - The minimum percentage of garbage in old region to be included in old cset.
* added `ShenandoahOldEvacReserve` - Percentage of old generation to withhold from old evacuation.
* added `ShenandoahGuaranteedOldGCInterval` - Minimum time between old generation collection (default is 10 min).
* refactored old generation heuristic class hierarchy to reduce code duplication
* added more detail to log messages and formatted byte sizes for easier human readability

### Testing
Passes tier3 shenandoah jtreg tests and Dacapo and Heapothesys on x86 and aarch64.


kdnilsen (approved):
Thanks for this cleanup.  Big improvement.  Am eager to test this out.

PR #80, opened 2021-10-05
Age cycle periodically

kdnilsen:
Add option to allow object ages to increment periodically instead of on every young cycle.


earthling-amzn (approved):
Looks good to me.

PR #82, opened 2021-10-07
Add an option to stream region sampling data to a file

earthling-amzn:
This change adds two command line options to enable and configure a feature that has Shenandoah stream region sampling data to disk. These changes are complemented by changes in the [Shenandoah Visualizer](https://github.com/openjdk/shenandoah-visualizer) project to load and replay the logged region sampling data.


src/hotspot/share/gc/shenandoah/shenandoahHeapRegionCounters.cpp
```diff
@@ -68,19 +71,25 @@ ShenandoahHeapRegionCounters::ShenandoahHeapRegionCounters() :
...
+  if (_log_file != NULL) FREE_C_HEAP_OBJ(_log_file);
 }
 
 void ShenandoahHeapRegionCounters::update() {
   if (ShenandoahRegionSampling) {
     jlong current = nanos_to_millis(os::javaTimeNanos());
     jlong last = _last_sample_millis;
     if (current - last > ShenandoahRegionSamplingRate &&
-            Atomic::cmpxchg(&_last_sample_millis, last, current) == last) {
+        Atomic::cmpxchg(&_last_sample_millis, last, current) == last) {
```
rkennke, line 92: Neither the previous nor the new version seems to align correctly.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegionCounters.hpp
```diff
@@ -86,13 +88,17 @@ class ShenandoahHeapRegionCounters : public CHeapObj<mtGC>  {
...
   ~ShenandoahHeapRegionCounters();
   void update();
 
 private:
   static jlong encode_heap_status(ShenandoahHeap* heap) ;
+
 };
 
 #endif // SHARE_GC_SHENANDOAH_SHENANDOAHHEAPREGIONCOUNTERS_HPP
+
```
rkennke, line 104: Bunch of unrelated new newlines.

src/hotspot/share/gc/shenandoah/shenandoahLogFileOutput.cpp
```diff
@@ -0,0 +1,216 @@
+/*
+ * Copyright (c) 2021, Amazon.com, Inc. All rights reserved.
+ * Copyright (c) 2015, 2020, Oracle and/or its affiliates. All rights reserved.
```
rkennke, line 3: Is this a copy of or derived from an existing source file? If not, only putting the 2021 header should be enough.

src/hotspot/share/gc/shenandoah/shenandoahLogFileOutput.hpp
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (c) 2013, 2021, Red Hat, Inc. All rights reserved.
```
rkennke, line 2: Is this a new file? Then the 2021-Amazon header should be enough. If it's derived from an existing file, then the header is ok as it is.

rkennke (approved):
Looks ok, I only have some minor nits.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegionCounters.cpp
earthling-amzn: Fixed

PR #88, opened 2021-10-12
Preempt old preparation

kdnilsen:
Previously, all old-gen regions not selected as collection set candidates were "made parseable" during the final-mark safepoint.   Making each of these regions parseable involves coalescing consecutive unmarked objects into contiguous larger objects that contain no pointers.  The process of identifying each span of consecutive unmarked objects within a large number of large heap regions was found to require hundreds of ms in some cases.  This patch changes the behavior so that preparation of heap regions for subsequent evacuation is performed during a concurrent phase of operation.  This concurrent phase must complete before any of the old heap regions placed in the candidate collection set can be evacuated.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/88/head:pull/88` \
`$ git checkout pull/88`

`$ git checkout pull/88` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/88/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 88`

`$ git pr show -t 88`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/88.diff">https://git.openjdk.java.net/shenandoah/pull/88.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
@@ -176,8 +176,8 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
     // all objects allocated within this region are above TAMS (and thus are implicitly marked).  In case this is an
     // OLD region and concurrent preparation for mixed evacuations visits this region before the start of the next
     // old-gen concurrent mark (i.e. this region is allocated following the start of old-gen concurrent mark but before
-    // concurrent preparation for mixed evacuations are completed), we mark this region as not requiring any
-    // coalesce-and-fill processing.y/ This code is only necessary if req.affiliation() is OLD, but harmless if not.
+    // concurrent preparations for mixed evacuations are completed), we mark this region as not requiring any
+    // coalesce-and-fill processing.  This code is only necessary if req.affiliation() is OLD, but harmless if not.
     r->finish_coalesce_and_fill();
     ctx->capture_top_at_mark_start(r);
 


src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
```diff
@@ -72,6 +72,21 @@ class ShenandoahBreakpointMarkScope : public StackObj {
   }
 };
 
+class ShenandoahGlobalCoalesceAndFill : public ShenandoahHeapRegionClosure {
```
earthling-amzn, line 75: Not sure how this ended up `shenandoahConcurrentGC.cpp`, last I checked it was defined in `shenandoahHeap.cpp`.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -171,6 +171,10 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
 
     r->set_affiliation(req.affiliation());
     r->set_update_watermark(r->bottom());
+
+    // Any OLD region allocated during concurrent coalesce-and-fill does not need to be coalesced and filled.
+    // This code is only necessary if req.affiliation() is old, but harmless if not.
+    r->finish_coalesce_and_fill();
```
earthling-amzn: Not sure I understand why that's the case. This region could still contain garbage objects that could break the remembered set scan. How does allocating out of the region make these objects parseable?

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -458,7 +464,17 @@ void ShenandoahHeapRegion::oop_fill_and_coalesce() {
       heap->card_scan()->coalesce_objects(obj_addr, fill_size);
       obj_addr = next_marked_obj;
     }
+    if (!ops_before_preempt_check--) {
```
earthling-amzn, line 467: I think the `C++` way is to compare against 0, not coerce to bool.

earthling-amzn (changes requested):

src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
kdnilsen: Not sure either.  It is in both places in my branch.  I see no reason for it to be here so am removing it.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
kdnilsen: Thanks.  I've changed the condition.

PR #90, opened 2021-10-14
Reset coalesce and fill boundary during global collect

earthling-amzn:
Normally, this happens during cset selection for the old generation. It also needs to happen for coalesce and fill at the end of global collections or objects may not be filled.

---------
### Progress

### Reviewers
 * [Kelvin Nilsen](https://openjdk.java.net/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to 50d110ba194e77a59e397c9bd2848e6084021d96

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/90/head:pull/90` \
`$ git checkout pull/90`

`$ git checkout pull/90` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/90/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 90`

`$ git pr show -t 90`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/90.diff">https://git.openjdk.java.net/shenandoah/pull/90.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -966,6 +966,10 @@ void ShenandoahHeap::coalesce_and_fill_old_regions() {
     virtual void heap_region_do(ShenandoahHeapRegion* region) override {
       // old region is not in the collection set and was not immediately trashed
       if (region->is_old() && region->is_active() && !region->is_humongous()) {
+        // Reset the coalesce and fill boundary because this is a global collect
+        // and cannot be preempted by young collects. We want to be sure the entire
+        // region is coalesced here and does not resume from a previously interrupted
+        // or completed coalescing.
         region->reset_coalesce_and_fill_boundary();
         region->oop_fill_and_coalesce();
       }


earthling-amzn:
We don't need it to re-register live objects, we need it to fill in the dead objects. The sequence of events looks like this:
1. FREE region services an allocation request and calls `finish_coalesce_and_fill` (this sets fill boundary to `_end`).
2. Global collection happens, dead objects in old regions outside of cset are coalesced. Coalescing begins from the coalesce boundary and walks objects until `_end`. But for regions in the state from step 1, coalescing does _nothing_.
3. Young generation remembered set scan attempts to iterate dead objects that should have been filled in step 2 and goes off the rails.

kdnilsen:
Probably means I chose bad names for these services.  :(

kdnilsen:
I'm still not sure this is the right place to invoke reset_coalesce_and_fill_boundary().  To support preemption of concurrent coalesce-and-fill, we reset_coalesce_and_fill_boundary() once for each region that needs to be coalesced and filled.  Then we begin the preemptible coalesce and fill effort.  If we preempt in the middle of processing a given region, we will come back to this region where coalesce-and-fill efforts left off, as preserved by the call to suspend_coalesce_and_fill().

Since concurrent GC also invokes this same heap_region_do() service, I believe the change you propose would make us go back to the start of each region and reprocess it multiple times each time we resume coalesce and fill following a preemption.

I think what we want here is to iterate through all relevant regions and reset_coalesce_and_fill_boundary() before a STW GC calls ShenandoahHeap::coalesce_and_fill_old_regions().

kdnilsen:
Better service names might be:
  s/reset_coalesce_and_fill_boundary/begin_preemptible_coalesce_and_fill/
  s/finish_coalesce_and_fill/end_preemptible_coalesce_and_fill/
  remove coalesce_and_fill_is_done() <--- this was apparently an incomplete thought of mine

earthling-amzn:
I think it's okay here because this code is only called to coalesce and fill regions during a global collect. Preemption during a global collect shouldn't be a concern (young collects can only preempt old marking, they can't preempt a global collect).

kdnilsen:
Sorry to be so slow to see how this works.  I see that the old-gen coalesce-and-fill effort instantiates a ShenandoahConcurrentCoalesceAndFillTask to perform coalesce-and-fill with preemption, and does not call into ShenandoahHeap::coalesce_and_fill_old_regions(), so I agree your change is ok.

It might be worthwhile to add a comment to ShenandoahHeap::coalesce_and_fill_old_regions() stating that this version of coalesce-and-fill is intended only for situations that do not need to respond to preemption requests by urgent young-gen collections.

PR #103, opened 2021-12-03
Only verify last object start for marked objects

earthling-amzn:
In some cases, verification could invoke `size` on an unmarked object resulting in a crash.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/103/head:pull/103` \
`$ git checkout pull/103`

`$ git checkout pull/103` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/103/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 103`

`$ git pr show -t 103`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/103.diff">https://git.openjdk.java.net/shenandoah/pull/103.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
@@ -399,7 +399,7 @@ ShenandoahScanRemembered<RememberedSet>::verify_registration(HeapWord* address, 
   if (max_offset > CardTable::card_size_in_words) {
     max_offset = CardTable::card_size_in_words;
   }
-  size_t prev_offset = offset;
+  size_t prev_offset;
   if (!ctx) {
     do {
       oop obj = cast_to_oop(base_addr + offset);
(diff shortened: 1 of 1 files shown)


src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
```diff
@@ -399,7 +399,7 @@ ShenandoahScanRemembered<RememberedSet>::verify_registration(HeapWord* address,
   if (max_offset > CardTable::card_size_in_words) {
     max_offset = CardTable::card_size_in_words;
   }
-  size_t prev_offset;
+  size_t prev_offset = offset;
```
kdnilsen, line 402: Is this change motivated by a detected bug, or is this just a "style" improvement?  It looks to me like this change is equivalent to the original code, because in both cases (when !ctx and ctx), we unconditionally set prev_offset to offset in the first iteration of the do-while loop.

earthling-amzn: Unconditionally setting prev_offset in the second branch (i.e., when `ctx` is not null) _is_ the bug. If the code there sets prev_offset to refer to refer to an _unmarked_ object, the following verification check to get the `size` on an unmarked object could crash. The intention here was to make sure that `prev_offset` never refers to an unmarked object, but this change isn't enough to guarantee that.

kdnilsen: Thanks for clarifying.  Looks good to me as now drafted.

kdnilsen (approved):

PR #105, opened 2022-01-27
Degenerate to marking phase if cancellation detected after final mark

earthling-amzn:
If an allocation failure occurs between the end of concurrent mark and the start of final mark, then the degenerate cycle resumes from the 'outside of cycle' degenerated point. For other modes, this is merely a performance problem (as described in [JDK-8261093](https://bugs.openjdk.java.net/browse/JDK-8261093)). For the generational mode, this is a correctness problem because the degenerated cycle will swap the remembered set cards again. Because the concurrent marking was incomplete, this could lose information about which cards should be dirty during the degenerated (re) mark.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/105/head:pull/105` \
`$ git checkout pull/105`

`$ git checkout pull/105` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/105/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 105`

`$ git pr show -t 105`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/105.diff">https://git.openjdk.java.net/shenandoah/pull/105.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
@@ -123,6 +123,8 @@ bool ShenandoahConcurrentGC::collect(GCCause::Cause cause) {
   // Complete marking under STW, and start evacuation
   vmop_entry_final_mark();
 
+  check_cancellation_and_abort(ShenandoahDegenPoint::_degenerated_mark);
+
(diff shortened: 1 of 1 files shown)


rkennke (approved):
Looks correct.

PR #107, opened 2022-02-01
Simplify crossing map implementation

kdnilsen:
The crossing map maintains a record of the first and last objects
starting within each remembered set card's range of memory addresses.
The previous implementations used bit-shifts.  It was detected during
testing that the c++ compiler was emitting incorrect code in certain
in-line expansions of the register_object() method.  gdb reports that
there are 56 distinct contexts holding the expansion of this function.
This simplified implementation does not exhibit the same broken
behavior.

---------
### Progress

### Reviewers
 * [William Kemper](https://openjdk.java.net/census#wkemper) (@earthling-amzn - Committer) ⚠️ Review applies to b562a0b4961ff1431a886ec83405b3bbf2c6702a

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/107/head:pull/107` \
`$ git checkout pull/107`

`$ git checkout pull/107` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/107/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 107`

`$ git pr show -t 107`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/107.diff">https://git.openjdk.java.net/shenandoah/pull/107.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
@@ -513,71 +513,64 @@ public:
   static const size_t CardsPerCluster = 64;
 
 private:
-  // This bit is set iff at least one object starts within a
-  // particular card region.
-  static const uint16_t ObjectStartsInCardRegion = 0x8000;
-  static const uint16_t FirstStartBits = 0x003f;
-  static const uint16_t LastStartBits = 0x0fc0;
-  static const uint16_t FirstStartShift = 0;
-  static const uint16_t LastStartShift = 6;
+  typedef struct cross_map { uint8_t first; uint8_t last; } xmap;
+  typedef union crossing_info { uint16_t short_word; xmap offsets; } crossing_info;
 
-  uint16_t *object_starts;
+  // ObjectStartsInCardRegion bit is set within a crossing_info.offsets.start iff at least one object starts within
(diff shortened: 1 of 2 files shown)


src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
```diff
@@ -513,71 +513,64 @@ class ShenandoahCardCluster: public CHeapObj<mtGC> {
...
       fatal("Insufficient memory for initializing heap");
     for (size_t i = 0; i < rs->total_cards(); i++)
-      object_starts[i] = 0;
+      object_starts[i].short_word = 0;
   }
 
   ~ShenandoahCardCluster() {
-    if (object_starts != NULL)
+    if (object_starts != nullptr)
       free(object_starts);
```
earthling-amzn, line 570: I know `ShenandoahCardCluster` is a singleton so this is a nit pick, but we should set `object_starts` to `nullptr` here to avoid double deletes.

earthling-amzn (approved):

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
kdnilsen: fixed.

PR #109, opened 2022-02-03
Improve handling of cancellation after final mark

earthling-amzn:
Handle case when gc is cancelled _after_ final mark but _before_ the control thread checks for cancellation. If final mark completes before the cancellation is detected, we can no longer degenerate to the marking phase.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/109/head:pull/109` \
`$ git checkout pull/109`

`$ git checkout pull/109` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/109/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 109`

`$ git pr show -t 109`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/109.diff">https://git.openjdk.java.net/shenandoah/pull/109.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
@@ -123,6 +123,14 @@ bool ShenandoahConcurrentGC::collect(GCCause::Cause cause) {
   // Complete marking under STW, and start evacuation
   vmop_entry_final_mark();
 
+  // If GC was cancelled before final mark, then the safepoint operation will do nothing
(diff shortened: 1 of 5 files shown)


kdnilsen (commented):
I think this needs a comment to explain how it works.  I'm also confused by the original code, but we don't really need to describe that if we are replacing it.  Would like a comment to clarify how concurrent_mark_in_progress() is still true after we're back from vmop_entry_final_mark(), and then how we assure that the condition that we want has not changed before we check_cancellation_and_abort().

kdnilsen (approved):

PR #110, opened 2022-02-04
Borrow from old gen

kdnilsen:
This commit represents multiple months of incremental performance improvements to allow generational shenandoah to run more efficiently, especially with larger heap sizes and high memory utilization.

Specific improvements are described in individual commit log messages.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/110/head:pull/110` \
`$ git checkout pull/110`

`$ git checkout pull/110` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/110/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 110`

`$ git pr show -t 110`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/110.diff">https://git.openjdk.java.net/shenandoah/pull/110.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
@@ -34,8 +34,6 @@
 #include "logging/logTag.hpp"
 #include "utilities/quickSort.hpp"
 
-#undef KELVIN_VERBOSE
-
 // These constants are used to adjust the margin of error for the moving
 // average of the allocation rate and cycle time. The units are standard
 // deviations.
@@ -149,33 +147,14 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
     // (((new_garbage < min_garbage) && (r->garbage() > ShenandoahSmallerGarbageThreshold)) || (r->garbage() > garbage_threshold))
 
     if ((new_cset <= max_cset) && ((r->garbage() > garbage_threshold) || (r->age() >= InitialTenuringThreshold))) {
-#ifdef KELVIN_VERBOSE
-      printf("CSET includes region " SIZE_FORMAT " (age: %u, garbage: " SIZE_FORMAT " (biased: " SIZE_FORMAT "), live: " SIZE_FORMAT ")\n",
-             r->index(), r->age(), r->garbage(), biased_garbage, r->get_live_data_bytes());
-#endif
       cset->add_region(r);
       live_bytes_in_collection_set += r->get_live_data_bytes();
       collected_region_count++;
       cur_cset = new_cset;
       // cur_garbage = new_garbage;
     } else if (biased_garbage == 0) {
-#ifdef KELVIN_VERBOSE
-      printf("CSET ignores region " SIZE_FORMAT " and all subsequent regions (age: %u, garbage: 0, live: " SIZE_FORMAT ")\n",
-             r->index(), r->age(), r->get_live_data_bytes());
-      for (size_t j = idx+1; j < size; j++) {
-        r = data[j]._region;
-        printf("data[" SIZE_FORMAT "]._garbage: " SIZE_FORMAT ", region " SIZE_FORMAT
-               " (age: %u, garbage: 0, live: " SIZE_FORMAT ")\n",
-               j, data[j]._garbage, r->index(), r->age(), r->get_live_data_bytes());
-      }
-
-#endif
       break;
     } else {
-#ifdef KELVIN_VERBOSE
-      printf("CSET ignores region " SIZE_FORMAT " (age: %u, garbage: " SIZE_FORMAT " out of threshold: " SIZE_FORMAT ", live: " SIZE_FORMAT ")\n",
-             r->index(), r->age(), r->garbage(), garbage_threshold, r->get_live_data_bytes());
-#endif
    }
   }
   cset->set_young_region_count(collected_region_count);
@@ -277,15 +256,6 @@ bool ShenandoahAdaptiveHeuristics::should_start_gc() {
 
   size_t min_threshold = capacity / 100 * ShenandoahMinFreeThreshold;
 
-#ifdef KELVIN_VERBOSE_SILENCED
-  // ShenandoahMinFreeThreshold is a percentage:  35 means 35%.  If available is ever below this amount of memory, then
-  // we should immediately start GC.  However, we do not endeavor to assure that we have this much working buffer available
-  // within the generation as of when we end the GC effort.
-  log_info(gc)("  available adjusted to: " SIZE_FORMAT ", min_threshold: " SIZE_FORMAT
-               ", ShenandoahMinFreeThreshold: " SIZE_FORMAT "%%",
-               available, min_threshold, ShenandoahMinFreeThreshold);
-#endif
-
   if (available < min_threshold) {
     log_info(gc)("Trigger (%s): Free (" SIZE_FORMAT "%s) is below minimum threshold (" SIZE_FORMAT "%s)",
                  _generation->name(),
@@ -369,15 +339,6 @@ bool ShenandoahAdaptiveHeuristics::should_start_gc() {
   //     ShenandoahControlIntervalMax.  The current control interval (or the max control interval) should also be added into
   //     the calculation of avg_cycle_time below.
 
-#ifdef KELVIN_VERBOSE_SILENCED
-  // Seems to me that ShenandoahAllocSpikeFactor and _gc_time_penalties might be absolute quantities rather than
-  // percentages of capacity.
-
-  log_info(gc)("  spike_headroom: " SIZE_FORMAT " (ShenandoahAllocSpikeFactor: " SIZE_FORMAT "), gc penalties: "
-               SIZE_FORMAT " (_gc_time_penalties: " SIZE_FORMAT ")",
-               spike_headroom, ShenandoahAllocSpikeFactor, penalties, _gc_time_penalties);
-#endif
-
   allocation_headroom -= MIN2(allocation_headroom, spike_headroom);
   allocation_headroom -= MIN2(allocation_headroom, penalties);
 
@@ -409,10 +370,6 @@ bool ShenandoahAdaptiveHeuristics::should_start_gc() {
     if (avg_cycle_time > original_cycle_time) {
       log_debug(gc)("%s: average GC time adjusted from: %.2f ms to %.2f ms because upward trend in live memory retention",
                     _generation->name(), original_cycle_time, avg_cycle_time);
-#ifdef KELVIN_VERBOSE
-      printf("%s: average GC time adjusted from: %.2f ms to %.2f ms because upward trend in live memory retention\n",
-                    _generation->name(), original_cycle_time, avg_cycle_time);
-#endif
     }
 
     log_info(gc)("Trigger (%s): Average GC time (%.2f ms) is above the time for average allocation rate (%.0f %sB/s) to deplete free headroom (" SIZE_FORMAT "%s) (margin of error = %.2f)",
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
@@ -89,9 +89,6 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
   // we hit max_cset. When max_cset is hit, we terminate the cset selection. Note that in this scheme,
   // ShenandoahGarbageThreshold is the soft threshold which would be ignored until min_garbage is hit.
 
-  // KELVIN OJO:
-  // max_cset is constrained by reserved_for_young_evac.  the computations are more complex than the following.
-
   size_t max_cset    = (ShenandoahHeap::heap()->get_young_evac_reserve() / ShenandoahEvacWaste);
   size_t capacity    = ShenandoahHeap::heap()->young_generation()->soft_max_capacity();
 
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
@@ -437,6 +437,7 @@ bool ShenandoahAdaptiveHeuristics::should_start_gc() {
                  _generation->name(), avg_cycle_time * 1000,
                  byte_size_in_proper_unit(rate), proper_unit_for_byte_size(rate),
                  byte_size_in_proper_unit(allocation_headroom), proper_unit_for_byte_size(allocation_headroom),
+
                  _spike_threshold_sd);
     _last_trigger = SPIKE;
     return true;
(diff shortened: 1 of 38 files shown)


src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
```diff
@@ -244,18 +273,96 @@ bool ShenandoahAdaptiveHeuristics::should_start_gc() {
...
+
   allocation_headroom -= MIN2(allocation_headroom, spike_headroom);
   allocation_headroom -= MIN2(allocation_headroom, penalties);
 
   double avg_cycle_time = _gc_time_history->davg() + (_margin_of_error_sd * _gc_time_history->dsd());
+
+  size_t last_live_memory = get_last_live_memory();
+  size_t penultimate_live_memory = get_penultimate_live_memory();
+  double original_cycle_time = avg_cycle_time;
+  if ((penultimate_live_memory < last_live_memory) && (penultimate_live_memory != 0)) {
```
earthling-amzn, line 341: The heuristic uses an exponentially weighted average for cycle times. The decay factor can be adjusted from the command line. The margin of error for estimating cycle times is also adjusted in response to degenerated cycles or successful concurrent cycles. Perhaps we should have the heuristic also adjust the decay factor? Have we evaluated the results of these changes on a variety of work loads?

src/hotspot/share/gc/shenandoah/shenandoahCollectionSet.hpp
```diff
@@ -47,6 +47,15 @@ class ShenandoahCollectionSet : public CHeapObj<mtGC> {
   size_t                _garbage;
   size_t                _used;
   size_t                _region_count;
+  size_t                _immediate_trash;
```
earthling-amzn, line 50: Might consider having collection set itself manage these new members through `add_region` API - rather than having similar looking code in each of the heuristics when the collection set is populated.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -112,15 +115,21 @@ HeapWord* ShenandoahFreeSet::allocate_single(ShenandoahAllocRequest& req, bool&
       break;
     }
     case ShenandoahAllocRequest::_alloc_gclab:
+      // GCLABs are for evacuation so we must be in evacuation phase.  If this allocation is successful, increment
+      // the relevant evac_expended rather than used value.
+      allocating_gclab = true;
```
earthling-amzn, line 120: Do we really need a new parameter here? The information is carried in the type of the allocation request (`req`) which is also passed to `allocate_with_affiliation` (the request goes all the way down to `try_allocate_in`).

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -81,6 +79,12 @@ class ShenandoahFreeSet : public CHeapObj<mtGC> {
 public:
   ShenandoahFreeSet(ShenandoahHeap* heap, size_t max_regions);
 
+  // How many regions dedicated to GC allocations (for evacuation or promotion) are currently free?
```
earthling-amzn, line 82: I agree with these comments - let's delete the question marks?

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
```diff
@@ -365,11 +375,25 @@ inline oop ShenandoahHeap::try_evacuate_object(oop p, Thread* thread, Shenandoah
...
         }
         case OLD_GENERATION: {
            if (ShenandoahUsePLAB) {
-             copy = allocate_from_plab(thread, size);
+             copy = allocate_from_plab(thread, size, is_promotion);;
+             if ((copy == nullptr) && (size < ShenandoahThreadLocalData::plab_size(thread))) {
+               // PLAB allocation failed because we are bumping up against the limit on old evacuation reserve.  Try resetting
+               // the desired PLAB size and retry PLAB allocation to avoid cascading of shared memory allocations.
+               ShenandoahThreadLocalData::set_plab_size(thread, PLAB::min_size());
+               copy = allocate_from_gclab(thread, size);
```
earthling-amzn, line 394: Should this be `allocate_from_plab` as above?

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -250,7 +250,7 @@ void ShenandoahHeapRegion::make_unpinned() {
 
 void ShenandoahHeapRegion::make_cset() {
   shenandoah_assert_heaplocked();
-  reset_age();
+  // Leave age untouched.  We need to consult the age when we are deciding whether to promote evacuated objects.
```
earthling-amzn, line 253: Many other state transitions also `reset_age` and I'm not sure it's necessary. Seems sufficient to do it only for `make_trash`.

src/hotspot/share/gc/shenandoah/shenandoahMarkClosures.cpp
```diff
@@ -60,21 +60,6 @@ void ShenandoahFinalMarkUpdateRegionStateClosure::heap_region_do(ShenandoahHeapR
...
-      // objects will sometimes lower top beneath the watermark, after which,
-      // attempts to read the watermark will assert out (watermark should not be
-      // higher than top). The right way™ to check for new allocations is to compare
-      // top with the TAMS as is done earlier in this function.
-      if (top > tams) {
-        // There have been allocations in this region since the start of the cycle.
-        // Any objects new to this region must not assimilate elevated age.
-        r->reset_age();
-      } else if (ShenandoahHeap::heap()->is_aging_cycle()) {
-        r->increment_age();
```
earthling-amzn, line 74: Hmm, we used to increment age for every region during final mark, but now we are doing it during final update refs. This means no regions will be aged for cycles that skip evacuation. Is that what we want? this will have the effect of lowering the promotion rate.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
```diff
@@ -274,12 +292,39 @@
...
+          "generation allocation pool is replenished.  During mixed "       \
+          "evacuations, the bound on amount of old-generation heap "        \
+          "regions included in the collecdtion set is the smaller "         \
+          "of the quantities specified by this parameter and the "          \
+          "size of ShenandoahEvacReserve as adjusted by the value of "      \
+          "ShenandoahOldEvacRatio.  In percents of total old-generation "   \
+          "heap size.")                                                     \
           range(1,100)                                                      \
                                                                             \
+  product(uintx, ShenandoahOldEvacRatioPer128, 16, EXPERIMENTAL,            \
```
earthling-amzn, line 309: These `Per128` options are evaluated once per cycle? I would just take the floating point hit for a more user friendly percentage.

earthling-amzn (changes requested):

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -335,9 +338,66 @@ class ShenandoahHeap : public CollectedHeap {
...
+  // that adjusted capacities equal traditional capacities except during evacuation and update refs.
+  //
+  // During evacuation, we assure that _young_evac_expended does not exceed _young_evac_reserve and that _old_evac_expended
+  // does not exceed _old_evac_reserve.  GCLAB allocations do not immediately affect used within the young generation
+  // since the adjusted capacity already accounts for the entire evacuation reserve.  Each GCLAB allocations increments
+  // _young_evac_expended rather than incrementing the affiliated generation's used value.
+  //
+  // At the end of update references, we perform the following bookkeeping activities:
+  //
+  // 1. Unadjust the capacity within young-gen and old-gen
```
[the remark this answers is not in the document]
kdnilsen: We remove the "borrowed memory" from capacity.  But by the time we get to here, we have finished evacuating and updating, so the entirety of the collection set is added back in to the allocation capacity.

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -335,9 +338,66 @@ class ShenandoahHeap : public CollectedHeap {
...
+  // of waste, it may be that one thread fails to grab a new GCLAB, this does not necessarily doom the associated evacuation
+  // effort.  If this happens, the requesting thread blocks until some other thread manages to evacuate the offending object.
+  // Only after "all" threads fail to evacuate an object do we consider the evacuation effort to have failed.
+
+  intptr_t _alloc_supplement_reserve;  // Bytes reserved for young allocations during evac and update refs
+  size_t _promotion_reserve;           // Bytes reserved within old-gen to hold the results of promotion
+
+
+  size_t _old_evac_reserve;            // Bytes reserved within old-gen to hold evacuated objects from old-gen collection set
+  size_t _old_evac_expended;           // Bytes of old-gen memory expended on old-gen evacuations?
```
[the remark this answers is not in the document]
kdnilsen: Thanks for noticing.  I'll change this.

src/hotspot/share/gc/shenandoah/shenandoahThreadLocalData.hpp
```diff
@@ -162,6 +169,50 @@ class ShenandoahThreadLocalData {
     data(thread)->_plab_size = v;
   }
 
+  static void enable_plab_promotions(Thread* thread) {
+    data(thread)->_plab_allows_promotion = true;
+  }
+
+  static void disable_plab_promotions(Thread* thread) {
```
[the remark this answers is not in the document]
kdnilsen: Good catch.  I got careless with copy and paste.  Thanks.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
kdnilsen: Exposure to variety of workload conditions is very limited.  All of our work on heuristics needs much more attention.  Maybe we don't want to fiddle with these "quick and dirty" fixes until after we can do more thorough experimentation with different workloads.

One important observation I've made is that being too aggressive with heuristics is more harmful with generational mode than traditional mode.  With traditional mode, being too aggressive means we spend a few more cycles in GC.  With generational mode, it might mean we "starve" old-gen GC, and that might result in degenerated or Full STW pauses.

I'm ok to pull out these heuristic tweaks.  I expect this will all be re-done eventually.  I'm not sure how to deliver the best out-of-box experience to early evaluators.

src/hotspot/share/gc/shenandoah/shenandoahCollectionSet.hpp
kdnilsen: This is a good idea.  I'll work on this change.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
kdnilsen: Fixed.

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
kdnilsen: Yes. Thank you for catching that.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
kdnilsen: This is the only case that was causing me "trouble".  Here are the other scenarios found by grep:
  shenandoahHeap.cpp: 2559: ShenandoahFinalUpdateRefsUpdateRegionStateClosure::heap_region_do()

src/hotspot/share/gc/shenandoah/shenandoahMarkClosures.cpp
kdnilsen: This is a good point.  I guess we can increment age in final_mark for any region that is not placed into the collection set.  In the case that there is sufficient immediate trash that we do not have to evacuate any regions, then all region ages can be incremented.  That would be better.  I'll work on this change.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
kdnilsen: That's a good improvement.  I'll work on this.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: Thank you.  I will change this.

src/hotspot/share/gc/shenandoah/shenandoahCollectionSet.hpp
kdnilsen: This change will be incorporated in next commit.

src/hotspot/share/gc/shenandoah/shenandoahMarkClosures.cpp
kdnilsen: It's a bit more complicated than at first glance.  We don't want to increment age before we choose the collection set because that would influence what goes into the collection set.  Also, we don't want the age of the region to be added to the object age when deciding whether to promote until after all the objects in this region have not been evacuaetd during an aging cycle.  I've made a change that implements the desired behavior.  It is basically some special fixup in the case that we short-circuit evacuation.  What do you think?

src/hotspot/share/gc/shenandoah/shenandoahCollectionSet.inline.hpp
```diff
@@ -61,13 +61,15 @@ size_t ShenandoahCollectionSet::get_immediate_trash() {
   return _immediate_trash;
 }
 
+#ifdef KELVIN_VERBOSE
```
earthling-amzn, line 64: Some instrumentation sneaked in here.

src/hotspot/share/gc/shenandoah/shenandoahCollectionSet.cpp
```diff
@@ -87,6 +87,20 @@ void ShenandoahCollectionSet::add_region(ShenandoahHeapRegion* r) {
   assert(!r->is_humongous(), "Only add regular regions to the collection set");
 
   _cset_map[r->index()] = 1;
+
+  if (ShenandoahHeap::heap()->mode()->is_generational()) {
+    if (ShenandoahHeap::heap()->is_aging_cycle()) {
+      r->decrement_age();
```
earthling-amzn, line 93: Why are we decreasing the age here? Putting the region in the collection set will call `make_cset` on the region which sets the age to zero.

src/hotspot/share/gc/shenandoah/shenandoahThreadLocalData.hpp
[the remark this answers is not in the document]
earthling-amzn: I didn't see where this is called. Is the used? Why would we disable plab promotion?

earthling-amzn (changes requested):

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
kdnilsen: This has been addressed.

src/hotspot/share/gc/shenandoah/shenandoahCollectionSet.cpp
kdnilsen: Sorry.  Thanks for catching this.  This was my "compensation" that I had inserted when I was trying to increment all ages right after final mark.  When I backed out that change, I forgot to remove this.  Gone now.

earthling-amzn (approved):

PR #121, opened 2022-03-02
Reset update watermark for cycles degenerated during evacuation

earthling-amzn:
When an out of memory error occurs during concurrent evacuation the load reference barrier (LRB) stops trying to evacuate objects (this appears to be by design). As the collector transitions to a degenerated cycle, mutators may load references to objects in the collection set. If one of these references is written into a region that was `empty committed` after final mark, the region will have its top at mark start (TAMS) and update watermark (UWM) equal to the bottom of the region. Such regions are effectively skipped during the update reference phase. The degenerated cycle completes with live references to reclaimed memory in the heap, resulting in crashes or corruption.

Rather than re-engineer the oom-during-evac protocol, upon entering a degenerated evacuation phase we again make thread local allocation buffers (TLABs) parsable and reset the UWM for active regions. Now, if the degenerated evacuation phase succeeds, the from-space pointers will be fixed in the update reference phase.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/121/head:pull/121` \
`$ git checkout pull/121`

`$ git checkout pull/121` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/121/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 121`

`$ git pr show -t 121`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/121.diff">https://git.openjdk.java.net/shenandoah/pull/121.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahDegeneratedGC.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahDegeneratedGC.cpp
@@ -176,7 +176,9 @@ void ShenandoahDegenGC::op_degenerated() {
           // it will not have TAMS or UWM updated. Such a region is effectively
           // skipped during update references which can lead to crashes and corruption
           // if the from-space reference is accessed.
-          heap->labs_make_parsable();
+          if (UseTLAB) {
+            heap->labs_make_parsable();
+          }
 
           for (size_t i = 0; i < heap->num_regions(); i++) {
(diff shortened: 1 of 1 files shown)


kdnilsen (approved):
Thanks for this fix.

PR #136, opened 2022-04-28
Balance without cancel

kdnilsen:
This commit does load balancing on remembered set scanning that happens during concurrent marking and during concurrent updating of references.  Experiments show that the concurrency of remembered set marking is now much closer to the number of concurrent GC threads (generally within 5%) whereas before this commit, the level of concurrency was often less than 1/4 of the number of concurrent GC threads.

An additional fix integrated in this commit is to not scan the entirety of humongous object arrays.  We only scan the portions of these arrays that overlay with dirty cards in the remembered set.

---------
### Progress
- [ ] Change must be properly reviewed (1 review required, with at least 1 committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/shenandoah pull/136/head:pull/136` \
`$ git checkout pull/136`

`$ git checkout pull/136` \
`$ git pull https://git.openjdk.java.net/shenandoah pull/136/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 136`

`$ git pr show -t 136`

</details>

<a href="https://git.openjdk.java.net/shenandoah/pull/136.diff">https://git.openjdk.java.net/shenandoah/pull/136.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
@@ -219,6 +219,9 @@ void ShenandoahHeuristics::choose_collection_set(ShenandoahCollectionSet* collec
                      byte_size_in_proper_unit(collection_set->garbage()),
                      proper_unit_for_byte_size(collection_set->garbage()),
                      cset_percent);
+
+  size_t bytes_evacuated = collection_set->get_bytes_reserved_for_evacuation();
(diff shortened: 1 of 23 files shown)


earthling-amzn:
Can we break out the changes to the remembered set scan, the refactoring of collection set selection and the scanning of humongous objects into separate, testable changes? I have some concerns about the performance of this change.

earthling-amzn:
I think we should close it.

PR #142, opened 2022-06-08
Fix fullgc

kdnilsen:
Two errors were recently discovered in the generational-mode implementation of FullGC:

1. There was an off-by-one error which caused full-gc to promote one cycle later than intent.
2. During a post-compaction phase of full-gc, certain old-gen regions were relabeled as young-gen, sometimes resulting in young-gen having usage above its capacity.  When this occurs, further allocations are blocked because available within young-gen is zero (actually is negative).

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Committer](https://openjdk.java.net/bylaws#committer))

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/142/head:pull/142` \
`$ git checkout pull/142`

`$ git checkout pull/142` \
`$ git pull https://git.openjdk.org/shenandoah pull/142/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 142`

`$ git pr show -t 142`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/142.diff">https://git.openjdk.org/shenandoah/pull/142.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
@@ -834,8 +834,8 @@ public:
       r->recycle();
     }
     if (r->is_cset()) {
-      // If generational, leave affiliation alone.
-      r->make_regular_bypass(!is_generational);
+      // Leave afffiliatoin unchanged.
+      r->make_regular_bypass();
     }
     if (r->is_empty_uncommitted()) {
       r->make_committed_bypass();
@@ -1269,7 +1269,11 @@ public:
 
     // Make empty regions that have been allocated into regular
     if (r->is_empty() && live > 0) {
-      r->make_regular_bypass(!is_generational);
+      if (!is_generational) {
+        r->set_affiliation(YOUNG_GENERATION);
+      }
+      // else, generational mode compaction has already established affiliation.
+      r->make_regular_bypass();
     }
 
     // Reclaim regular regions that became empty
@@ -1343,8 +1347,8 @@ void ShenandoahFullGC::compact_humongous_objects() {
         ShenandoahRegionAffiliation original_affiliation = r->affiliation();
         for (size_t c = old_start; c <= old_end; c++) {
           ShenandoahHeapRegion* r = heap->get_region(c);
-          // Consider regions emptied by humongous evacuation to be young
-          r->make_regular_bypass(true);
+          // Leave humongous region affiliation unchanged.
+          r->make_regular_bypass();
           r->set_top(r->bottom());
         }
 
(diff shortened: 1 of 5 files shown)


src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
```diff
@@ -823,11 +829,13 @@ class ShenandoahEnsureHeapActiveClosure: public ShenandoahHeapRegionClosure {
...
   ShenandoahEnsureHeapActiveClosure() : _heap(ShenandoahHeap::heap()) {}
   void heap_region_do(ShenandoahHeapRegion* r) {
+    bool is_generational = _heap->mode()->is_generational();
     if (r->is_trash()) {
       r->recycle();
     }
     if (r->is_cset()) {
-      r->make_regular_bypass();
+      // If generational, leave affiliation alone.
+      r->make_regular_bypass(!is_generational);
```
earthling-amzn, line 838: So - if we are not in generational mode, we will set the affiliation to `YOUNG`, is that necessary when we are not in generational mode? Wouldn't it already be young (or free) in that mode?

And when we are in generational mode, we don't want to change the affiliation. Should we just remove this affiliation changing side-effect of `make_regular_bypass`?

src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
```diff
@@ -1334,7 +1343,8 @@ void ShenandoahFullGC::compact_humongous_objects() {
         ShenandoahRegionAffiliation original_affiliation = r->affiliation();
         for (size_t c = old_start; c <= old_end; c++) {
           ShenandoahHeapRegion* r = heap->get_region(c);
-          r->make_regular_bypass();
+          // Consider regions emptied by humongous evacuation to be young
+          r->make_regular_bypass(true);
```
earthling-amzn, line 1347: If we remove the side-effect of `make_regular_bypass`, we would have something like:
```java
if (r->is_humongous_start() || r->is_humongous_cont()) {
  r->set_affiliation(YOUNG);
}
```

earthling-amzn (changes requested):
Does it make sense to remove the side-effect of `make_regular_bypass`. It's probably a vestige of an out-dated version of the generational support for Full GC.

src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
kdnilsen: You raise good questions here.  I was not sure why make_regular_bypass() wanted to overwrite affiliation with young.  I knew it shouldn't do it in the case I was observing.  I was trying to effect only my use case so as not to break other "use cases".

I'll think a bit harder on this.

src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
kdnilsen: I've implemented and tested a change that removes the change-affiliation side effect of make_regular_bypass.

earthling-amzn (approved):
Looks good - thank you!

PR #144, opened 2022-06-14
Improve humongous remset scan

kdnilsen:
These changes allow humongous objects residing within the remembered set to be scanned by muitiple concurrent GC worker threads, allowing more efficient parallelization of the remembered set scanning effort.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/144/head:pull/144` \
`$ git checkout pull/144`

`$ git checkout pull/144` \
`$ git pull https://git.openjdk.org/shenandoah pull/144/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 144`

`$ git pr show -t 144`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/144.diff">https://git.openjdk.org/shenandoah/pull/144.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -2462,15 +2462,7 @@ private:
               // old-gen heap regions.
               if (r->is_humongous()) {
                 // Need to examine both dirty and clean cards during mixed evac.
-                ShenandoahHeapRegion* start_region = r->humongous_start_region();
-                oop obj = cast_to_oop(start_region->bottom());
-                size_t obj_size = obj->size();
-                HeapWord* end_of_object = start_region->bottom() + obj_size;
-                HeapWord* start_of_scan = r->bottom();
-                size_t card_size_in_words = CardTable::card_size_in_words();
-                size_t words_to_scan = end_of_object - start_of_scan;
-                size_t size_up = (words_to_scan - 1 + card_size_in_words) & ~(card_size_in_words - 1);
-                r->oop_iterate_humongous_slice(&cl, false, start_of_scan, size_up, true, CONCURRENT);
+                r->oop_iterate_humongous(&cl);
               } else {
                 // This is a mixed evacuation.  Old regions that are candidates for collection have not been coalesced
                 // and filled.  Use mark bits to find objects that need to be updated.
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -2461,7 +2461,16 @@ private:
               // old-gen region in the most recent collection set, or if this card holds pointers to other non-specific
               // old-gen heap regions.
               if (r->is_humongous()) {
-                r->oop_iterate_humongous(&cl);
+                // Need to examine both dirty and clean cards during mixed evac.
+                ShenandoahHeapRegion* start_region = r->humongous_start_region();
+                oop obj = cast_to_oop(start_region->bottom());
+                size_t obj_size = obj->size();
+                HeapWord* end_of_object = start_region->bottom() + obj_size;
+                HeapWord* start_of_scan = r->bottom();
+                size_t card_size_in_words = CardTable::card_size_in_words();
+                size_t words_to_scan = end_of_object - start_of_scan;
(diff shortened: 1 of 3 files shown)


kdnilsen:
Before this change, a single thread would scan the entire humongous object, even if the object spans multiple regions.  With this change, different worker threads will be able to scan different parts of the humongous object.  The other improvement here is that when a thread is scanning a humongous object, it only scans the portions of the humongous object that are spanned by DIRTY remembered set cards.  Previously, the entire humongous object was scanned by a single thread.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -580,6 +580,59 @@ void ShenandoahHeapRegion::global_oop_iterate_objects_and_fill_dead(OopIterateCl
   }
 }
 
+// DO NOT CANCEL.  If this worker thread has accepted responsibility for scanning a particular range of addresses, it
+// must finish the work before it can be cancelled.
+void ShenandoahHeapRegion::oop_iterate_humongous_slice(OopIterateClosure* blk, bool dirty_only,
```
earthling-amzn, line 585: Should this also be used during marking scan?

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -2461,7 +2461,16 @@ class ShenandoahUpdateHeapRefsTask : public WorkerTask {
               // old-gen region in the most recent collection set, or if this card holds pointers to other non-specific
               // old-gen heap regions.
               if (r->is_humongous()) {
-                r->oop_iterate_humongous(&cl);
+                // Need to examine both dirty and clean cards during mixed evac.
```
earthling-amzn, line 2464: Sorry, I'm not seeing it. This code looks like it will compute the same `end_of_object` for every region in the humongous object (line 2468), but `start_of_scan` will be `r->bottom`. It seems like this code will have every thread that encounters a region in the humongous object scan the regions between `r` and `end_of_object`. The original method `oop_iterate_humongous` would scan only between `r->bottom` and `r->top`. Maybe we want `end_of_object` to be `r->top` and combine that with only scanning dirty cards?

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -580,6 +580,59 @@ void ShenandoahHeapRegion::global_oop_iterate_objects_and_fill_dead(OopIterateCl
...
+        if (scanner->is_card_dirty(card_index++)) {
+          obj->oop_iterate(blk, MemRegion(start, start + CardTable::card_size_in_words()));
+        }
+        start += CardTable::card_size_in_words();
+      }
+    }
+  } else {
+    // Scan all data, regardless of whether cards are dirty
+    while (num_cards-- > 0) {
+      obj->oop_iterate(blk, MemRegion(start, start + CardTable::card_size_in_words()));
```
earthling-amzn, line 621: Could this just be `obj->oop_iterate(blk, MemRegion(start, start + CardTable::card_size_in_words() * num_cards))`?

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
earthling-amzn: Or, in this case, since we don't care about cards being dirty and we will scan the whole region, could just use the original method?

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -2461,7 +2461,16 @@ class ShenandoahUpdateHeapRefsTask : public WorkerTask {
...
+                ShenandoahHeapRegion* start_region = r->humongous_start_region();
+                oop obj = cast_to_oop(start_region->bottom());
+                size_t obj_size = obj->size();
+                HeapWord* end_of_object = start_region->bottom() + obj_size;
+                HeapWord* start_of_scan = r->bottom();
+                size_t card_size_in_words = CardTable::card_size_in_words();
+                size_t words_to_scan = end_of_object - start_of_scan;
```
kdnilsen, line 2471: With some "last minute" editing, I introduced an error here.  This code should scan to end of region or end of object, whichever comes first.  William is correct to point out that we are rescanning portions of humongous objects multiple times.

earthling-amzn (approved):

PR #145, opened 2022-06-24
Pack old evacuations tightly

kdnilsen:
This pull request includes several improvements to reduce the likelihood of evacuation failures and to reduce the frequency of old-gen collections.  Key contributions:
1. If a PLAB allocation fails, resize the the thread's preferred PLAB size to minimum size and retry the allocation request
2. Place a limit on how large the preferred PLAB size grows for individual threads.  Otherwise a small number of threads can consume the full promotion budget, leaving nothing left for the PLABs required by other threads.
3. If a PLAB promotion fails, and the remaining words within the PLAB is greater than minimum size, that means this is a request to promote a relatively large object.  In this case, do not retire the PLAB.  Instead, use a shared allocation to promote this large object.
4. If a PLAB promotion fails and the remaining words is less than the PLAB minimum size, try to replace the PLAB and then retry the PLAB allocation.  If this still fails, evacuate the object to young-gen using a GCLAB.  Don't try to promote with a shared allocation because that is too "expensive".

This commit has demonstrated fairly significant benefits on certain workloads.  For specjbb2015, we got a 17% improvement in critical jops, from 3093.847 to 3622.  For an extremem 2020 workload,
we improved p99.999 for customer preparation processing by more than four fold, from 607,107 microseconds to 142,286 microseconds, decreasing the number of concurrent GCs from 99 to 37, old GCs from 12 to 7, full GCs from 3 to 0.

---------
### Progress

### Reviewers
 * [William Kemper](https://openjdk.org/census#wkemper) (@earthling-amzn - Committer) ⚠️ Review applies to [d5263dcb](https://git.openjdk.org/shenandoah/pull/145/files/d5263dcbfc222fbafe5ac407735cfff6e07fcb10)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/145/head:pull/145` \
`$ git checkout pull/145`

`$ git checkout pull/145` \
`$ git pull https://git.openjdk.org/shenandoah pull/145/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 145`

`$ git pr show -t 145`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/145.diff">https://git.openjdk.org/shenandoah/pull/145.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
@@ -85,8 +85,9 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
   // we hit max_cset. When max_cset is hit, we terminate the cset selection. Note that in this scheme,
   // ShenandoahGarbageThreshold is the soft threshold which would be ignored until min_garbage is hit.
 
-  size_t max_cset    = (ShenandoahHeap::heap()->get_young_evac_reserve() / ShenandoahEvacWaste);
-  size_t capacity    = ShenandoahHeap::heap()->young_generation()->soft_max_capacity();
+  ShenandoahHeap* heap = ShenandoahHeap::heap();
+  size_t max_cset    = (size_t) (heap->get_young_evac_reserve() / ShenandoahEvacWaste);
+  size_t capacity    = heap->young_generation()->soft_max_capacity();
 
   // As currently implemented, we are not enforcing that new_garbage > min_garbage
   // size_t free_target = (capacity / 100) * ShenandoahMinFreeThreshold + max_cset;
@@ -106,13 +107,17 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
   // particular, regions that have reached tenure age will be sorted into this array before younger regions that contain
   // more garbage.  This represents one of the reasons why we keep looking at regions even after we decide, for example,
   // to exclude one of the regions because it might require evacuation of too much live data.
-
+  bool is_generational = heap->mode()->is_generational();
   for (size_t idx = 0; idx < size; idx++) {
     ShenandoahHeapRegion* r = data[idx]._region;
-    size_t biased_garbage = data[idx]._garbage;
-
-    size_t new_cset    = cur_cset + r->get_live_data_bytes();
-
+    size_t new_cset;
+    if (is_generational && (r->age() >= InitialTenuringThreshold)) {
+      // Entire region will be promoted, This region does not impact young-gen evacuation reserve.  Memory has already
+      // been set aside to hold evacuation results as advance_promotion_reserve.
+      new_cset = cur_cset;
+    } else {
+      new_cset = cur_cset + r->get_live_data_bytes();
+    }
     // As currently implemented, we are not enforcing that new_garbage > min_garbage
     // size_t new_garbage = cur_garbage + r->garbage();
 
@@ -139,8 +144,6 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
       cset->add_region(r);
       cur_cset = new_cset;
       // cur_garbage = new_garbage;
-    } else if (biased_garbage == 0) {
-      break;
     }
   }
 }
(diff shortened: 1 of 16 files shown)


src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
```diff
@@ -106,13 +107,17 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
...
-
+  bool is_generational = heap->mode()->is_generational();
   for (size_t idx = 0; idx < size; idx++) {
     ShenandoahHeapRegion* r = data[idx]._region;
-    size_t biased_garbage = data[idx]._garbage;
-
-    size_t new_cset    = cur_cset + r->get_live_data_bytes();
-
+    size_t new_cset;
+    if (is_generational && (r->age() >= InitialTenuringThreshold)) {
```
earthling-amzn, line 114: Are we changing region affiliation en masse?

kdnilsen: We don't "promote" regions by relabeling region affiliation.  But I found that the computed budget for promotion reserve is way too low in the case that entire regions have reached tenure age.  So I made an adjustment to compute larger promotion reserves when we know that all live data within a collection set region is known to be promoted.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.hpp
```diff
@@ -153,6 +153,8 @@ class ShenandoahHeuristics : public CHeapObj<mtGC> {
 
   virtual void record_requested_gc();
 
+  virtual size_t prioritize_aged_regions(size_t old_available, size_t num_regions, bool preselected_regions[]);
```
earthling-amzn, line 156: Maybe call this `select_aged_regions`?

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
```diff
@@ -247,109 +249,352 @@ void  ShenandoahGeneration::prepare_regions_and_collection_set(bool concurrent)
...
+      //             through evacuation by more than young-gen available.
+      //        iii. Other conditions may be enforced as appropriate for specific heuristics.
+      //
+      //       Note that regions are considered for inclusion in the selection set in order of decreasing amounts of garbage.
+      //       It is possible that a region with a larger amount of garbage will be rejected because it also has a larger
+      //       amount of live data and some region that follows this region in candidate order is included in the collection
+      //       set (because it has less live data and thus can fit within the evacuation limits even though it has less
+      //       garbage).
+
+      size_t young_evacuation_reserve = (young_generation->soft_max_capacity() * ShenandoahEvacReserve) / 100;
```
earthling-amzn, line 396: Should we try to align this more closely with how `ShenandoahFreeSet` behaves? It uses `max_capacity`, for example. It could also tell us exactly how much it really reserved.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -991,6 +1017,15 @@ void ShenandoahHeap::retire_plab(PLAB* plab) {
   }
 }
 
+void ShenandoahHeap::retire_plab(PLAB* plab) {
+  if (!mode()->is_generational()) {
+    plab->retire();
```
earthling-amzn, line 1022: Do we have plabs in the other modes?

earthling-amzn (approved):

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
kdnilsen: PLABs only exist in generational mode

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
kdnilsen: I will change this.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.hpp
kdnilsen: will make this change.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
kdnilsen: I should fix this.  sorry

PR #147, opened 2022-06-27
Adaptive heuristic should take reserved memory into account

earthling-amzn:
Before this change, the heuristic considered the memory available to both the collector and mutators. Since the collector memory is not available to mutators, they may experience an allocation failure before the heuristic could anticipate. With this change, the heuristic looks at the memory available to mutators only. This has the effect of making the heuristic tend to start cycles earlier, thus avoiding degenerated cycles.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/147/head:pull/147` \
`$ git checkout pull/147`

`$ git checkout pull/147` \
`$ git pull https://git.openjdk.org/shenandoah pull/147/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 147`

`$ git pr show -t 147`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/147.diff">https://git.openjdk.org/shenandoah/pull/147.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
@@ -233,6 +233,13 @@ bool ShenandoahAdaptiveHeuristics::should_start_gc() {
   // Make sure the code below treats available without the soft tail.
   size_t soft_tail = max_capacity - capacity;
   available = (available > soft_tail) ? (available - soft_tail) : 0;
+  size_t usable = ShenandoahHeap::heap()->free_set()->available();
+  if (usable < available) {
+    log_debug(gc)("Usable (" SIZE_FORMAT "%s) is less than available (" SIZE_FORMAT "%s)",
+                  byte_size_in_proper_unit(usable), proper_unit_for_byte_size(usable),
+                  byte_size_in_proper_unit(available), proper_unit_for_byte_size(available));
+    available = usable;
+  }
 
   // Track allocation rate even if we decide to start a cycle for other reasons.
   double rate = _allocation_rate.sample(allocated);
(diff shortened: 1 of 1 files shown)


src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
```diff
@@ -233,14 +233,31 @@ bool ShenandoahAdaptiveHeuristics::should_start_gc() {
...
+                  byte_size_in_proper_unit(usable), proper_unit_for_byte_size(usable),
+                  byte_size_in_proper_unit(available), proper_unit_for_byte_size(available));
+    available = usable;
+  }
+
+  // Check if allocation headroom is still okay. This also factors in:
+  //   1. Some space to absorb allocation spikes
+  //   2. Accumulated penalties from Degenerated and Full GC
+  size_t allocation_headroom = available;
+  size_t spike_headroom = capacity / 100 * ShenandoahAllocSpikeFactor;
```
ysramakrishna, line 248: Should you move the associated comments at lines 281-324 up here as well. The comment further below now dangles on its own.

earthling-amzn: Sure, I've moved up the relevant comments.

rkennke:
Is this something we should consider to put directly in upstream?

kdnilsen (approved):

earthling-amzn:
@rkennke - yes, I think it makes sense to make a similar change upstream. We wouldn't be able to apply this change directly because of surrounding generational changes. I will open a ticket and PR for this.

rkennke:
> Just had a look at the heuristic in upstream and it _is_ using freeset available, so this change fixes a regression introduced only on the generational mode branch.

Ok great, then!
Thanks for checking!

PR #148, opened 2022-06-28
Refactor budgeting to make the logic cleaner

kdnilsen:
This commit consolidates computations related to budgeting of reserves for old-gen evacuation and old-gen promotion within shenandoahGeneration::prepare_regions_and_collection_set().  Previously, these computations had been scattered between several functional units, including shenandoahHeuristics::choose_collection_set() and shenandoahOldHeuristics::prime_collection_set().  A previous pull request brought some of the redundant functionality into prepare_regions_and_collection_set().  This pull request removes the redundant computations from the other locations.

---------
### Progress

### Reviewers
 * [Roman Kennke](https://openjdk.org/census#rkennke) (@rkennke - **Reviewer**) ⚠️ Review applies to [be1fb125](https://git.openjdk.org/shenandoah/pull/148/files/be1fb125ea0c2856f38f165d71f64baa2b7d573b)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/148/head:pull/148` \
`$ git checkout pull/148`

`$ git checkout pull/148` \
`$ git pull https://git.openjdk.org/shenandoah pull/148/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 148`

`$ git pr show -t 148`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/148.diff">https://git.openjdk.org/shenandoah/pull/148.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
@@ -436,7 +436,7 @@ void ShenandoahGeneration::adjust_evacuation_budgets(ShenandoahHeap* heap, Shena
     assert(consumed_by_advance_promotion <= (collection_set->get_young_bytes_to_be_promoted() * ShenandoahEvacWaste * 33) / 32,
            "Round-off errors should be less than 3.125%%, consumed by advance: " SIZE_FORMAT ", promoted: " SIZE_FORMAT,
            consumed_by_advance_promotion, (size_t) (collection_set->get_young_bytes_to_be_promoted() * ShenandoahEvacWaste));
-  
+
     collection_set->abandon_preselected();
     if (old_evacuated_committed > old_evacuation_reserve) {
       // This should only happen due to round-off errors when enforcing ShenandoahEvacWaste
@@ -520,7 +520,7 @@ void ShenandoahGeneration::adjust_evacuation_budgets(ShenandoahHeap* heap, Shena
                                                 old_bytes_loaned_for_young_evac + old_bytes_reserved_for_alloc_supplement);
 
     // We experimented with constraining promoted_reserve to be no larger than 4 times the size of previously_promoted,
-    // but this constraint was too limiting, resulting in failure of legitimate promotions.  This was tried before we 
+    // but this constraint was too limiting, resulting in failure of legitimate promotions.  This was tried before we
     // had special handling in place for advance promotion.  We should retry now that advance promotion is handled
     // specially.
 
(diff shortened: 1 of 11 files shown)


src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
```diff
@@ -686,6 +686,7 @@ size_t ShenandoahHeap::capture_old_usage(size_t old_usage) {
 }
 
 void ShenandoahHeap::set_previous_promotion(size_t promoted_bytes) {
+  shenandoah_assert_heaplocked();
```
rkennke, line 689: Is this related and/or relevant?

kdnilsen: We can remove that.  It was "somewhat" related to the historical progression that eventually resulted in this refactoring code.  At one point in my development, I had accidentally tried to set_previous_promotion() without having the heap lock, and the results were unpredictable.

rkennke: Hmm ok. No, leave it there. If it checks an important condition, then it should be useful.

rkennke (approved):

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
```diff
@@ -221,7 +221,352 @@ void ShenandoahGeneration::prepare_gc(bool do_old_gc_bootstrap) {
   }
 }
 
-void  ShenandoahGeneration::prepare_regions_and_collection_set(bool concurrent) {
+void ShenandoahGeneration::compute_evacuation_budgets(ShenandoahHeap* heap, ShenandoahCollectionSet* collection_set,
```
ysramakrishna, line 224: The refactoring of the large chunks of code into the smaller methods is definitely a step in the right direction in making the code easier to follow.

Nit: Would it make the code more readable to consolidate the long sections of comments into a block comment at the top of each of the methods and have much shorter comments in the body of the methods?

src/hotspot/share/gc/shenandoah/shenandoahGeneration.hpp
```diff
@@ -111,6 +111,16 @@ class ShenandoahGeneration : public CHeapObj<mtGC> {
   // Used by concurrent and degenerated GC to reset regions.
   void prepare_gc(bool do_old_gc_bootstrap);
 
+  // Compute evacuation budgets prior to choosing collection set.
```
ysramakrishna, line 114: Do these methods need to be public? They looked more like helper methods in the way they are currently being used.

Another thing to think about is whether the reference parameters in one signature and regular ones in the other are potential sources of confusion or error in the future. I can see why they are like they are in their current use, but might indicate that perhaps a cleaner refactoring of the methods and the state that it uses or modifies, and where that state should reside.

Nothing to do at the moment, but worth thinking about.

ysramakrishna (commented):
Left a few minor syntactic/stylistic comments.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
kdnilsen: Thanks for your suggestion.  Maybe you can do a subsequent pass (in a future PR) to clean up the structure of comments.  I'm sure these can be made better.  Sometimes, a "reader" of the comments has the best perspective on how to make the comments more valuable to other readers.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.hpp
kdnilsen: Agree with suggestion to make compute_evacuation_budgets and adjust_evacuation_budgets private.  Done.

I welcome further refactoring improvements in the future.  Am eager to get this code integrated now (as is) so that we can proceed with other performance improvements.

earthling-amzn (approved):

PR #150, opened 2022-07-05
cli options for young/old ShenandoahMinFreeThreshold

caojoshua:
We currently can set only a single `ShenandoahMinFreeThreshold`, which will run a GC cycle if an allocation space fills up to a certain capacity. This PR allows us to specify separate thresholds for young and old generations.

---------
### Progress

### Reviewers
 * [William Kemper](https://openjdk.org/census#wkemper) (@earthling-amzn - Committer) ⚠️ Review applies to [92adffa0](https://git.openjdk.org/shenandoah/pull/150/files/92adffa0501de7590f87be66217a9fa9fe40c6ac)
 * [Y. Srinivas Ramakrishna](https://openjdk.org/census#ysr) (@ysramakrishna - no project role) ⚠️ Review applies to [92adffa0](https://git.openjdk.org/shenandoah/pull/150/files/92adffa0501de7590f87be66217a9fa9fe40c6ac)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/150/head:pull/150` \
`$ git checkout pull/150`

`$ git checkout pull/150` \
`$ git pull https://git.openjdk.org/shenandoah pull/150/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 150`

`$ git pr show -t 150`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/150.diff">https://git.openjdk.org/shenandoah/pull/150.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
@@ -117,6 +117,13 @@
           "max heap size.")                                                 \
           range(0,100)                                                      \
                                                                             \
+  product(uintx, ShenandoahOldMinFreeThreshold, 5, EXPERIMENTAL,            \
+          "Percentage of free heap memory below which most heuristics "     \
+          "old generation trigger collection independent of other   "       \
+          "triggers. Provides a safety margin for many heuristics. In "     \
+          "percents of (soft) max heap size.")                              \
+          range(0,100)                                                      \
+                                                                            \
   product(uintx, ShenandoahAllocationThreshold, 0, EXPERIMENTAL,            \
           "How many new allocations should happen since the last GC cycle " \
           "before some heuristics trigger the collection. In percents of "  \
--- a/src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
@@ -111,17 +111,18 @@
           range(0,100)                                                      \
                                                                             \
   product(uintx, ShenandoahMinFreeThreshold, 10, EXPERIMENTAL,              \
-          "Percentage of free heap memory below which most heuristics "     \
-          "trigger collection independent of other triggers. Provides "     \
-          "a safety margin for many heuristics. In percents of (soft) "     \
-          "max heap size.")                                                 \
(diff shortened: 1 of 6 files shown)


earthling-amzn (approved):
Looks great! Thank you!

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
```diff
@@ -117,6 +117,13 @@
           "max heap size.")                                                 \
           range(0,100)                                                      \
                                                                             \
+  product(uintx, ShenandoahOldMinFreeThreshold, 5, EXPERIMENTAL,            \
+          "Percentage of free heap memory below which most heuristics "     \
+          "old generation trigger collection independent of other   "       \
```
ysramakrishna, line 122: "Percentage of free old generation memory below which most heuristics will trigger its collection independent of other ..."

Similarly, I'd modify the documentation of "ShenandoahMinFreeThreshold" to state:
"Percentage of free heap memory (or young generation memory, in
 generational mode) below which most heuristics trigger its collection independent of ..."

ASIDE: I was thinking that these may be worth making "manageable" in the future, but I realize now that there is not precedent for that, e.g. in the case of InitiatingHeapOccupancyPercent. I do wonder why not though, since it can allow for some flexibility and external dynamic adaptation of the VM's settings.

ysramakrishna (approved):
Minor documentation suggestions. Otherwise looks good to me. Reviewed!

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
earthling-amzn: I agree with the suggestions for the documentation changes.

The `SoftMaxHeapSize` is "manageable" and that is a factor when checking these thresholds. However, the soft capacity cannot be directly configured for each generation (it is calculated based on `NewSize`, `NewRatio`).

caojoshua: Updating docs based on @ysramakrishna's suggestions. Its not exactly word for word.

> Percentage of free old generation memory below which most heuristics will trigger its collection independent of other ..

I did not include the `will`.

PR #153, opened 2022-07-27
Load balance remembered set scanning

kdnilsen:
This branch divides remembered set scanning into smaller units of work so that multiple cores can more effectively share the workload between them.  The benefit is to reduce concurrent scan remembered set times and to increase the parallelism of this phase.

---------
### Progress

### Reviewers
 * [William Kemper](https://openjdk.org/census#wkemper) (@earthling-amzn - Committer) ⚠️ Review applies to [a806698e](https://git.openjdk.org/shenandoah/pull/153/files/a806698e736230ccf9866ff055d72df65b87ecae)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/153/head:pull/153` \
`$ git checkout pull/153`

`$ git checkout pull/153` \
`$ git pull https://git.openjdk.org/shenandoah pull/153/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 153`

`$ git pr show -t 153`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/153.diff">https://git.openjdk.org/shenandoah/pull/153.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -2470,12 +2470,15 @@ class ShenandoahUpdateHeapRefsTask : public WorkerTask {
 private:
   ShenandoahHeap* _heap;
   ShenandoahRegionIterator* _regions;
+  ShenandoahRegionChunkIterator* _work_chunks;
 
 public:
-  explicit ShenandoahUpdateHeapRefsTask(ShenandoahRegionIterator* regions) :
+  explicit ShenandoahUpdateHeapRefsTask(ShenandoahRegionIterator* regions,
+                                        ShenandoahRegionChunkIterator* work_chunks) :
     WorkerTask("Shenandoah Update References"),
     _heap(ShenandoahHeap::heap()),
-    _regions(regions)
+    _regions(regions),
+    _work_chunks(work_chunks)
   {
   }
 
@@ -2504,60 +2507,24 @@ private:
       assert (update_watermark >= r->bottom(), "sanity");
 
       log_debug(gc)("ShenandoahUpdateHeapRefsTask::do_work(%u) looking at region " SIZE_FORMAT, worker_id, r->index());
+      bool region_progress = false;
       if (r->is_active() && !r->is_cset()) {
         if (!_heap->mode()->is_generational() || (r->affiliation() == ShenandoahRegionAffiliation::YOUNG_GENERATION)) {
           _heap->marked_object_oop_iterate(r, &cl, update_watermark);
+          region_progress = true;
         } else if (r->affiliation() == ShenandoahRegionAffiliation::OLD_GENERATION) {
           if (_heap->active_generation()->generation_mode() == GLOBAL) {
-            _heap->marked_object_oop_iterate(r, &cl, update_watermark);
-          } else {
-            // Old region in a young cycle or mixed cycle.
-            if (!is_mixed) {
-              // This is a young evac..
-              _heap->card_scan()->process_region(r, &cl, true);
-            } else {
-              // This is a _mixed_evac.
-              //
-              // TODO: For _mixed_evac, consider building an old-gen remembered set that allows restricted updating
-              // within old-gen HeapRegions.  This remembered set can be constructed by old-gen concurrent marking
-              // and augmented by card marking.  For example, old-gen concurrent marking can remember for each old-gen
-              // card which other old-gen regions it refers to: none, one-other specifically, multiple-other non-specific.
-              // Update-references when _mixed_evac processess each old-gen memory range that has a traditional DIRTY
-              // card or if the "old-gen remembered set" indicates that this card holds pointers specifically to an
-              // old-gen region in the most recent collection set, or if this card holds pointers to other non-specific
-              // old-gen heap regions.
-              if (r->is_humongous()) {
-                // Need to examine both dirty and clean cards during mixed evac.
-                r->oop_iterate_humongous(&cl);
-              } else {
-                // This is a mixed evacuation.  Old regions that are candidates for collection have not been coalesced
-                // and filled.  Use mark bits to find objects that need to be updated.
-                //
-                // Future TODO: establish a second remembered set to identify which old-gen regions point to other old-gen
-                // regions which are in the collection set for a particular mixed evacuation.
-                HeapWord *p = r->bottom();
-                ShenandoahObjectToOopBoundedClosure<T> objs(&cl, p, update_watermark);
+            // Note that GLOBAL collection is not as effectively balanced as young and mixed cycles.  This is because
+            // concurrent GC threads are parceled out entire heap regions of work at a time and there
+            // is no "catchup phase" consisting of remembered set scanning, during which parcels of work are smaller
+            // and more easily distributed more fairly across threads.
 
-                // Anything beyond update_watermark was allocated during evacuation.  Thus, it is known to not hold
-                // references to collection set objects.
-                while (p < update_watermark) {
-                  oop obj = cast_to_oop(p);
-                  if (ctx->is_marked(obj)) {
-                    objs.do_object(obj);
-                    p += obj->size();
-                  } else {
-                    // This object is not marked so we don't scan it.
-                    HeapWord* tams = ctx->top_at_mark_start(r);
-                    if (p >= tams) {
-                      p += obj->size();
-                    } else {
-                      p = ctx->get_next_marked_addr(p, tams);
-                    }
-                  }
-                }
-              }
-            }
+            // TODO: Consider an improvement to load balance GLOBAL GC.
+            _heap->marked_object_oop_iterate(r, &cl, update_watermark);
+            region_progress = true;
           }
+          // Otherwise, this is an old region in a young or mixed cycle.  Process it during a second phase, below.
+          // Don't bother to report pacing progress in this case.
         } else {
           // Because updating of references runs concurrently, it is possible that a FREE inactive region transitions
           // to a non-free active region while this loop is executing.  Whenever this happens, the changing of a region's
@@ -2574,7 +2541,7 @@ private:
                  affiliation_name(r->affiliation()), r->index());
         }
       }
-      if (ShenandoahPacing) {
+      if (region_progress && ShenandoahPacing) {
         _heap->pacer()->report_updaterefs(pointer_delta(update_watermark, r->bottom()));
       }
       if (_heap->check_cancelled_gc_and_yield(CONCURRENT)) {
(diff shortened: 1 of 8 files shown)


src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.cpp
```diff
@@ -87,29 +87,215 @@ void ShenandoahDirectCardMarkRememberedSet::merge_overreach(size_t first_cluster
...
+    group_span /= 2;
+  }
+  // Not necessary, but keeps things tidy
+  for (size_t i = _num_groups; i < _maximum_groups; i++) {
+    _region_index[i] = 0;
+    _group_offset[i] = 0;
+  }
+}
+
+ShenandoahRegionChunkIterator::ShenandoahRegionChunkIterator(ShenandoahHeap* heap, size_t worker_count) :
```
earthling-amzn, line 274: Can we use [constructor delegation](https://en.wikipedia.org/wiki/C++11#Object_construction_improvement) here to remove some duplicated code? Or could we just have clients always pass in the heap reference? One constructor is missing an assert.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
```diff
@@ -1002,20 +1022,97 @@ class ShenandoahScanRemembered: public CHeapObj<mtGC> {
   void roots_do(OopIterateClosure* cl);
 };
 
+typedef struct ChunkOfRegion {
```
earthling-amzn, line 1025: Why `typedef`? Could this just be `struct ShenandoahRegionChunk`?

earthling-amzn (changes requested):
Why do we want to group chunks? Why not let the worker threads grab chunks on a first-come-first-serve basis?

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
kdnilsen: Ok.  I've eliminated the typedef and replaced everywhere with struct ShenandoahRegionChunk.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.cpp
kdnilsen: Thanks for this suggestion.  I've removed body of one constructor and replaced with constructor delegation.

kdnilsen:
The reason to "group chunks" is to reduce the need for synchronized coordination between worker threads.  Dividing work into entire ShenandoahHeapRegions seems to work fairly well for some workloads.  The idea is to start out each remembered set scan effort with work assignments that are a "large number" chunks.  Then, as we get closer to the end of the total scanning effort, we began dividing the worker into smaller and smaller assignments.  This way, a thread that gets an early "dense" work assignment might only complete that single heavy assignment while other worker threads can each complete a large number of lighter worker assignments.

kdnilsen:
I should also emphasize that there are other benefits besides reduced coordination overhead to scanning larger numbers of chunks at a time.  For example, the code that Ramki is working on can cache the starting location of a large object that spans multiple chunks most easily if the starting location is within the range of its current work assignment.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -2574,30 +2541,149 @@ class ShenandoahUpdateHeapRefsTask : public WorkerTask {
...
+                // Need to examine both dirty and clean cards during mixed evac.
+                r->oop_iterate_humongous_slice(&cl, false, start_of_range, assignment._chunk_size, true, CONCURRENT);
+              }
+            } else {
+              // Since this is mixed evacuation, old regions that are candidates for collection have not been coalesced
+              // and filled.  Use mark bits to find objects that need to be updated.
+              //
+              // Future TODO: establish a second remembered set to identify which old-gen regions point to other old-gen
+              // regions which are in the collection set for a particular mixed evacuation.
+              if (start_of_range < end_of_range) {
```
earthling-amzn, line 2593: Before this change, we had update references and marking share most of the card scanning code. Is it no longer possible to share the card scanning code as before? Could we put some of this code in the scanner?

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.cpp
```diff
@@ -87,29 +87,195 @@ void ShenandoahDirectCardMarkRememberedSet::merge_overreach(size_t first_cluster
...
+  size_t cumulative_group_span = 0;
+  size_t current_group_span = _first_group_chunk_size * _group_size;
+  size_t smallest_group_span = _smallest_chunk_size * _group_size;
+  while (unspanned_heap_size > 0) {
+    if (current_group_span <= unspanned_heap_size) {
+      unspanned_heap_size -= current_group_span;
+      num_chunks += _group_size;
+      num_groups++;
+
+      if (num_groups >= _num_groups) {
```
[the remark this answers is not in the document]
kdnilsen: I'm adding a comment and changing num_groups local variable to spanned_groups.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.cpp
```diff
@@ -87,29 +87,195 @@ void ShenandoahDirectCardMarkRememberedSet::merge_overreach(size_t first_cluster
...
   }
 }
+
+
+size_t ShenandoahRegionChunkIterator::calc_group_size() {
+  // First group does roughly half of heap, one region at a time.
+  // Second group does roughly one quarter of heap, half of a region at a time, and so on.
+  // Last group does the remnant of heap, one _smallest_chunk_size at a time.
+  // Round down.
+  return _heap->num_regions() / 2;
```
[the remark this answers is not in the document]
kdnilsen: I've rewritten the comment here in an attempt to make the intention more clear.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
kdnilsen: This is a good question.  Before, we had "if (!is_mixed) { heap->card_scan->process_region(); }  Now, in that case (the code is below), we have "else { ...  scanner->process_region_slice(); ...  }  This sharing of code still exists, but it is made a little more complicated because we don't always process entire regions at once, and I reordered the control flow.

The is_mixed case was always handled locally (unshared code) with explicit iteration over objects.  This is still handled locally, but it is also more complex in the new version of the code because we are working with "slices" of regions at a time.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.cpp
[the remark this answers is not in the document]
earthling-amzn: I get it. I think it would be more clear if this method were called `initial_group_size` as the reduction of subsequent group sizes does not happen in this method.

earthling-amzn (approved):

ysramakrishna:
> This branch divides remembered set scanning into smaller units of work so that multiple cores can more effectively share the workload between them. The benefit is to reduce concurrent scan remembered set times and to increase the parallelism of this phase.
> 

Are any specific or general performance numbers available to share for this change? Thanks!

kdnilsen:
> Are any specific or general performance numbers available to share for this change? Thanks!

In a 20-minute Extremem workload comparing mainline against this branch, the average time for remembered-set scanning was reduced from 28,634 microseconds to 21,810 microseconds, an improvement of 17.55%.  This was out of n = 1893 scans for the balanced branch and n = 1925 for the mainline.  I don't have a full analysis of why n is different between the two workloads.  I have seen previously that unbalanced scanning of remembered set sometimes has much more significant impact on GC completion times, and can result in allocation failures and degeneration.  On this workload, I don't see evidence of this.  But shorter GC times perhaps means less floating garbage, and that might be one reason we needed fewer total GC cycles.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.cpp
```diff
@@ -87,29 +87,200 @@ void ShenandoahDirectCardMarkRememberedSet::merge_overreach(size_t first_cluster
...
+  struct ShenandoahRegionChunk assignment;
+  bool has_work = _work_list->next(&assignment);
+  while (has_work) {
+#ifdef ENABLE_REMEMBERED_SET_CANCELLATION
+    // This check is currently disabled to avoid crashes that occur
+    // when we try to cancel remembered set scanning
+    if (heap->check_cancelled_gc_and_yield(_is_concurrent)) {
+      return;
+    }
+#endif
```
ysramakrishna, line 127: The crashes are likely because you are pulling an assignment off the worklist, and if you see a cancellation, you are dropping it on the floor and returning.

I think the check should be at the end of the loop after the assignment pulled off the work list has been processed but before the next one is pulled.

Here's a diff that should work:

```
diff --git a/src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.cpp b/src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.cpp
index 2d7cbbb97dd..03f647f23b2 100644
--- a/src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.cpp
@@ -81,15 +81,7 @@ void ShenandoahScanRememberedTask::do_work(uint worker_id) {
   // set up thread local closure for shen ref processor
   _rp->set_mark_closure(worker_id, &cl);
   struct ShenandoahRegionChunk assignment;
-  bool has_work = _work_list->next(&assignment);
-  while (has_work) {
-#ifdef ENABLE_REMEMBERED_SET_CANCELLATION
-    // This check is currently disabled to avoid crashes that occur
-    // when we try to cancel remembered set scanning
-    if (heap->check_cancelled_gc_and_yield(_is_concurrent)) {
-      return;
-    }
-#endif
+  while (_work_list->next(&assignment)) {
     ShenandoahHeapRegion* region = assignment._r;
     log_debug(gc)("ShenandoahScanRememberedTask::do_work(%u), processing slice of region "
                   SIZE_FORMAT " at offset " SIZE_FORMAT ", size: " SIZE_FORMAT,
@@ -101,13 +93,15 @@ void ShenandoahScanRememberedTask::do_work(uint worker_id) {
       assert(clusters * cluster_size == assignment._chunk_size, "Chunk assignments must align on cluster boundaries");
       HeapWord* end_of_range = region->bottom() + assignment._chunk_offset + assignment._chunk_size;
 
-      // During concurrent mark, region->top() equals TAMS with respect to the current young-gen pass.  */
+      // During concurrent mark, region->top() equals TAMS with respect to the current young-gen pass.
       if (end_of_range > region->top()) {
         end_of_range = region->top();
       }
       scanner->process_region_slice(region, assignment._chunk_offset, clusters, end_of_range, &cl, false, _is_concurrent);
     }
-    has_work = _work_list->next(&assignment);
+    if (heap->check_cancelled_gc_and_yield(_is_concurrent)) {
+      return;
+    }
   }
 }
```

ysramakrishna (commented):
Noticed this when playing with the code. I didn't review the rest of the changes, but I think this might be a possible explanation (and a potential fix) for the crashes you were seeing.

I have the fix running SPECjbb successfully, but have not checked if your original code would have elicited the crash you saw with SPECjbb.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.cpp
kdnilsen: Thanks very much for figuring out this problem.  I'd like to address this change in a different pull request to keep the history cleaner.

ysramakrishna: 👍

ysramakrishna: I left a SPECjbb run overnight with the changes, but alas it still does crash (slowdebug mode) during a deoptimization, presumably because of a bad oop on the stack. I'll file a ticket to track the issue, but it does seem like there's some other issue besides the one pointed out above.

PR #156, opened 2022-08-05
Fix budgeting assertion

kdnilsen:
An assertion failure revealed problems with evacuation budgeting computations.  Three changes are reflected in this pull request:

1. Memory set aside for the promotion reserve was not excluded from the allocation supplement
2. When the old evacuation reserve is larger than the memory consumed by old evacuation collection set, we loan the excess memory to allocation supplement to make sure this memory is not consumed by promotions.  The number of regions set aside from this reserve for the allocation supplement must be rounded down from the total excess memory.  (Before this patch, it was rounded up.)
3. The assertion test that was failing needed a <= comparison instead of a < comparison.

---------
### Progress

### Reviewers
 * [William Kemper](https://openjdk.org/census#wkemper) (@earthling-amzn - Committer) ⚠️ Review applies to [7d256dba](https://git.openjdk.org/shenandoah/pull/156/files/7d256dbafb6285684f3961ce538b3576ea25fede)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/156/head:pull/156` \
`$ git checkout pull/156`

`$ git checkout pull/156` \
`$ git pull https://git.openjdk.org/shenandoah pull/156/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 156`

`$ git pr show -t 156`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/156.diff">https://git.openjdk.org/shenandoah/pull/156.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
@@ -123,7 +123,11 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
             add_region = true;
             old_cur_cset = new_cset;
           }
-        } else if (cset->is_preselected(idx)) {
+        } else if (cset->is_preselected(r->index())) {
+#undef KELVIN_VERBOSE
+#ifdef KELVIN_VERBOSE
+          printf("GLOBAL: Preselected region " SIZE_FORMAT " added to cset, age is: %d\n", r->index(), r->age());
+#endif
           assert(r->age() >= InitialTenuringThreshold, "Preselected regions must have tenure age");
           // Entire region will be promoted, This region does not impact young-gen or old-gen evacuation reserve.
           // This region has been pre-selected and its impact on promotion reserve is already accounted for.
@@ -170,8 +174,12 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
         size_t region_garbage = r->garbage();
         size_t new_garbage = cur_young_garbage + region_garbage;
         bool add_region = false;
+
         if (!r->is_old()) {
-          if (cset->is_preselected(idx)) {
+          if (cset->is_preselected(r->index())) {
+#ifdef KELVIN_VERBOSE
+            printf("YOUNG or MIXED: Preselected region " SIZE_FORMAT " added to cset, age is: %d\n", r->index(), r->age());
+#endif
             assert(r->age() >= InitialTenuringThreshold, "Preselected regions must have tenure age");
             // Entire region will be promoted, This region does not impact young-gen evacuation reserve.  Memory has already
             // been set aside to hold evacuation results as advance_promotion_reserve.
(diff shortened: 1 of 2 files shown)


ysramakrishna (approved):
LGTM!

kdnilsen:
There's another assertion failure related to evacuating budgeting that has turned up.  After investigating this issue, I want to make one other change to this code before we merge it.

earthling-amzn (approved):

src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
```diff
@@ -123,12 +123,18 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
...
           // Entire region will be promoted, This region does not impact young-gen or old-gen evacuation reserve.
           // This region has been pre-selected and its impact on promotion reserve is already accounted for.
           add_region = true;
           cur_young_garbage += r->garbage();
-        } else {
+          // Since all live data in this region is being evacuated from young-gen, it is as if this memory
+          // is garbage insofar as young-gen is concerned.  Counting this as garbage reduces the need to
+          // reclaim highly utilized young-gen regions just for the sake of finding min_garbage to reclaim
+          // within youn-gen memory
+          cur_young_garbage += r->get_live_data_bytes();
```
ysramakrishna, line 136: Isn't `r->garbage() + r->get_live_data_bytes()` equal to `r->used()`?

I am not sure I understand the reasoning here. I would much rather we named the `cur_young_garbage` variable more correctly if it is counting the amount of young regions that are in some sense being "removed from the young generation" (if indeed that's how to interpret it).

ysramakrishna: Might be a good idea to describe at line 97, what `cur_young_garbage` will track as you process regions and build the collection set. Similarly, young_curset etc.

I see three separate outer if/else statements covering respectively the cases of generational/global, generational/not global, and non-generational cases. Would it read more easily and make for more easily maintained code to refactor into 3 separate cset builders called for each of the three cases, if necessary from the top level method (but even better if called from different contexts to specialize early hoisting the check on the type of collection up into the highest caller that has the information, rather than having all of them in one single, somewhat long and somewhat unwieldy method?

ysramakrishna (commented):
Some minor comments on structure of the method and a possible refactoring, and on naming & documenting of some variables.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
kdnilsen: Thanks for these suggestions.  Adding a comment and "folding" those two addition operations into one.

kdnilsen: Not refactoring the body of method for now, but that can be done at a future time.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
```diff
@@ -123,12 +123,18 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
             add_region = true;
             old_cur_cset = new_cset;
           }
-        } else if (r->age() >= InitialTenuringThreshold) {
+        } else if (cset->is_preselected(r->index())) {
```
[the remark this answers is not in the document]
kdnilsen: Thanks.  This is a good idea.  Let's put this into a future pr...

PR #159, opened 2022-08-27
Use only up to ConcGCThreads for concurrent RS scanning.

ysramakrishna:
The concurrent RS scanning phase was using (up to) ParallelGCThreads, rather than ConcGCThreads as was the original intention for concurrent phases.


kdnilsen (approved):
Thanks for putting this together.

earthling-amzn (approved):

PR #155, opened 2022-08-02
Use os::malloc instead of malloc

earthling-amzn:
This fixes a compiler error with gcc-10.3.0 used by github actions

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Committer](https://openjdk.org/bylaws#committer))

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/155/head:pull/155` \
`$ git checkout pull/155`

Update a local copy of the PR: \
`$ git checkout pull/155` \
`$ git pull https://git.openjdk.org/shenandoah pull/155/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 155`

View PR using the GUI difftool: \
`$ git pr show -t 155`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/shenandoah/pull/155.diff">https://git.openjdk.org/shenandoah/pull/155.diff</a>

</details>

PR #160, opened 2022-09-02
Fix off-by-one error when verifying object registrations

earthling-amzn:
This change originally started with an effort to fix a build error caused by using the global `malloc` and `free` calls. Replacing these calls with hotspot idioms caused remembered set verification errors. The verification errors were, ultimately, due to the verification code reading past the end of an array. The `NEW_C_HEAP_ARRAY` macro tacks a bit of information onto the allocated array for the purpose of tracking native memory use (NMT). This was enough to change the behavior when the verifier read past the end of the array. This PR subsumes https://github.com/openjdk/shenandoah/pull/155 and fixes the root cause of the verification errors. This change also removes the unused `overreach_map` as this was also using global `malloc` and `free` calls.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/160/head:pull/160` \
`$ git checkout pull/160`

`$ git checkout pull/160` \
`$ git pull https://git.openjdk.org/shenandoah pull/160/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 160`

`$ git pr show -t 160`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/160.diff">https://git.openjdk.org/shenandoah/pull/160.diff</a>

</details>


src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
```diff
@@ -419,9 +400,9 @@ ShenandoahScanRemembered<RememberedSet>::verify_registration(HeapWord* address,
     //        cannot use card_index_for_addr(base_addr + offset) because it asserts arg < end of whole heap
     size_t end_card_index = index + offset / CardTable::card_size_in_words();
 
-    if (end_card_index > index) {
+    if (end_card_index > index && end_card_index < _rs->last_valid_index()) {
```
kdnilsen, line 403: If _rs->last_valid_index() is actually a valid index, why couldn't we make this test <=?  Or should we rename the last_valid_index() to be index_beyond_end()?

earthling-amzn: Good catch. I will test that.

tstuefe (commented):
Just some drive-by-comments, I'm not a shenandoah developer.

Note that the trailing information after os::malloc/NEW_C_HEAP_ARRAY confusing your reader are canaries we set in NMT to catch overflow errors.

Cheers, Thomas

earthling-amzn:
Appreciate you stopping by @tstuefe , thanks for the review.

kdnilsen (approved):

PR #170, opened 2022-11-14
Change affiliation representation

kdnilsen:
With generational mode of Shenandoah, each region is associated with OLD, YOUNG, or FREE.  During certain marking and update-refs activities, the region affiliation is frequently consulted.  In the original implementation, the region affiliation was stored in a field of the ShenandoahHeapRegion object.  In this code, we maintain a separate array, indexed by region number, to represent the affiliation of each region.  This saves one level of indirection and improves cache locality for looking up the affiliation of each region.

Measurements show significant improvement in throughput.  One workload that was configured to perform back-to-back old-gen collections was able to increase the frequency of old-gen collections by almost 5 fold.  With a 20-minute Extremem workload using a 48G heap, 20G old-gen, the P95 latency improvement was 0.54% (2.395 ms) and the P99.999 latency improvement was 58.21% (29.195 ms) in comparison to the implementation before this patch.

---------
### Progress

### Reviewers
 * [Y. Srinivas Ramakrishna](https://openjdk.org/census#ysr) (@ysramakrishna - Author) ⚠️ Review applies to [1033b0cd](https://git.openjdk.org/shenandoah/pull/170/files/1033b0cdc5f00af5261f96d5e945c3b9b9ac4c13)
 * [William Kemper](https://openjdk.org/census#wkemper) (@earthling-amzn - Committer) ⚠️ Review applies to [1033b0cd](https://git.openjdk.org/shenandoah/pull/170/files/1033b0cdc5f00af5261f96d5e945c3b9b9ac4c13)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/170/head:pull/170` \
`$ git checkout pull/170`

`$ git checkout pull/170` \
`$ git pull https://git.openjdk.org/shenandoah pull/170/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 170`

`$ git pr show -t 170`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/170.diff">https://git.openjdk.org/shenandoah/pull/170.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -358,7 +358,7 @@ jint ShenandoahHeap::initialize() {
   } else {
     _affiliations = NULL;
   }
-  _Free_set = new ShenandoahFreeSet(this, _num_regions);
+  _free_set = new ShenandoahFreeSet(this, _num_regions);
 
   {
     ShenandoahHeapLocker locker(lock());
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -353,10 +353,12 @@ jint ShenandoahHeap::initialize() {
   }
 
   _regions = NEW_C_HEAP_ARRAY(ShenandoahHeapRegion*, _num_regions, mtGC);
+  _affiliations = NEW_C_HEAP_ARRAY(uint8_t, _num_regions, mtGC);
   _free_set = new ShenandoahFreeSet(this, _num_regions);
 
   {
     ShenandoahHeapLocker locker(lock());
+
 
     for (size_t i = 0; i < _num_regions; i++) {
       HeapWord* start = (HeapWord*)sh_rs.base() + ShenandoahHeapRegion::region_size_words() * i;
@@ -369,6 +371,8 @@ jint ShenandoahHeap::initialize() {
       _marking_context->initialize_top_at_mark_start(r);
       _regions[i] = r;
       assert(!collection_set()->is_in(i), "New region should not be in collection set");
+
+      _affiliations[i] = ShenandoahRegionAffiliation::FREE;
     }
 
     // Initialize to complete
@@ -509,6 +513,7 @@ ShenandoahHeap::ShenandoahHeap(ShenandoahCollectorPolicy* policy) :
   _heap_region_special(false),
   _num_regions(0),
   _regions(NULL),
+  _affiliations(NULL),
   _update_refs_iterator(this),
   _alloc_supplement_reserve(0),
   _promoted_reserve(0),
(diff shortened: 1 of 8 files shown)


src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -353,11 +353,13 @@ jint ShenandoahHeap::initialize() {
   }
 
   _regions = NEW_C_HEAP_ARRAY(ShenandoahHeapRegion*, _num_regions, mtGC);
+  _affiliations = NEW_C_HEAP_ARRAY(uint8_t, _num_regions, mtGC);
```
earthling-amzn, line 356: Should we only initialize this for generational mode? or are there calls to check affiliation in the other modes too?

kdnilsen: Good point.  I think it's only needed for generational mode.

ysramakrishna (approved):
The change looks good, modulo William's comment. (I also wonder if it might make sense to have a GenerationalShenandoahHeap derived from ShenandoahHeap that might make some of this kind of separation a bit cleaner. But that would be a bigger change for the future if one thought it was worthwhile to do a clean separation of the code for the two collectors.)

A high level question, unrelated to the changes here. 

I imagine that we never consult a region's affiliation while it may be concurrently subject to update? (or when that happens, that any resulting race is benign)? Would it be worth adding a brief comment to that effect somewhere? (may be where the _affiliation array is declared.)

I am guessing the affiliation state graph looks like:

free -> {young, old}
young -> {old, free}
old -> free

Would this be worth adding as a comment may be where set_affiliation() is implemented?

None of these need to be done right away, but thought I'd leave these thoughts in the review anyway.

earthling-amzn (approved):

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
kdnilsen: After further investigation, I see that we need _affiliation for non-generational mode to keep track of FREE status.  I've added a comment to this effect.

kdnilsen:
In the most recently committed refinements, I have added comments and assertions to clarify the use of heap lock and safepoints as mechanisms to assure coherency of region affiliations.

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
```diff
@@ -550,10 +550,102 @@ inline bool ShenandoahHeap::clear_old_evacuation_failure() {
...
+                                                        ShenandoahRegionAffiliation new_affiliation) {
+  // A lock is required when changing from FREE to NON-FREE.  Though it may be possible to elide the lock when
+  // transitioning from in-use to FREE, the current implementation uses a lock for this transition.  A lock is
+  // not required to change from YOUNG to OLD (i.e. when promoting humongous region).
+  //
+  //         new_affiliation is:     FREE   YOUNG   OLD
+  //  orig_affiliation is:  FREE      X       L      L
+  //                       YOUNG      L       X
+  //                         OLD      L       X      X
+  //  X means state transition won't happen (so don't care)
```
ysramakrishna, line 619: Nits:
1. Instead of the blank entry in the cell for Young->Old, may be use `N` and replace `Blank` in the legend below?
2. If we don't expect some transitions (and believe those are incorrect), it might make sense to assert that in `set_affiliation()`?
3. Further below, it looks like all current (and expected) uses of `assert_lock_for_affiliation` will be from assertion checking code in non-release builds. In that case, you can avoid the `#ifdef ASSERT` verbage at call sites by declaring the method as `PRODUCT_RETURN` which should elide the empty method in release builds.

PR #171, opened 2022-11-17
Do not apply evacuation budgets in non-generational mode

earthling-amzn:
CSet selection for non-generational mode was using a value configured on the young generation. In non-generational modes we do not maintain all the attributes of the young generation (instead we maintain the _global_ generation). This caused many cycles in which the maximum CSet was zero (or close to it). This, in turn, caused the collector to run much more frequently (approximately 3x on specjbb) which caused severe performance regression in critical jops.

I'm not sure why the diff algorithm is struggling so much with these changes. I pulled up the `mode()->is_generational()` out of `compute_evacuation_budgets` and `adjust_evacuation_budgets` into the caller and replaced the check with an assert (re-indenting the code in the method).

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/171/head:pull/171` \
`$ git checkout pull/171`

`$ git checkout pull/171` \
`$ git pull https://git.openjdk.org/shenandoah pull/171/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 171`

`$ git pr show -t 171`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/171.diff">https://git.openjdk.org/shenandoah/pull/171.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
@@ -216,193 +216,184 @@ void ShenandoahGeneration::prepare_gc() {
 void ShenandoahGeneration::compute_evacuation_budgets(ShenandoahHeap* heap, bool* preselected_regions,
                                                       ShenandoahCollectionSet* collection_set,
                                                       size_t &consumed_by_advance_promotion) {
+  assert(heap->mode()->is_generational(), "Only generational mode uses evacuation budgets.");
   size_t region_size_bytes = ShenandoahHeapRegion::region_size_bytes();
   size_t regions_available_to_loan = 0;
   size_t minimum_evacuation_reserve = ShenandoahOldCompactionReserve * region_size_bytes;
   size_t old_regions_loaned_for_young_evac = 0;
   consumed_by_advance_promotion = 0;
-  if (heap->mode()->is_generational()) {
-    ShenandoahGeneration* old_generation = heap->old_generation();
-    ShenandoahYoungGeneration* young_generation = heap->young_generation();
-    size_t avail_evac_reserve_for_loan_to_young_gen = 0;
-    size_t old_evacuation_reserve = 0;
-    size_t num_regions = heap->num_regions();
 
-    // During initialization and phase changes, it is more likely that fewer objects die young and old-gen
-    // memory is not yet full (or is in the process of being replaced).  During these times especially, it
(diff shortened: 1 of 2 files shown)


kdnilsen (commented):
Thanks for this fix.  I think I may have introduced the bad code.

ysramakrishna:
> I'm not sure why the diff algorithm is struggling so much with these changes. 

Are the diff struggles in git (I assume), or in github (I assume not). I have seen git have issues with indentation that caused issues with merges in like manner recently. :-(

I wonder if there's a git option for indentation that would rended this diff better?

> ```
> --indent-heuristic
> Enable the heuristic that shifts diff hunk boundaries to make patches easier to read. This is the default.
> 
> --no-indent-heuristic
> Disable the indent heuristic.
> ```

earthling-amzn:
The diff looked hairy even in `p4merge`.

ysramakrishna:
Interesting, I wonder if this would make any diff (pun not intended):

```
--diff-algorithm={patience|minimal|histogram|myers}
Choose a diff algorithm. The variants are as follows:

default, myers
The basic greedy diff algorithm. Currently, this is the default.

minimal
Spend extra time to make sure the smallest possible diff is produced.

patience
Use "patience diff" algorithm when generating patches.

histogram
This algorithm extends the patience algorithm to "support low-occurrence common elements".

For instance, if you configured the diff.algorithm variable to a non-default value and want to use the default one, then you have to use --diff-algorithm=default option.
```

earthling-amzn:
I think my editor applied some bonus formatting... One second, I will fix up the patch.

earthling-amzn:
Try hiding whitespace on github - looks much better to me:

image

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
```diff
@@ -791,40 +779,40 @@ void ShenandoahGeneration::prepare_regions_and_collection_set(bool concurrent) {
       ShenandoahFinalMarkUpdateRegionStateClosure old_cl(nullptr);
       heap->old_generation()->parallel_heap_region_iterate(&old_cl);
     }
-
-    heap->assert_pinned_region_status();
```
earthling-amzn, line 795: This is called again immediately inside collection set selection so I removed this invocation.

kdnilsen (approved):

PR #173, opened 2022-11-18
Load balance remset scan

kdnilsen:
Prior to this change, the initial group of remembered set assignments was given to worker threads one entire region at a time.  We found that with large region sizes (e.g. 16 MiB and above), this resulted in too much imbalance in the work performed by individual threads.  A few threads assigned to scan 16 MiB regions with high density of "interesting pointers" were still scanning after all other worker threads finished their scanning efforts.

This change caps the maximum assignment size for worker threads at 4 MiB.  This results in better distribution of efforts between multiple concurrent threads.  With 13 worker threads and 16 MiB heap regions, we observe the following benefits on an Extremem workload (46_064 MiB heap size, 27_648 MiB new size):

Latency for customer preparation processing improved by 0.79% for P50, 2.26% for P95, 8.21% for p99, 28.17% for p99.9, 86.59% for p99.99, 86.77% for p99.999.  The p100 response improved only slightly, by 1.99%.

Average time for concurrent remembered set marking scan improved by 1.92%.  The average time for concurrent update refs time, which includes remembered set scanning, improved by 1.72%.

---------
### Progress

### Reviewers
 * [William Kemper](https://openjdk.org/census#wkemper) (@earthling-amzn - Committer) ⚠️ Review applies to [9704b348](https://git.openjdk.org/shenandoah/pull/173/files/9704b3487d2e4383fadcfead4cce2b3990ff1f5b)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/173/head:pull/173` \
`$ git checkout pull/173`

`$ git checkout pull/173` \
`$ git pull https://git.openjdk.org/shenandoah pull/173/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 173`

`$ git pr show -t 173`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/173.diff">https://git.openjdk.org/shenandoah/pull/173.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.cpp
@@ -109,28 +109,37 @@ void ShenandoahScanRememberedTask::do_work(uint worker_id) {
   }
 }
 
-size_t ShenandoahRegionChunkIterator::calc_group_size() {
-  // The group size s calculated from the number of regions.  Every group except the last processes the same number of chunks.
-  // The last group processes however many chunks are required to finish the total scanning effort.  The chunk sizes are
+size_t ShenandoahRegionChunkIterator::calc_regular_group_size() {
+  // The group size is calculated from the number of regions.  Suppose the entire heap size is N.  The first group processes
+  // N/2 of total heap size.  The second group processes N/4 of total heap size.  The third group processes N/2 of total heap
+  // size, and so on.  Note that N/2 + N/4 + N/8 + N/16 + ...  sums to N if expanded to infinite terms.
+  //
+  // The normal group size is the number of regions / 2, with the normal 
+
+  // In the case that the region_size_words is greater than _maximum_chunk_size_words, the first group_size is 
+  // larger than the normal group size because each chunk in the group will be smaller than the region size.
+  //
+  // The last group also has more than the normal entries because it finishes the total scanning effort.  The chunk sizes are
   // different for each group.  The intention is that the first group processes roughly half of the heap, the second processes
   // a quarter of the remaining heap, the third processes an eight of what remains and so on.  The smallest chunk size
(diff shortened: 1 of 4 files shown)


src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.cpp
```diff
@@ -168,29 +177,46 @@ size_t ShenandoahRegionChunkIterator::calc_total_chunks() {
...
   size_t unspanned_heap_size = _heap->num_regions() * region_size_words;
   size_t num_chunks = 0;
-  size_t spanned_groups = 0;
   size_t cumulative_group_span = 0;
-  size_t current_group_span = _first_group_chunk_size * _group_size;
-  size_t smallest_group_span = _smallest_chunk_size * _group_size;
+  size_t current_group_span = _first_group_chunk_size_b4_rebalance * _regular_group_size;
+  size_t smallest_group_span = _smallest_chunk_size_words * _regular_group_size;
+
+  // Tyhe first group gets special handling because the first chunk size can be no larger than _largets_chunk_size_words
```
earthling-amzn, line 184: `s/Thye/The`

earthling-amzn:
I could use an overview here. I understand the advantage of lowering the maximum chunk size, but what is the advantage of using successively smaller chunks?

earthling-amzn (approved):

kdnilsen:
I'll add a comment to explain the rationale behind successively smaller chunks.  General idea is that as we get closer to the end of the total effort, we want to be more careful to avoid giving one of the worker threads a disproportionately large amount of work to do.  Early in the total effort, it's ok for one thread to get a larger assignment than the others.  In this case, the thread with the larger effort will chew away on that large assignment while all the other threads repeatedly receive and finish work assignments that can be completed more quickly.

PR #184, opened 2022-12-15
JDK-8298597 : HdrSeq: support for a merge() method

ysramakrishna:
A merge functionality on stats (distributions) was needed for the remembered set scan that I was using in some companion work. This PR implements a first cut at that, which is sufficient for our first (and only) use case.

Unfortunately, for expediency, I am deferring work on decaying statistics, as a result of which users that want decaying statistics will get NaNs instead (or trigger guarantees).

In the sort term, before I open this draft for review, I'll:

- [x] add tests
- [x] ensure that if a merge action has been taken on a distribution, then any attempt to access a decayed statistic causes an error
- [x] open a linked ticket to take care of the decayed statistics

An important goal here was to have an API that would be efficient and correct. The API shape may change when we have considered how to handle decaying statistics.

---------
### Progress

### Issue
 * [JDK-8298597](https://bugs.openjdk.org/browse/JDK-8298597): HdrSeq: support for a merge() method

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/184/head:pull/184` \
`$ git checkout pull/184`

`$ git checkout pull/184` \
`$ git pull https://git.openjdk.org/shenandoah pull/184/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 184`

`$ git pr show -t 184`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/184.diff">https://git.openjdk.org/shenandoah/pull/184.diff</a>

</details>


src/hotspot/share/gc/shenandoah/shenandoahNumberSeq.cpp
```diff
@@ -56,7 +56,7 @@ void HdrSeq::add(double val) {
   int mag;
   if (v > 0) {
     mag = 0;
-    while (v > 1) {
```
ysramakrishna, line 59: This is a bug fix that has been independently pushed to tip. You can ignore it and it'll find its way into shenandoah in due course.

src/hotspot/share/gc/shenandoah/shenandoahNumberSeq.hpp
```diff
@@ -50,6 +50,9 @@ class HdrSeq: public NumberSeq {
 
   virtual void add(double val);
   double percentile(double level) const;
+
+  // Merge this HdrSeq into hdr2, optionally clearing this HdrSeq
+  void merge(HdrSeq& hdr2, bool clear_this = true);
```
ysramakrishna, line 55: The default setting here is based on the way its only current client (RS scan instrumentation) makes use of it, but I am happy to change it if reviewers feel that might be better for API hygiene reasons.

src/hotspot/share/utilities/numberSeq.cpp
```diff
@@ -110,6 +110,29 @@ double AbsSeq::dsd() const {
...
+
+  if (num() == 0) return;  // nothing to do
+
+  abs2._num += _num;
+  abs2._sum += _sum;
+  abs2._sum_of_squares += _sum_of_squares;
+
+  // Decaying stats need a bit more thought
+  assert(abs2._alpha == _alpha, "Caution: merge incompatible?");
+  // guarantee(false, "NYI");
```
ysramakrishna, line 124: This will expand into setting some breadcrumbs in `abs2`, such that any attempt to query a decayed stat from the object will result in an error.

ysramakrishna (commented):
Will leave these comments here in the draft PR, until the last two steps are completed and the PR opened for formal review.

ysramakrishna:
This PR is open for review.

test/hotspot/gtest/gc/shenandoah/test_shenandoahNumberSeq.cpp
```diff
@@ -0,0 +1,133 @@
+/*
```
ysramakrishna, line 1: An earlier version of this test is in tip for an earlier bug fix. I am happy to consult if there is any confusion during a merge from tip. In this specific case, the contents of this file should take precedence.

kdnilsen (approved):

kdnilsen (approved):

test/hotspot/gtest/gc/shenandoah/test_shenandoahNumberSeq.cpp
ysramakrishna: merged from master; resolving.

ysramakrishna:
/integrate

(pending sponsorship; thanks for the review @kdnilsen)

PR #185, opened 2022-12-16
Initial sizing refactor

earthling-amzn:
Some things to highlight here:
* This change borrows a bit of code from G1 to handle processing of command line arguments used to size the young generation.
* A (hard coded for now) threshold on the difference between young/old time has been added to reduce resizing churn.
* The adaptive heuristic doesn't consider the `soft_tail` anymore. `available` is already adjusted for the soft max capacity.
* `SoftMaxHeapSize` is used to compute the soft max size and max size for the young generation.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/185/head:pull/185` \
`$ git checkout pull/185`

`$ git checkout pull/185` \
`$ git pull https://git.openjdk.org/shenandoah pull/185/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 185`

`$ git pr show -t 185`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/185.diff">https://git.openjdk.org/shenandoah/pull/185.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -447,7 +447,33 @@ jint ShenandoahHeap::initialize() {
   return JNI_OK;
 }
 
+size_t ShenandoahHeap::max_size_for(ShenandoahGeneration* generation) const {
+  switch (generation->generation_mode()) {
+    case YOUNG:  return _generation_sizer.max_young_size();
+    case OLD:    return max_capacity() - _generation_sizer.min_young_size();
+    case GLOBAL: return max_capacity();
+    default:
+      ShouldNotReachHere();
+      return 0;
+  }
+}
+
+size_t ShenandoahHeap::min_size_for(ShenandoahGeneration* generation) const {
+  switch (generation->generation_mode()) {
+    case YOUNG:  return _generation_sizer.min_young_size();
+    case OLD:    return max_capacity() - _generation_sizer.max_young_size();
+    case GLOBAL: return min_capacity();
+    default:
+      ShouldNotReachHere();
+      return 0;
+  }
+}
+
 void ShenandoahHeap::initialize_generations() {
+  // Max capacity is the maximum _allowed_ capacity. That is, the maximum allowed capacity
+  // for old would be total heap - minimum capacity of young. This means the sum of the maximum
+  // allowed for old and young could exceed the total heap size. It remains the case that the
+  // _actual_ capacity of young + old = total.
   _generation_sizer.heap_size_changed(soft_max_capacity());
   size_t initial_capacity_young = _generation_sizer.max_young_size();
   size_t max_capacity_young = _generation_sizer.max_young_size();
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -449,14 +449,14 @@ jint ShenandoahHeap::initialize() {
 
 void ShenandoahHeap::initialize_generations() {
   _generation_sizer.heap_size_changed(soft_max_capacity());
-  size_t initial_capacity_young = _generation_sizer.min_young_size();
+  size_t initial_capacity_young = _generation_sizer.max_young_size();
   size_t max_capacity_young = _generation_sizer.max_young_size();
   size_t initial_capacity_old = max_capacity() - max_capacity_young;
   size_t max_capacity_old = max_capacity() - initial_capacity_young;
 
   _young_generation = new ShenandoahYoungGeneration(_max_workers, max_capacity_young, initial_capacity_young);
   _old_generation = new ShenandoahOldGeneration(_max_workers, max_capacity_old, initial_capacity_old);
-  _global_generation = new ShenandoahGlobalGeneration(_max_workers, max_capacity(), max_capacity());
+  _global_generation = new ShenandoahGlobalGeneration(_max_workers, soft_max_capacity(), soft_max_capacity());
 }
 
 void ShenandoahHeap::initialize_heuristics() {
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -449,13 +449,14 @@ jint ShenandoahHeap::initialize() {
 
 void ShenandoahHeap::initialize_generations() {
   _generation_sizer.heap_size_changed(soft_max_capacity());
+  size_t initial_capacity_young = _generation_sizer.min_young_size();
   size_t max_capacity_young = _generation_sizer.max_young_size();
-  size_t max_capacity_old = max_capacity() - max_capacity_young;
-  size_t max_capacity_global = max_capacity_young + max_capacity_old;
+  size_t initial_capacity_old = max_capacity() - max_capacity_young;
+  size_t max_capacity_old = max_capacity() - initial_capacity_young;
 
-  _young_generation = new ShenandoahYoungGeneration(_max_workers, max_capacity_young, max_capacity_young);
-  _old_generation = new ShenandoahOldGeneration(_max_workers, max_capacity_old, max_capacity_old);
-  _global_generation = new ShenandoahGlobalGeneration(_max_workers, max_capacity_global, max_capacity_global);
+  _young_generation = new ShenandoahYoungGeneration(_max_workers, max_capacity_young, initial_capacity_young);
+  _old_generation = new ShenandoahOldGeneration(_max_workers, max_capacity_old, initial_capacity_old);
+  _global_generation = new ShenandoahGlobalGeneration(_max_workers, max_capacity(), max_capacity());
 }
 
 void ShenandoahHeap::initialize_heuristics() {
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -448,15 +448,14 @@ jint ShenandoahHeap::initialize() {
 }
 
 void ShenandoahHeap::initialize_generations() {
-  size_t max_capacity_new      = young_generation_capacity(max_capacity());
-  size_t soft_max_capacity_new = young_generation_capacity(soft_max_capacity());
-  size_t max_capacity_old      = max_capacity() - max_capacity_new;
-  size_t soft_max_capacity_old = soft_max_capacity() - soft_max_capacity_new;
+  _generation_sizer.heap_size_changed(soft_max_capacity());
+  size_t max_capacity_young = _generation_sizer.max_young_size();
+  size_t max_capacity_old = max_capacity() - max_capacity_young;
+  size_t max_capacity_global = max_capacity_young + max_capacity_old;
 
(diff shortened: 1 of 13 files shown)


earthling-amzn:
Converted this to a draft because the changes are tripping an assert:
```
#  Internal Error (/codebuild/output/src797/src/s3/00/src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp:986), pid=1289, tid=1312
--
520 | #  assert(regions * ShenandoahHeapRegion::region_size_bytes() <= heap->young_generation()->adjusted_capacity()) failed: Number of young regions cannot exceed adjusted capacity
```

earthling-amzn:
Ready for review - the assertion noted earlier is unrelated to these changes and will be fixed under a separate PR.

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
```diff
@@ -83,56 +82,157 @@ ShenandoahMmuTracker::~ShenandoahMmuTracker() {
 }
 
 void ShenandoahMmuTracker::record(ShenandoahGeneration* generation) {
-  // This is only called by the control thread.
+  // This is only called by the control thread or the VM thread.
```
ysramakrishna, line 85: Would it be worthwhile asserting the calling thread's identity here, just to catch the problem early if someone were to try to do it from a different thread in the future? (although i can't imagine why anyone would.)

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -448,15 +448,15 @@ jint ShenandoahHeap::initialize() {
...
+  size_t initial_capacity_old = max_capacity() - max_capacity_young;
+  size_t max_capacity_old = max_capacity() - initial_capacity_young;
 
-  _young_generation = new ShenandoahYoungGeneration(_max_workers, max_capacity_new, soft_max_capacity_new);
-  _old_generation = new ShenandoahOldGeneration(_max_workers, max_capacity_old, soft_max_capacity_old);
-  _global_generation = new ShenandoahGlobalGeneration(_max_workers, max_capacity_new + max_capacity_old,
-                                                      soft_max_capacity_new + soft_max_capacity_old);
+  _young_generation = new ShenandoahYoungGeneration(_max_workers, max_capacity_young, initial_capacity_young);
+  _old_generation = new ShenandoahOldGeneration(_max_workers, max_capacity_old, initial_capacity_old);
+  _global_generation = new ShenandoahGlobalGeneration(_max_workers, soft_max_capacity(), soft_max_capacity());
```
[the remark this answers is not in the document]
earthling-amzn: I've been thinking of the max capacity as the maximum _allowed_ capacity. For example, the maximum _allowed_ capacity for old would be `total heap - minimum capacity of young`. So, the sum of the maximum allowed for old and young could exceed the total. If that makes sense, I will put the explanation in a comment here.

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
```diff
@@ -83,56 +83,157 @@ ShenandoahMmuTracker::~ShenandoahMmuTracker() {
...
-  assert(bytes < heap_capacity, "Must be less than total capacity");
-  return size_t(100.0 * double(bytes) / double(heap_capacity));
+  if (delta > 0) {
+    // young is busier than old, increase size of young to raise MMU
+    return transfer_capacity(old, young);
+  } else {
+    // old is busier than young, increase size of old to raise MMU
+    return transfer_capacity(young, old);
+  }
 }
```
kdnilsen, line 235: I still question whether this is the right trigger for when to enlarge old-gen.  A properly running generational GC will spend nearly all its time doing young-gen and very little time doing old-gen.

The trigger for enlarging old-gen should be that we experience promotion failures (and/or that we identify at the end of init mark that we have more live data in aged regions than will fit in the current old-gen).

Old-gen collection triggers need to be refined when we are auto-sizing.  We can't use "percent free in old" or even "time to collect old > time to exhaust old", because we are trying to auto-tune to maintain that the percent free in old is very small.  We need a new way to trigger old-gen GCs.

Maybe, we trigger old-gen GC (rather than enlarging old-gen) if an old-gen enlargement request would cause us to exceed a target max-size for old-gen?

This doesn't have to be addressed in current PR.  But I'm just thinking that refinements of these heuristics may be necessary eventually.

kdnilsen: Maybe the target max-size for old-gen is auto-tuned to, e.g., 10% larger than the maximum old-gen live memory following old-gen concurrent mark with some bias given to more recently observed measurements of old-gen live-memory.

kdnilsen (approved):

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
earthling-amzn: I agree we can wire up more signals to the resizing mechanism. In the scenario you describe, where old generation has become _too small_ and old collections are running _too frequently_, the MMU based resizing would enlarge the old generation.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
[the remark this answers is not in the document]
ysramakrishna: Makes sense; thanks for the updated comments!

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
```diff
@@ -182,9 +265,34 @@ bool ShenandoahMmuTracker::transfer_capacity(ShenandoahGeneration* from, Shenand
...
 
-void ShenandoahMmuTracker::initialize() {
-  _initial_process_time_s = process_time_seconds();
-  _initial_collector_time_s = gc_thread_time_seconds();
-  _initial_verify_collector_time_s = _initial_collector_time_s;
-  _mmu_periodic_task->enroll();
-}
\ No newline at end of file
+size_t round_down_to_multiple_of_region_size(size_t bytes) {
```
ysramakrishna, line 268: I could have sworn there was a rounding utility/macro extensively used in sizing code, but the only one I found was a power of 2 rounder. The alternative, if one maintained a log of heap region size (being a power of 2) would be to use a bit-mask here.

Anyway, nothing to do here; this looks good for now.

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
```diff
@@ -83,56 +83,157 @@ ShenandoahMmuTracker::~ShenandoahMmuTracker() {
...
+  const double transfer_threshold = 3.0;
+  double delta = young_time_s - old_time_s;
+
   log_info(gc)("Thread Usr+Sys YOUNG = %.3f, OLD = %.3f, GLOBAL = %.3f", young_time_s, old_time_s, global_time_s);
 
-  if (old_time_s > young_time_s) {
-    return transfer_capacity(young, old);
-  } else {
-    return transfer_capacity(old, young);
+  if (abs(delta) <= transfer_threshold) {
```
ysramakrishna, line 222: I thought the original idea was to use the difference in MMU's for old and young as the error signal to drive the (direction of the) transfer, rather than the difference in the actual times? Am I misinterpreting what `reset_collection_time` returns? You do refer to it as `thread utilization` (akin to MMU) in the log message below.

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
ysramakrishna: What do you do before any data is available for one of the MMU trackers? (for example before the first old collection cycle has happened.) I'd assume that the control algorithm wouldn't kick in until the data driving the control signal was valid and available. Where is that done?

src/hotspot/share/gc/shenandoah/shenandoahYoungGeneration.hpp
```diff
@@ -52,6 +52,12 @@ class ShenandoahYoungGeneration : public ShenandoahGeneration {
     return _old_gen_task_queues;
   }
 
+  // Returns true if the young generation is configured to enqueue old
+  // oops for the old generation mark queues.
+  bool is_bootstrap_cycle() {
```
ysramakrishna, line 57: Why is this called a `bootstrap cycle`? I must be missing some big picture background of nomenclature here.

ysramakrishna (approved):
General direction looks good to me, as well as the refactoring which greatly improved the structure of the code.

I just had one question about the error signal that drives the control actuation for resizing. I'd expected that to be the _difference of mmu averages_ measured whenever new data is available for these control signals, rather than _difference of collection times_, which it looked like to me (although it's possible I misunderstood).

Also, it'd be great to compare some basic performance numbers with the control algorithm off vs on to show its efficacy.

Thanks!

src/hotspot/share/gc/shenandoah/shenandoahYoungGeneration.hpp
earthling-amzn: Every old generation cycle is preceded by a young collection. We call this a bootstrap cycle because it populates the old generation mark queues with old objects it encountered during the marking of young. Otherwise, we'd have to maintain a reverse-remembered set for young->old pointers.

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
earthling-amzn: I reasoned that the denominator in a comparison between MMU's (i.e., process time) would be the same on both sides of the comparison, so I omitted it.

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
earthling-amzn: I wasn't sure if region size was coerced to a power of a 2, but it seems to be the case:
```
size_t ShenandoahHeapRegion::setup_sizes(size_t max_heap_size) {
// ...
  int region_size_log = log2i(region_size);
  // Recalculate the region size to make sure it's a power of
  // 2. This means that region_size is the largest power of 2 that's
  // <= what we've calculated so far.
  region_size = size_t(1) << region_size_log;
```
I'll make that change.

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
ysramakrishna: My worry is that using the latest sample might lead to too much variance. Using a TruncatedSeq's decaying average would help with damping the error signal and lead to smoother control over sizing changes.

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
earthling-amzn: We generally have a flurry of "learning cycles" at startup, but it makes sense to add an explicit check to avoid premature resizing. To Kelvin's earlier point, it may also make sense to re-initiate the learning phase after resizing the generations.

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
earthling-amzn: I'll experiment with that. I was going for responsiveness. Smoothing out the signal may also introduce a delay in resizing.

PR #188, opened 2022-12-21
Avoid divide by zero error, improve variable names

earthling-amzn:
Depending on when the periodic thread runs the accounting task and how much CPU time the process receives, we may see a very small elapsed process time. Such a time should not be used to compute MMU.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/188/head:pull/188` \
`$ git checkout pull/188`

`$ git checkout pull/188` \
`$ git pull https://git.openjdk.org/shenandoah pull/188/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 188`

`$ git pr show -t 188`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/188.diff">https://git.openjdk.org/shenandoah/pull/188.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
@@ -28,9 +28,9 @@
 #include "gc/shenandoah/shenandoahHeap.inline.hpp"
 #include "gc/shenandoah/shenandoahOldGeneration.hpp"
 #include "gc/shenandoah/shenandoahYoungGeneration.hpp"
+#include "logging/log.hpp"
 #include "runtime/os.hpp"
 #include "runtime/task.hpp"
-#include "logging/log.hpp"
 
 
 class ShenandoahMmuTask : public PeriodicTask {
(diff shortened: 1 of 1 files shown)


kdnilsen (approved):

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
```diff
@@ -94,12 +94,17 @@ void ShenandoahMmuTracker::report() {
   // This is only called by the periodic thread.
   double process_time_s = process_time_seconds();
   double elapsed_process_time_s = process_time_s - _process_reference_time_s;
+  if (elapsed_process_time_s <= 0.01) {
```
ysramakrishna, line 97: Is there a different mechanism one could use to check if there was an update? (like a count of GC cycles started or ended, rather than relying on the magnitude of the difference, although the magnitude of the difference may still be sufficient. One related question: is this unit "seconds" or some smaller unit? The `_s` implies it's a second and 0.01 s is 10 ms which is not insubstantial...)

ysramakrishna (approved):

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
ysramakrishna: In general the synchronous, time-based, sampling of the MMU suffers from this issue of high variance by catching the co-initial or co-terminal portion of a cycle. Have you looked at the variance of MMUs to see if you should sample only at the end of complete cycles `synchronous with cycles` rather than using a `time-based sampling trigger`? Anyway something to think about and look into.

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
```diff
@@ -94,12 +94,17 @@ void ShenandoahMmuTracker::report() {
...
   _process_reference_time_s = process_time_s;
-  double verify_time_s = gc_thread_time_seconds();
-  double verify_elapsed = verify_time_s - _collector_reference_time_s;
-  _collector_reference_time_s = verify_time_s;
-  double verify_mmu = ((elapsed_process_time_s - verify_elapsed) / elapsed_process_time_s) * 100;
-  _mmu_average.add(verify_mmu);
+  double collector_time_s = gc_thread_time_seconds();
+  double elapsed_collector_time_s = collector_time_s - _collector_reference_time_s;
+  _collector_reference_time_s = collector_time_s;
+  double minimum_mutator_utilization = ((elapsed_process_time_s - elapsed_collector_time_s) / elapsed_process_time_s) * 100;
```
ysramakrishna, line 106: Alternative: You could check if result `isnan()` and not use it, although I don't know if divide by zero is costly.

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
earthling-amzn: I didn't want to use GC count changing because then it would miss 100% MMU (which is something to celebrate). This should be called once every GCPauseIntervalMillis (5000 by default). It's hard to imagine the process didn't have any CPU time over such an interval, but not impossible. Seems more likely the scheduled task thread fell behind and executed the task twice in quick succession. I'm not totally sure how this happens.

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
earthling-amzn: I'd just as soon catch it earlier and I'm not 100% certain that other values won't also cause this error. Also, `isnan` won't flag the result of division by zero (`isinf` would though).

PR #192, opened 2023-01-05
Fix allocate aligned

kdnilsen:
An error was discovered in the implementation and use of allocate_aligned(), which is used to allocate PLABs that align with remembered set card boundaries.  In the previous implementation, if the required alignment padding was smaller than the minimum filler object, the additional card's memory worth of padding might cause the PLAB to span beyond the end of the selected heap region.  This PR addresses the error.  This code successfully runs our internal pipeline of tests without any regressions.

In the current context, this code is not fully exercised due to very limited allocation of PLABs within heap regions that already hold previously allocated PLABs.  This code has also been exercised in the context of code that more aggressively packs multiple PLABs into heap regions.  That additional code will be integrated with a future PR.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/192/head:pull/192` \
`$ git checkout pull/192`

`$ git checkout pull/192` \
`$ git pull https://git.openjdk.org/shenandoah pull/192/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 192`

`$ git pr show -t 192`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/192.diff">https://git.openjdk.org/shenandoah/pull/192.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
@@ -267,7 +267,7 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
         assert(r->top() <= r->end(), "Allocation cannot span end of region");
         // actual_size() will be set to size below.
         assert((result == nullptr) || (size % CardTable::card_size_in_words() == 0),
-               "PLAB size must multiple of card size");
+               "PLAB size must be multiple of card size");
         assert((result == nullptr) || (((uintptr_t) result) % CardTable::card_size_in_words() == 0),
                "PLAB start must align with card boundary");
 
--- a/src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
@@ -197,9 +197,7 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
       r->is_trash()) {
     return NULL;
   }
-
   try_recycle_trashed(r);
-
   if (r->affiliation() == ShenandoahRegionAffiliation::FREE) {
     ShenandoahMarkingContext* const ctx = _heap->complete_marking_context();
     r->set_affiliation(req.affiliation());
@@ -227,7 +225,6 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
   }
 
   in_new_region = r->is_empty();
-
   HeapWord* result = NULL;
   size_t size = req.size();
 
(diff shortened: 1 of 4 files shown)


src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.inline.hpp
```diff
@@ -44,18 +44,26 @@ HeapWord* ShenandoahHeapRegion::allocate_aligned(size_t size, ShenandoahAllocReq
...
+    if (pad_words > 0) {
       ShenandoahHeap::fill_with_object(obj, pad_words);
+      // register the filled pad object
       ShenandoahHeap::heap()->card_scan()->register_object(obj);
       obj += pad_words;
     }
 
+    // We don't need to register the PLAB.  Its content will be registered as objects are allocated within it and/or
+    // when the PLAB is retired.
+    ShenandoahHeap::heap()->card_scan()->register_object(obj);
```
kdnilsen, line 66: In reviewing my own code, it looks like my implementation contradicts the comment.  I'm going to retest without line 66.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -279,13 +303,20 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
...
     if (size <= usable_free) {
-      assert(size % CardTable::card_size_in_words() == 0, "PLAB size must be multiple of remembered set card size");
-
       result = r->allocate_aligned(size, req, CardTable::card_size());
-      if (result != nullptr) {
+      assert(result != nullptr, "Allocation cannot fail");
+      assert(r->top() <= r->end(), "Allocation cannot span end of region");
+      assert(req.actual_size() % CardTable::card_size_in_words() == 0, "PLAB start must align with card boundary");
+      assert(((uintptr_t) result) % CardTable::card_size_in_words() == 0, "PLAB start must align with card boundary");
+      if (result != nullptr && free > usable_free) {
```
earthling-amzn, line 319: Line 315 asserts that `result` cannot be `nullptr`, do we need to check for non-null again here?

earthling-amzn (changes requested):

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.inline.hpp
kdnilsen: Making this change and testing on regression suite pipeline before integration.

ysramakrishna (approved):
Changes look fine.

I'd like to understand the original rationale for making PLAB boundaries exactly card-aligned. Perhaps it's described/documented somewhere in the code? (Something to do with simplifying card-scanning concurrently with allocating out of PLABs?)

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -279,13 +303,20 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
...
     }
+    assert(size % CardTable::card_size_in_words() == 0, "PLAB size must be multiple of remembered set card size");
     if (size <= usable_free) {
-      assert(size % CardTable::card_size_in_words() == 0, "PLAB size must be multiple of remembered set card size");
-
       result = r->allocate_aligned(size, req, CardTable::card_size());
-      if (result != nullptr) {
+      assert(result != nullptr, "Allocation cannot fail");
+      assert(r->top() <= r->end(), "Allocation cannot span end of region");
+      assert(req.actual_size() % CardTable::card_size_in_words() == 0, "PLAB start must align with card boundary");
```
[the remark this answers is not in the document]
kdnilsen: This allows us to register objects in PLABs without acquiring a lock.  Otherwise, we need a lock because two threads might be registering objects within the same card in parallel.

kdnilsen:
After confirming that the two fixes motivated by review do not introduce regressions on our CI pipelines, I will close this with integration.

/integrate

earthling-amzn (approved):

PR #195, opened 2023-01-06
Plab fallback to minsize

kdnilsen:
If there is not sufficient memory within a region to allocate the desired size of a PLAB, try to allocate a smaller PLAB that is still larger than the minimum PLAB size.  This allows more efficient packing of PLABs into available old-gen heap regions.

This patch has been exercised by an internal CI regression testing pipeline and has not introduced any regressions.

Modest reductions in evacuation efforts are seen on benchmarks that do heavy promotion.  This is presumably because there are fewer promotion failures, thus less need to repeatedly copy "old" objects within young-gen spaces.  Of note, but not entirely explained, is the observation that some evacuation effort has shifted from mutator threads to GC worker threads.

Some regression in jhiccup pause times were observed.  This will be explored further after additional patches are integrated.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/195/head:pull/195` \
`$ git checkout pull/195`

`$ git checkout pull/195` \
`$ git pull https://git.openjdk.org/shenandoah pull/195/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 195`

`$ git pr show -t 195`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/195.diff">https://git.openjdk.org/shenandoah/pull/195.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
@@ -261,8 +261,17 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
         adjusted_min_size = adjusted_min_size - remnant + CardTable::card_size_in_words();
       }
       if (size >= adjusted_min_size) {
+#undef KELVIN_DEBUG
+#ifdef KELVIN_DEBUG
+        log_info(gc, ergo)("tai() size (" SIZE_FORMAT ") > adjusted_min_size (" SIZE_FORMAT ")",
+                           size, adjusted_min_size);
+#endif
         result = r->allocate_aligned(size, req, CardTable::card_size());
         assert(result != nullptr, "Allocation cannot fail");
+#ifdef KELVIN_DEBUG
+        log_info(gc, ergo)("aa() returned " PTR_FORMAT, p2i(result));
+#endif
+        size = req.actual_size();
         assert(r->top() <= r->end(), "Allocation cannot span end of region");
         // actual_size() will be set to size below.
         assert((result == nullptr) || (size % CardTable::card_size_in_words() == 0),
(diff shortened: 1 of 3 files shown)


earthling-amzn:
The evacuation metrics between mutators and gc workers are fairly unstable - likely because it depends so much on when and which threads get scheduled. I've been thinking of masking them in the reports for this reason.

earthling-amzn (approved):

ysramakrishna:
Generally looks good, but I wonder if one should more carefully define the pre- and post-conditions of the two allocate methods to avoid duplicated computation between them (especially wrt minimum size etc.) One way to achieve that would be to have more specialized allocate methods that are called by subsets of clients. Having a leaf method called by several could lead to such duplication of checks. e.g. I see a bunch of "result != null" for values returned from a method that does checks and trimming of its own. If so, the checks in the leaf method for that caller may be wasteful.

This is a general comment, but I'll look more carefully at the code to understand this better.

One question: do PLAB requests that give you smaller PLABs slow down all subsequent PLAB requests in that region? Does this then result in a donwsizing of PLAB requests in the same cycle or subsequent ones? (I guess I am asking how often / what cycle PLAB/GCLAB resizing happens.)

ysramakrishna (approved):
Changes look good, modulo general comments for longer-term consideration.

kdnilsen:
The code for sizing PLAB requests is in allocate_from_plab_slow().  In general, each thread starts out with PLABs of size PLAB::min_size().  Each time the thread exhausts its existing PLAB, it tries to allocate a new PLAB that's twice as large as its previously preferred PLAB size, even if its previous PLAB is smaller than its previously preferred PLAB size.  The consequence of downsizing a particular PLAB is that the thread will end up depleting the downsized PLAB more quickly than normal, which will result in this thread subsequently receiving an even larger PLAB sooner.

PR #202, opened 2023-01-16
8298647: GenShen require heap size 2MB granularity

stooart-mon:
Generational Shenandoah requires 2MB granularity in order for card tables to cover the allocated heap. Each byte in a page of card table represents 512 heap bytes. As card tables are allocated 4KB at a time, 4KB * 512 = 2MB.

There is a circular dependency between the region calculations and the heap size calculations. This unconditionally rounds up the heap size to 2MB. It might be preferable to do this only when generational mode is enabled.

Running with:
  java -Xlog:gc*=trace -XX:+UseShenandoahGC -mx495m \
       -XX:ShenandoahGCMode=generational -version

on a debug build is sufficient to reproduce this problem.

---------
### Progress

### Issue
 * [JDK-8298647](https://bugs.openjdk.org/browse/JDK-8298647): GenShen require heap size 2MB granularity

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/202/head:pull/202` \
`$ git checkout pull/202`

`$ git checkout pull/202` \
`$ git pull https://git.openjdk.org/shenandoah pull/202/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 202`

`$ git pr show -t 202`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/202.diff">https://git.openjdk.org/shenandoah/pull/202.diff</a>

</details>


src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -724,6 +725,9 @@ size_t ShenandoahHeapRegion::setup_sizes(size_t max_heap_size) {
     FLAG_SET_DEFAULT(ShenandoahMinRegionSize, MIN_REGION_SIZE);
   }
 
+  // Generational Shenandoah needs this alignment for card tables.
```
earthling-amzn, line 728: Thank you for this fix! It would be nice if this constraint were only applied for generation mode, but these sizes are computed quite earlier during startup. You'd need to factor the code out of `ShenandoahHeap::initialize_heuristics` to know whether the constraint is required at this point.

earthling-amzn (changes requested):

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
stooart-mon: Yes, I thought I'd start with a simple fix to highlight the problem first. I experimented with exactly what you suggested, parsing ShenandoahGCMode. I can update this the 2MB alignment isn't desired unconditionally.

earthling-amzn: Yes please. We've tried to limit the impact of the generational mode on Shenandoah's other modes.

stooart-mon:
I included with the patch the statistics for the new patterns, here they are:

Adds new patterns to match a memory address and a constant, which are
emitted as two loads. C2 expects that constants INTs will always be
manifested by immediate loads. However, as these intrinsics are avoiding
loading into GPRs and moving them into the vector registers, there is an
addition to allow integers to be loaded as a constant.

There isn't a pattern to match two LoadI nodes - C2 doesn't handle nodes
with two memory nodes for calculating anti-dependencies.
```

Benchmark                  Result  Units  % against non-SVE2
Integers.compress          2.009   µs/op
Integers.compress-SVE      1.435   µs/op  71.43%
Integers.compress-SVE+mem  1.263   µs/op  62.87%
Integers.expand            2.129   µs/op
Integers.expand-SVE        1.433   µs/op  67.31%
Integers.expand-SVE+mem    1.32    µs/op  62.00%
Longs.compress             2.504   µs/op
Longs.compress-SVE         1.445   µs/op  57.71%
Longs.compress-SVE+mem     1.269   µs/op  50.68%
Longs.expand               2.614   µs/op
Longs.expand-SVE           1.489   µs/op  56.96%
Longs.expand-SVE+mem       1.272   µs/op  48.66%
```

earthling-amzn (approved):
Thank you for this contribution!

PR #203, opened 2023-01-17
Do not reset learning cycles after resizing

earthling-amzn:
Also, require more than 10 gc cycles before trigger can resize generations. The behavior on the extremem 'phased' workload looks much better:
```
[1334.381s][info][gc,stats       ]    66 Successful Concurrent GCs
[1334.381s][info][gc,stats       ]       0 invoked explicitly
[1334.381s][info][gc,stats       ]       0 invoked implicitly
[1334.381s][info][gc,stats       ] 
[1334.381s][info][gc,stats       ]     7 Completed Old GCs
[1334.381s][info][gc,stats       ]       0 mixed
[1334.381s][info][gc,stats       ]       0 interruptions
[1334.381s][info][gc,stats       ] 
[1334.381s][info][gc,stats       ]     1 Degenerated GCs
[1334.381s][info][gc,stats       ]       1 caused by allocation failure
[1334.381s][info][gc,stats       ]         1 happened at Mark
[1334.381s][info][gc,stats       ]       1 upgraded to Full GC
[1334.381s][info][gc,stats       ] 
[1334.381s][info][gc,stats       ]     0 Abbreviated GCs
[1334.381s][info][gc,stats       ] 
[1334.381s][info][gc,stats       ]     1 Full GCs
[1334.381s][info][gc,stats       ]       0 invoked explicitly
[1334.381s][info][gc,stats       ]       0 invoked implicitly
[1334.381s][info][gc,stats       ]       0 caused by allocation failure
[1334.381s][info][gc,stats       ]       1 upgraded from Degenerated GC
```
The full cycle here was the first cycle after the last of the initial learning cycles.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/203/head:pull/203` \
`$ git checkout pull/203`

`$ git checkout pull/203` \
`$ git pull https://git.openjdk.org/shenandoah pull/203/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 203`

`$ git pr show -t 203`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/203.diff">https://git.openjdk.org/shenandoah/pull/203.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
@@ -495,6 +495,7 @@ bool ShenandoahAdaptiveHeuristics::resize_and_evaluate() {
     return true;
   }
 
+  log_info(gc)("Increased size of %s generation, re-evaluate trigger criteria", _generation->name());
   return should_start_gc();
 }
 
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
@@ -55,6 +55,8 @@ const double ShenandoahAdaptiveHeuristics::HIGHEST_EXPECTED_AVAILABLE_AT_END = 0
 const double ShenandoahAdaptiveHeuristics::MINIMUM_CONFIDENCE = 0.319; // 25%
 const double ShenandoahAdaptiveHeuristics::MAXIMUM_CONFIDENCE = 3.291; // 99.9%
 
+const uint ShenandoahAdaptiveHeuristics::MINIMUM_RESIZE_INTERVAL = 10;
+
 ShenandoahAdaptiveHeuristics::ShenandoahAdaptiveHeuristics(ShenandoahGeneration* generation) :
   ShenandoahHeuristics(generation),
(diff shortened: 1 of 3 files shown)


ysramakrishna (approved):
LGTM, reviewed. But curious if you noticed any difference wrt, e.g., specjbb.

Some more thoughts:
I wonder if the # of cycles to wait would be proportional to the long-run ratio of minor to major collection cycles.

I agree though that waiting about 10 cycles between resizing decisions would have the salubrious effect of smoothing out any temporary spikes. Somewhat relatedly, and something that I only vaguely paid attention to before: What's the default decay factor for the MMU decaying average, and what constitutes an MMU sample: the occurrence of a GC (minor or major), or just a synchronous 5-second sample of both (which might decay very quickly to zero, losing almost all the information in the signal after 6-7 samples, i.e. 30-35 seconds in this case).

kdnilsen (approved):
Thanks.  This looks like a very good improvement.

(I assume we are separately looking into the case where the MMU triggers cause young to expand even when a more enlightened perspective would instead expand OLD.)

earthling-amzn:
@ysramakrishna , I didn't notice the trouble with these learning cycles originally because on specjbb, the heuristic quickly maxes out the young generation size and keeps it there. The decay factor is set by `ShenandoahAdaptiveDecayFactor` (default is 0.5) and the MMU is updated every `GCPauseIntervalMillis` (default is 5 seconds).

@kdnilsen , I will look for workloads that defeat the heuristic. It's much better behaved now on the extremem 'phased' workload. Perhaps heapothesys with a high occupancy rate?

PR #220, opened 2023-02-18
Performance improvements for non-generational modes

earthling-amzn:
There are a basket of changes here:
* Make heuristics a bit less aggressive:
   * Revert threshold triggers to use `available` rather than adjusted head room.
   * Remove code that adjusted average gc time, it was causing wild fluctuations and making heuristic behave erratically
* Skip over more generational mode constraints on the allocation path
* Disable `ShenadoahEvacLABRatio` by default as it causes unused bytes in survivors regions, which in turn leads to these regions to being included in the mutator view of free memory. This contributed to higher fragmentation rates.
* More detailed debugging messages under the `gc+free` tags.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah pull/220/head:pull/220` \
`$ git checkout pull/220`

`$ git checkout pull/220` \
`$ git pull https://git.openjdk.org/shenandoah pull/220/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 220`

`$ git pr show -t 220`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/220.diff">https://git.openjdk.org/shenandoah/pull/220.diff</a>

</details>


earthling-amzn:
Converted to draft form pending the results of performance testing. I also intend to factor out new heuristic classes for generational mode, and also untangle the allocation paths for generational mode.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
```diff
@@ -311,7 +311,7 @@
           "which upgrades to full GC whenever the budget is exceeded.")     \
           range(1.0,100.0)                                                  \
                                                                             \
-  product(uintx, ShenandoahMaxEvacLABRatio, 16, EXPERIMENTAL,               \
+  product(uintx, ShenandoahMaxEvacLABRatio, 0, EXPERIMENTAL,                \
```
kdnilsen, line 314: Surprised this would have an effect on single-generation performance.   Have we measured the costs/benefits of setting this default differently?  If 0 works best, maybe we should remove this entirely?

kdnilsen (approved):
I'll need to be careful not to overwrite these improvements with my next merge.  Some of the same code is touched.

Thanks.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
earthling-amzn: Yes, it was counter intuitive. This setting causes small chunks of memory to remain free in regions that have otherwise been consumed by gc lab allocations. When the evacuation reserve is exhausted, the collector will steal regions from the mutator. Because of the free memory in these regions, they are retained in the mutator's view of the freeset after the collection is complete. This in turn increases fragmentation in the mutator's view and reduces the amount of available memory to the mutators, causing the triggers to run more often.

I only tested the default setting (16) and turning it off completely. I will test additional values.

earthling-amzn: Additional tests with ShenandoahMaxEvacLABRatio=0, 8, 16, 32 shows the best performance with 0. All other values resulted in nearly full survivor regions being stuck in the mutator's free set.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
```diff
@@ -423,33 +430,11 @@ bool ShenandoahAdaptiveHeuristics::should_start_gc() {
 
   double avg_cycle_time = _gc_cycle_time_history->davg() + (_margin_of_error_sd * _gc_cycle_time_history->dsd());
```
ysramakrishna, line 432: Here again, we seem to have ended up with a discrepancy between the comments and the control flow that those comments applied to, but which has now disappeared? I'd remove the comments in case they are no longer applicable. If the idea is to revisit this in the future, might it be better to capture in a ticket rather than leaving in obsolete comments? In any case, it might make sense to go over these comments and make sure they reflect current reality and clearly call out future plans.

ysramakrishna (commented):
Sorry to not get these comments back in time.

I didn't understand some of the changes, although the majority seemed reasonable.

I just had a few comments where the logging, control flow, and comments seemed to have become inconsistent (or at least confusing to the naive reader) on account of the changes.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
```diff
@@ -361,7 +368,7 @@ bool ShenandoahAdaptiveHeuristics::should_start_gc() {
 
   size_t min_threshold = min_free_threshold();
 
-  if (allocation_headroom < min_threshold) {
+  if (available < min_threshold) {
     log_info(gc)("Trigger (%s): Free (" SIZE_FORMAT "%s) is below minimum threshold (" SIZE_FORMAT "%s)",
                  _generation->name(),
                  byte_size_in_proper_unit(allocation_headroom), proper_unit_for_byte_size(allocation_headroom),
```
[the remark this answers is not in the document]
earthling-amzn: Using `available` here is as the code is on `openjdk/jdk:master`.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
earthling-amzn: I'm not sure why you think the comment doesn't match the control flow. I left in the `TODO` intentionally. The preceding comment does describe how the heuristic operates - maintaining a model of an average allocation rate, but also looking for 'spikes' that are outside of what the model expects.

PR #228, opened 2023-03-20
Do not visit gc threads during shutdown

earthling-amzn:
Discovered a strange crash during shutdown while working on jtreg tests.

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -1801,6 +1801,10 @@ void ShenandoahHeap::prepare_for_verify() {
 }
 
 void ShenandoahHeap::gc_threads_do(ThreadClosure* tcl) const {
+  if (_shenandoah_policy->is_at_shutdown()) {
+    return;
+  }
+
   tcl->do_thread(_control_thread);
   tcl->do_thread(_regulator_thread);
   workers()->threads_do(tcl);


kdnilsen (approved):
Thanks for quick solution.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -1801,6 +1801,10 @@ void ShenandoahHeap::prepare_for_verify() {
 }
 
 void ShenandoahHeap::gc_threads_do(ThreadClosure* tcl) const {
+  if (_shenandoah_policy->is_at_shutdown()) {
+    return;
+  }
+
```
ysramakrishna, line 1807: Some related questions so I can understand the shutdown sequence a bit better, and what caused the issue that @kdnilsen recorded in the ticket:

1. I suppose we have determined that the shutdown sequence doesn't itself need to iterate over these GC-related threads to do anything.
2. was the original issue because one of the threads gone by the time we started iterating over it with the closure below?
3. what prevents a race between the check above and the exit of the threads subject to iteration below, so that some of the threads may be asynchronously gone between the check above and the iteration over that thread below?

Thanks!

ysramakrishna (commented):
Changes look reasonable, but would like a couple of clarifications to determine in my mind the robustness of the proposed fix. Thanks!

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
earthling-amzn: 1. The shutdown sequence doesn't need to iterate over the threads.
2. The crashing thread wasn't "gone", but it's termination had been requested. It crashed in a call to get the thread's clock time (pthread_getcpuclockid). The thread's native handle should still have been valid. I didn't fully root cause the crash.
3. In this case, it was the control thread itself that was trying to iterate the threads. The fix here will stop the control thread from starting an iteration of the threads _after_ the shutdown sequence has begun, but it is not so robust to cover a shut down during an iteration of the threads.

ysramakrishna: > 1. The shutdown sequence doesn't need to iterate over the threads.

Good!

> 2. The crashing thread wasn't "gone", but it's termination had been requested. It crashed in a call to get the thread's clock time (pthread_getcpuclockid). The thread's native handle should still have been valid. I didn't fully root cause the crash.

I assume you meant "the _GC thread that the control thread was iterating over_ wasn't gone, but its termination had been requested"?  I'd consider that as good as gone, absent any interlocking between the iteration and the termination of the target thread.

> 3. In this case, it was the control thread itself that was trying to iterate the threads. The fix here will stop the control thread from starting an iteration of the threads _after_ the shutdown sequence has begun, but it is not so robust to cover a shut down during an iteration of the threads.

May be one then needs a handshake/interlock with the control thread during the shutdown sequence if it's the only thread that iterates over these threads. (Or their termination could be delegated to the control thread, which is morally equivalent, and perhaps cleaner.) Is the case then that you have by your change above substantially narrowed the window of vulnerability, but not completely closed it? Or did I misunderstand?

earthling-amzn: The `core` file I had was on `/tmp` and was lost in one of the power outages last week. I meant that the control thread's termination was requested and it was the control thread itself that was iterating the other threads. I don't recall which visited thread caused the crash. As I think on this, I'd like to make another small change to make it more robust.

Yes, that's fair. This change narrows the window, but doesn't completely close it.

PR #230, opened 2023-03-20
Restrict region dump to young

kdnilsen:
We are primarily interested in the young-gen aged regions.  It is confusing for the log report to add the live data pertaining to old-gen regions which happen to have age > 0.


earthling-amzn:
I'm okay with this, but I want @ysramakrishna to weigh in on it. He may want to see the distribution of ages in old regions for the automatic tenuring threshold work.

kdnilsen:
Sure.  If we want report census of old-gen regions, we should probably report that in a separate table.

ysramakrishna (approved):

earthling-amzn (approved):

PR #231, opened 2023-03-21
Put regulator thread stop after gc thread iteration is prevented

earthling-amzn:
Record that shutdown sequence has begun before we request the regulator thread to stop.

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -2357,11 +2357,11 @@ uint ShenandoahHeap::max_workers() {
 void ShenandoahHeap::stop() {
   // The shutdown sequence should be able to terminate when GC is running.
 
+  // Step 0. Notify policy to disable event recording and prevent visiting gc threads during shutdown
+  _shenandoah_policy->record_shutdown();
+
   // Step 0a. Stop requesting collections.
   regulator_thread()->stop();
-
-  // Step 0. Notify policy to disable event recording.
-  _shenandoah_policy->record_shutdown();
 
   // Step 1. Notify control thread that we are in shutdown.
   // Note that we cannot do that with stop(), because stop() is blocking and waits for the actual shutdown.


src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -2357,12 +2357,12 @@ uint ShenandoahHeap::max_workers() {
 void ShenandoahHeap::stop() {
   // The shutdown sequence should be able to terminate when GC is running.
 
+  // Step 0. Notify policy to disable event recording and prevent visiting gc threads during shutdown
+  _shenandoah_policy->record_shutdown();
+
   // Step 0a. Stop requesting collections.
   regulator_thread()->stop();
```
ysramakrishna, line 2365: May be rename them Steps 0.a and 0.b in the order in which they are executed?

Could you state the advantage of doing steps in this rather than the old order? (I confess I looked at the code but couldn't find an immediate reason why this may be preferable to the previous order, e.g. does it accelerate the shutdown if the actions are in this new order, or something else such as prevention of a potential deadlock or a crash?)

earthling-amzn: The previous code would have allowed these events:
1. shutdown thread stops regulator thread
2. control thread begins iteration of gc threads (shutdown not 'recorded' yet)
3. shutdown thread records shutdown (meant to prevent iteration of stopped threads)
4. control thread visits stopped regulator thread (which could be unsafe)

This is change reduces the chances of the control thread iterating the gc threads after the regulator thread is stopped. Of course, without a robust locking sequence it may _still_ happen, but this code is closer to the intended sequence.

ysramakrishna: OK that makes sense; thanks!

ysramakrishna (approved):

shipilev (approved):

kdnilsen (approved):

PR #235, opened 2023-03-29
Recompute generation accounts after fullgc

kdnilsen:
This fix addresses a problem found during verification following full gcs.  This recomputes the generation usage at the end of full GC as part of a 5th phase of full gc.

We may be able to remove this 5th phase at a later time after we find and fix whatever errors exist in generation accounting in the existing full gc implementation.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/235/head:pull/235` \
`$ git checkout pull/235`

`$ git checkout pull/235` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/235/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 235`

`$ git pr show -t 235`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/235.diff">https://git.openjdk.org/shenandoah/pull/235.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
@@ -322,6 +322,10 @@ void ShenandoahFullGC::do_it(GCCause::Cause gc_cause) {
     phase3_update_references();
 
     phase4_compact_objects(worker_slices);
+
+    if (heap->mode()->is_generational()) {
+      phase5_restore_generation_accounts();
+    }
   }
 
   {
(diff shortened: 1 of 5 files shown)


src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
```diff
@@ -1478,3 +1482,61 @@ void ShenandoahFullGC::phase4_compact_objects(ShenandoahHeapRegionSet** worker_s
 
   heap->clear_cancelled_gc(true /* clear oom handler */);
 }
+
+void ShenandoahFullGC::phase5_restore_generation_accounts() {
```
earthling-amzn, line 1486: Maybe call this `phase5_restore_generation_usage` for consistency with `establish_usage`?

src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
```diff
@@ -1478,3 +1482,61 @@ void ShenandoahFullGC::phase4_compact_objects(ShenandoahHeapRegionSet** worker_s
...
+
+  for (size_t i = 0; i < num_regions; i++) {
+    switch (heap->region_affiliation(i)) {
+      case ShenandoahRegionAffiliation::FREE:
+        break;
+      case ShenandoahRegionAffiliation::YOUNG_GENERATION:
+        r = heap->get_region(i);
+        young_regions++;
+        young_usage += r->used();
+        if (r->is_humongous()) {
```
ysramakrishna, line 1506: Two nits:

1. There is code in the two arms here that can be consolidated into a work method that can then be used to update the appropriate young or old metric.
2. It's conceptually simpler to check if is_humongous() and then add size of object if is_humongous_start. You can there also do arithmetic for the waste in the last humongous region for that object. Then you can just skip all the is_humongous_cont() regions, and save some time.

Other than that, this looks good to me.

ysramakrishna (approved):
Two small nits, but can be ignored if this is "stop-gap" code for now anyway.

Reviewed!

shipilev (commented):
I think `phase4_compact_objects` does the "epilog" for the Full GC code, including setting the region and heap properties. Stands to reason the generation properties fixups should be there too? Also, `phase4_compact_objects` does "FullGC done: GLOBAL usage ..." logging, which in this version happens before we establish the usages for generations?

I'd propose to have the same `phase5_...` method you have here, but move everything starting from `// Reset complete bitmap...` from `phase4_...` into it.

earthling-amzn (approved):

ysramakrishna (approved):

src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
kdnilsen: I made this change, and then @shipilev suggested another change below.

PR #236, opened 2023-03-29
Abandon mixed collections if all candidates are pinned

earthling-amzn:
A new `WAITING_FOR_FILL` state has been added. This state is now entered when all the candidates for a mixed collection cycle are found to be pinned (which should be quite rare). This state permits a new old generation collection cycle to begin (which itself begins with making these regions parseable).

This change also adds a new diagnostic flag `ShenandoahCoalesceChance` which is the probability (expressed as a percentage) of abandoning all the candidates of a mixed collection. Any abandoned candidate regions will be made parseable.

### DIFF
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
@@ -112,7 +112,20 @@ bool ShenandoahOldHeuristics::prime_collection_set(ShenandoahCollectionSet* coll
   }
 
   if (unprocessed_old_collection_candidates() == 0) {
+    // We have added the last of our collection candidates to a mixed collection.
     _old_generation->transition_to(ShenandoahOldGeneration::IDLE);
+  } else if (included_old_regions == 0) {
+    // We have candidates, but none were included for evacuation - they must all be pinned.
+    // We don't want a region with a stuck pin to prevent subsequent old collections, so
+    // we transition to a state that will allow us to make these uncollected (pinned) regions
+    // parseable.
+#ifdef ASSERT
+    for (uint i = _next_old_collection_candidate; i < _last_old_collection_candidate; ++i) {
+      auto region = _region_data[i]._region;
+      assert(region->is_pinned(), "Expected region " SIZE_FORMAT " to be pinned.", region->index());
+    }
+#endif
+    _old_generation->transition_to(ShenandoahOldGeneration::WAITING_FOR_FILL);
   }
 
   return (included_old_regions > 0);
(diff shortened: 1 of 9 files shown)


src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
```diff
@@ -451,6 +451,12 @@
   product(bool, ShenandoahAllocFailureALot, false, DIAGNOSTIC,              \
           "Testing: make lots of artificial allocation failures.")          \
                                                                             \
+  product(uintx, ShenandoahCoalesceChance, 0, DIAGNOSTIC,                   \
```
kdnilsen, line 454: Should this read: "Abandon remaining mixed collections with this likelihood.  Following each mixed collection, we abandon all remaining mixed collection candidate regions with likelihood ShenandoahCoalesceChance." ?  (I don't think this is precisely abandon this percentage of mixed collections".)

kdnilsen (approved):
this is nice.  thanks.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
earthling-amzn: Yes, that's a more accurate description. I'll update this description.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
```diff
@@ -255,6 +255,9 @@ bool ShenandoahOldGeneration::coalesce_and_fill() {
   uint nworkers = workers->active_workers();
 
   log_debug(gc)("Starting (or resuming) coalesce-and-fill of old heap regions");
+  // This code will see the same set of regions to filled on each resumption as it did
+  // on the initial run. That's okay because each region keeps track of its own coalesce
+  // and fill state. Region that were filled on a prior attempt will not try to fill again.
```
ysramakrishna, line 260: Regions

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -112,12 +112,39 @@ bool ShenandoahOldHeuristics::prime_collection_set(ShenandoahCollectionSet* coll
   }
 
   if (unprocessed_old_collection_candidates() == 0) {
+    // We have added the last of our collection candidates to a mixed collection.
     _old_generation->transition_to(ShenandoahOldGeneration::IDLE);
+  } else if (included_old_regions == 0) {
+    // We have candidates, but none were included for evacuation - are they all be pinned?
```
ysramakrishna, line 118: are they all pinned?  

(no be)

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
```diff
@@ -386,29 +391,34 @@ bool ShenandoahOldGeneration::validate_transition(State new_state) {
   ShenandoahHeap* heap = ShenandoahHeap::heap();
```
ysramakrishna, line 391: Could you update the ascii art of the sate transition diagram to include WAITING_FOR* states instead of the old WAITING state?

ysramakrishna (approved):
LGTM!

PR #238, opened 2023-03-30
JDK-8305334: GenShen: reference processing needs a card-marking barrier

ysramakrishna:
Global collections can create new cross-generational pointers during j.l.r.Reference processing which should be added to the card marking remembered set. The issue was found with dacapo during heap verification and happens somewhat rarely. I added the requisite barrier and provided a comment describing the sole situation in reference processing that should need the barrier. Assertions check this condition, but the card is also redundantly dirtied for young collections too where it's strictly not needed.

I have tested with and without heap verification using product and fastdebug builds using dacapo and specjbb, and through our internal tests pipelines.

I have not run any specific performance comparisons to assess the impact of the checks for applications that traffic heavily in j.l.r.Refs, but I do not expect much impact.

---------
### Progress

### Issue
 * [JDK-8305334](https://bugs.openjdk.org/browse/JDK-8305334): GenShen: reference processing needs a card-marking barrier

### Reviewers
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [cca3d009](https://git.openjdk.org/shenandoah/pull/238/files/cca3d009e66f7b36b581f62d1c8b4f871794130e)
 * [William Kemper](https://openjdk.org/census#wkemper) (@earthling-amzn - Committer) ⚠️ Review applies to [cca3d009](https://git.openjdk.org/shenandoah/pull/238/files/cca3d009e66f7b36b581f62d1c8b4f871794130e)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/238/head:pull/238` \
`$ git checkout pull/238`

`$ git checkout pull/238` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/238/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 238`

`$ git pr show -t 238`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/238.diff">https://git.openjdk.org/shenandoah/pull/238.diff</a>

</details>


src/hotspot/share/gc/shenandoah/shenandoahReferenceProcessor.cpp
```diff
@@ -58,17 +59,44 @@ static const char* reference_type_name(ReferenceType type) {
   }
 }
 
+template <typename T>
```
earthling-amzn, line 62: This looks similar to the code we have for dropping a reference.
https://github.com/openjdk/shenandoah/blob/2785875de3bf86ab20779c5a24ec7e0f69e452aa/src/hotspot/share/gc/shenandoah/shenandoahReferenceProcessor.cpp#L397

Could we use this function there?

ysramakrishna: Let me look into this a bit more closely, in regards to how the card-scanning code treats the traversal of weak pointers.

In other words, it must either be the case that the card-mark done in the code you pointed out above (in `drop`) is not needed or, if it is needed, then at least my comment (and attendant assertion at lines 81 & 82) below is wrong and needs to be suitably adjusted and retested:
```
        Those cards will continue to
    // remain dirty on account of this cross-generational pointer to the referent.
```

I'll update after looking into this more closely. (I'll also then see if the code can be shared like you asked.)

ysramakrishna: A visual examination of the card-scanning code reveals that the card-mark in the portion of code you pointed out is redundant because the card-scanning code always dirties the card containing a cross-generational pointer, irrespective of whether it's a weak or strong pointer. I'll work on eliding that unnecessary check and mark in a separate PR as I want to add some assertions and check it carefully.

Thanks for pointing out this code which I had just glossed over until you pointed it out to me above.

The diffs in my original PR stay unchanged and I'll not try to do anything about sharing the code, since I expect that piece of card-marking code in `drop()` to go away in the PR I promised above, for which I'll open a ticket momentarily.

ysramakrishna: I opened https://bugs.openjdk.org/browse/JDK-8305335 to follow up on that.

kdnilsen (approved):

earthling-amzn (approved):

ysramakrishna:
Thanks @kdnilsen & @earthling-amzn : Could one of you /sponsor then I can land the change. Thanks!

/summary

Global collections can create new cross-generational pointers during j.l.r.Reference processing which should be added to the card marking remembered set. The issue was found with dacapo during heap verification and happens somewhat rarely. I added the requisite barrier and provided a comment describing the sole situation in reference processing that should need the barrier. Assertions check this condition, but the card is also redundantly dirtied for young collections too where it's strictly not needed.

I have tested with and without heap verification using product and fastdebug builds using dacapo and specjbb, and through our internal tests pipelines.

I have not run any specific performance comparisons to assess the impact of the checks for applications that traffic heavily in j.l.r.Refs, but I do not expect much impact.

/integrate

src/hotspot/share/gc/shenandoah/shenandoahReferenceProcessor.cpp
ysramakrishna: > Maybe that can be done in a separate PR? (I'm not sure the commentary of this method is entirely relevant to that other situation. If it is, feel free to move forward with testing of the change...

Correct; we'll follow up separately on that in a separate PR & ticket, linked i comment above.

PR #241, opened 2023-04-03
Remove assertions constraining card marking for reference processing

earthling-amzn:
Shenandoah does _referent_ based processing for weak reference processing. We may see a mix of young and old references on the discovered list. We therefore need to mark cards for old references linked on the discovered list to young references to satisfy remembered set verification. These references may be mixed in any collection cycle.

---------
### Progress

### Reviewers
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [1185eabf](https://git.openjdk.org/shenandoah/pull/241/files/1185eabfa2705a92bbaa817c2c8834fb8de58677)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/241/head:pull/241` \
`$ git checkout pull/241`

`$ git checkout pull/241` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/241/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 241`

`$ git pr show -t 241`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/241.diff">https://git.openjdk.org/shenandoah/pull/241.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/241#issuecomment-1496225381)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahReferenceProcessor.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahReferenceProcessor.cpp
@@ -69,7 +69,8 @@ static void card_mark_barrier(T* field, oop value) {
     // of old and young _references_. These references are linked together through the
     // discovered field in java.lang.Reference. In some cases, creating or editing this
     // list may result in the creation of _new_ old-to-young pointers which must dirty
-    // the corresponding card. Failing to do this may cause heap verification errors.
+    // the corresponding card. Failing to do this may cause heap verification errors and
+    // lead to incorrect GC behavior.
     heap->card_scan()->mark_card_as_dirty(reinterpret_cast<HeapWord*>(field));
   }
 }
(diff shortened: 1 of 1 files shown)


kdnilsen (approved):
(On the comment: we say failure to dirty the card may result in verification errors.  I think it may also result in failure to "take care" of interesting pointers, which can cause GC failures including crashes...  Or maybe this is strictly a verification problem, because the reference list is discarded before GC has to deal with these interesting pointers?)

earthling-amzn:
> (On the comment: we say failure to dirty the card may result in verification errors. I think it may also result in failure to "take care" of interesting pointers, which can cause GC failures including crashes... Or maybe this is strictly a verification problem, because the reference list is discarded before GC has to deal with these interesting pointers?)

I will update the comment. I think if the verifier can see this, it means the object graph had an old->young pointer that wasn't marked _at a safepoint_, which could lead to problems for subsequent GC phases or cycles.

PR #242, opened 2023-04-04
Disable redundant generation usage after fullgc

kdnilsen:
After a recent integration, we ended up recomputing generation usage twice.  This change integrates the two efforts with a single more efficient pass.


src/hotspot/share/gc/shenandoah/shenandoahReferenceProcessor.cpp
earthling-amzn, line 1: Do you want to leave this out of this PR? It's somewhat unrelated and there is a separate PR for it: https://github.com/openjdk/shenandoah/pull/241

kdnilsen: I'll remove this.  thanks.  (i had to comment out asserts to pass regression tests.)

earthling-amzn (approved):

PR #246, opened 2023-04-06
Move affiliation and mode out of ShenandoahGenerationalMode

shipilev:
`ShenandoahGenerationalMode` currently carries `ShenandoahRegionAffiliation` and `GenerationMode` definitions, plus some utility methods. This forces inclusions of `shenandoahGenerationalMode.hpp` in weird places. Plus, we expose generic helper methods in global symbol table, which might clash with symbols from other compilation units, producing linkage errors. We have the unspoken rule to prepend every Shenandoah-related global symbol with `(s|S)henandoah` to avoid this.

This PR moves `ShenandoahRegionAffiliation` into separate header file, includes it in few places. The usages are rewritten to use helper methods, which drastically limits the number of files we need to include new header into. It also renames it to `ShenandoahAffiliation`, since I believe there is not ambiguity here and shorter identifiers are more readable.

The same is done with `GenerationMode`, but PR cuts it a bit deeper by renaming it to `ShenandoahGenerationType`.

Most substitutions were done mechanically, but please review those with care. Thanks!

Testing:
 - [x] macos-aarch64-server-fastdebug build
 - [x] macos-aarch64-server-fastdebug `hotspot_gc_shenandoah` (some pre-existing failures)

---------
### Progress

### Reviewers
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [7c2bc479](https://git.openjdk.org/shenandoah/pull/246/files/7c2bc47949c2f3d819af86ecb479528857e481aa)
 * [William Kemper](https://openjdk.org/census#wkemper) (@earthling-amzn - Committer) ⚠️ Review applies to [7c2bc479](https://git.openjdk.org/shenandoah/pull/246/files/7c2bc47949c2f3d819af86ecb479528857e481aa)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/246/head:pull/246` \
`$ git checkout pull/246`

`$ git checkout pull/246` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/246/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 246`

`$ git pr show -t 246`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/246.diff">https://git.openjdk.org/shenandoah/pull/246.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/246#issuecomment-1498936429)


src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
```diff
@@ -181,7 +181,8 @@ const char* ShenandoahOldGeneration::name() const {
 }
 
 bool ShenandoahOldGeneration::contains(ShenandoahHeapRegion* region) const {
-  return region->affiliation() != YOUNG_GENERATION;
+  // TODO: Should this be region->is_old() instead?
+  return !region->is_young();
```
kdnilsen, line 185: Agree with proposed change.  How could this not be a problem?   Is it used?

kdnilsen (approved):

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
earthling-amzn: We have to be careful with this one because some uses expect to visit regions that are both old _and_ free (or young _and_ free). Maybe we should get rid of this method and be explicit about it?

src/hotspot/share/gc/shenandoah/shenandoahControlThread.hpp
```diff
@@ -90,7 +90,7 @@ class ShenandoahControlThread: public ConcurrentGCThread {
   ShenandoahSharedFlag _do_counters_update;
   ShenandoahSharedFlag _force_counters_update;
   GCCause::Cause       _requested_gc_cause;
-  GenerationMode       _requested_generation;
+  ShenandoahGenerationType       _requested_generation;
```
earthling-amzn, line 93: White space got a little weird here.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
shipilev: Yeah, I would have expected `ShenandoahOldGeneration::contains` to look like the alias for `is_old`. I think `ShenandoahYoungGeneration::contains` has the opposite problem. That is something for a separate PR.

src/hotspot/share/gc/shenandoah/shenandoahControlThread.hpp
shipilev: Good spot, fixed. (In not yet published commit.)

earthling-amzn (approved):

PR #248, opened 2023-04-06
DRAFT: Expand old on demand

kdnilsen:
This PR describes several proposed changes to dynamically adjust the sizes of old-gen and young-gen.  In general, the objective is to keep old-gen as small as possible so that there is an abundance of memory available for the young-gen allocation runway.

As currently drafted, there are regression failures.  This DRAFT PR is published for the purpose of facilitating a careful code review.

### DIFF
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
@@ -30,6 +30,8 @@
 #include "gc/shenandoah/shenandoahFreeSet.hpp"
 #include "gc/shenandoah/shenandoahHeap.inline.hpp"
 #include "gc/shenandoah/shenandoahGeneration.hpp"
+#include "gc/shenandoah/shenandoahHeapRegion.inline.hpp"
+#include "gc/shenandoah/shenandoahOldGeneration.hpp"
 #include "gc/shenandoah/shenandoahYoungGeneration.hpp"
 #include "logging/log.hpp"
 #include "logging/logTag.hpp"
@@ -55,8 +57,6 @@ const double ShenandoahAdaptiveHeuristics::HIGHEST_EXPECTED_AVAILABLE_AT_END = 0
 // as bounds on the adjustments applied at the outcome of a GC cycle.
 const double ShenandoahAdaptiveHeuristics::MINIMUM_CONFIDENCE = 0.319; // 25%
 const double ShenandoahAdaptiveHeuristics::MAXIMUM_CONFIDENCE = 3.291; // 99.9%
-
-const uint ShenandoahAdaptiveHeuristics::MINIMUM_RESIZE_INTERVAL = 10;
 
 ShenandoahAdaptiveHeuristics::ShenandoahAdaptiveHeuristics(ShenandoahGeneration* generation) :
   ShenandoahHeuristics(generation),
@@ -106,12 +106,38 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
 
   // Better select garbage-first regions
   QuickSort::sort<RegionData>(data, (int)size, compare_by_garbage, false);
+  
 
   if (is_generational) {
+#undef KELVIN_CSET
+#ifdef KELVIN_CSET
+    log_info(gc, ergo)("Looking for preselected from 0 .. " SIZE_FORMAT, size);
+#endif
+    for (size_t idx = 0; idx < size; idx++) {
+      ShenandoahHeapRegion* r = data[idx]._region;
+      if (cset->is_preselected(r->index())) {
+        assert(r->age() >= InitialTenuringThreshold, "Preselected regions must have tenure age");
+#ifdef KELVIN_CSET
+        log_info(gc, ergo)("Found preselected region " SIZE_FORMAT " (age: %d, used: " SIZE_FORMAT ", live: " SIZE_FORMAT ")",
+                           r->index(), r->age(), r->used(), r->get_live_data_bytes());
+#endif
+        // Entire region will be promoted, This region does not impact young-gen or old-gen evacuation reserve.
+        // This region has been pre-selected and its impact on promotion reserve is already accounted for.
+
+        // r->used() is r->garbage() + r->get_live_data_bytes()
+        // Since all live data in this region is being evacuated from young-gen, it is as if this memory
+        // is garbage insofar as young-gen is concerned.  Counting this as garbage reduces the need to
+        // reclaim highly utilized young-gen regions just for the sake of finding min_garbage to reclaim
+        // within youn-gen memory.
+
+        cur_young_garbage += r->garbage();
+        cset->add_region(r);
+      }
+    }
     if (is_global) {
       size_t max_young_cset    = (size_t) (heap->get_young_evac_reserve() / ShenandoahEvacWaste);
       size_t young_cur_cset = 0;
-      size_t max_old_cset    = (size_t) (heap->get_old_evac_reserve() / ShenandoahEvacWaste);
+      size_t max_old_cset    = (size_t) (heap->get_old_evac_reserve() / ShenandoahOldEvacWaste);
       size_t old_cur_cset = 0;
       size_t free_target = (capacity * ShenandoahMinFreeThreshold) / 100 + max_young_cset;
       size_t min_garbage = (free_target > actual_free) ? (free_target - actual_free) : 0;
@@ -124,24 +150,31 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
 
       for (size_t idx = 0; idx < size; idx++) {
         ShenandoahHeapRegion* r = data[idx]._region;
+        if (cset->is_preselected(r->index())) {
+#undef KELVIN_GLOBAL
+#ifdef KELVIN_GLOBAL
+          log_info(gc, ergo)("Ignoring preselected %s region " SIZE_FORMAT " (used: " SIZE_FORMAT ", live: " SIZE_FORMAT ")",
+                             affiliation_name(r->affiliation()), r->index(), r->used(), r->get_live_data_bytes());
+#endif
+          continue;
+        }
         bool add_region = false;
         if (r->is_old()) {
           size_t new_cset = old_cur_cset + r->get_live_data_bytes();
           if ((new_cset <= max_old_cset) && (r->garbage() > garbage_threshold)) {
             add_region = true;
             old_cur_cset = new_cset;
+#ifdef KELVIN_GLOBAL
+            log_info(gc, ergo)("Adding old region " SIZE_FORMAT " (used: " SIZE_FORMAT ", live: " SIZE_FORMAT "), old_cur_cset: " SIZE_FORMAT,
+                               r->index(), r->used(), r->get_live_data_bytes(), old_cur_cset);
+#endif
           }
-        } else if (cset->is_preselected(r->index())) {
-          assert(r->age() >= InitialTenuringThreshold, "Preselected regions must have tenure age");
-          // Entire region will be promoted, This region does not impact young-gen or old-gen evacuation reserve.
-          // This region has been pre-selected and its impact on promotion reserve is already accounted for.
-          add_region = true;
-          // r->used() is r->garbage() + r->get_live_data_bytes()
-          // Since all live data in this region is being evacuated from young-gen, it is as if this memory
-          // is garbage insofar as young-gen is concerned.  Counting this as garbage reduces the need to
-          // reclaim highly utilized young-gen regions just for the sake of finding min_garbage to reclaim
-          // within youn-gen memory.
-          cur_young_garbage += r->used();
+#ifdef KELVIN_GLOBAL
+          else {
+            log_info(gc, ergo)("Not adding old region " SIZE_FORMAT " (used: " SIZE_FORMAT ", live: " SIZE_FORMAT "), old_cur_cset: "
+                               SIZE_FORMAT, r->index(), r->used(), r->get_live_data_bytes(), old_cur_cset);
+          }
+#endif
         } else if (r->age() < InitialTenuringThreshold) {
           size_t new_cset = young_cur_cset + r->get_live_data_bytes();
           size_t region_garbage = r->garbage();
@@ -151,8 +184,24 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
             add_region = true;
             young_cur_cset = new_cset;
             cur_young_garbage = new_garbage;
+#ifdef KELVIN_GLOBAL
+            log_info(gc, ergo)("Adding young region " SIZE_FORMAT " (used: " SIZE_FORMAT ", live: " SIZE_FORMAT "), young_cur_cset: " SIZE_FORMAT,
+                               r->index(), r->used(), r->get_live_data_bytes(), young_cur_cset);
+#endif
           }
+#ifdef KELVIN_GLOBAL
+          else {
+            log_info(gc, ergo)("Not adding young region " SIZE_FORMAT " (used: " SIZE_FORMAT ", live: " SIZE_FORMAT "), young_cur_cset: "
+                               SIZE_FORMAT, r->index(), r->used(), r->get_live_data_bytes(), young_cur_cset);
+          }
+#endif
         }
+#ifdef KELVIN_GLOBAL
+        else {
+          log_info(gc, ergo)("Not adding aged young region " SIZE_FORMAT " (age: %d, used: " SIZE_FORMAT ", live: " SIZE_FORMAT
+                             ") because not preselected", r->index(), r->age(), r->used(), r->get_live_data_bytes());
+        }
+#endif
         // Note that we do not add aged regions if they were not pre-selected.  The reason they were not preselected
         // is because there is not sufficient room in old-gen to hold their to-be-promoted live objects.
 
@@ -172,39 +221,46 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
                          byte_size_in_proper_unit(max_cset),    proper_unit_for_byte_size(max_cset),
                          byte_size_in_proper_unit(actual_free), proper_unit_for_byte_size(actual_free));
 
+#ifdef KELVIN_CSET
+      log_info(gc, ergo)(" cur_young_garbage: " SIZE_FORMAT " vs. min_garbage: " SIZE_FORMAT
+                         ", free_target: " SIZE_FORMAT
+                         ", garbage_threshold: " SIZE_FORMAT
+                         ", ignore_threshold: " SIZE_FORMAT ", young_evac_reserve: " SIZE_FORMAT,
+                         cur_young_garbage, min_garbage, free_target, garbage_threshold, ignore_threshold,
+                         heap->get_young_evac_reserve());
+#endif
       for (size_t idx = 0; idx < size; idx++) {
         ShenandoahHeapRegion* r = data[idx]._region;
-        bool add_region = false;
-
-        if (!r->is_old()) {
-          if (cset->is_preselected(r->index())) {
-            assert(r->age() >= InitialTenuringThreshold, "Preselected regions must have tenure age");
-            // Entire region will be promoted, This region does not impact young-gen evacuation reserve.  Memory has already
-            // been set aside to hold evacuation results as advance_promotion_reserve.
-            add_region = true;
-            // Since all live data in this region is being evacuated from young-gen, it is as if this memory
-            // is garbage insofar as young-gen is concerned.  Counting this as garbage reduces the need to
-            // reclaim highly utilized young-gen regions just for the sake of finding min_garbage to reclaim
-            // within youn-gen memory
-            cur_young_garbage += r->get_live_data_bytes();
-          } else if  (r->age() < InitialTenuringThreshold) {
-            size_t new_cset = cur_cset + r->get_live_data_bytes();
-            size_t region_garbage = r->garbage();
-            size_t new_garbage = cur_young_garbage + region_garbage;
-            bool add_regardless = (region_garbage > ignore_threshold) && (new_garbage < min_garbage);
-            if ((new_cset <= max_cset) && (add_regardless || (region_garbage > garbage_threshold))) {
-              add_region = true;
-              cur_cset = new_cset;
-              cur_young_garbage = new_garbage;
-            }
-          }
-          // Note that we do not add aged regions if they were not pre-selected.  The reason they were not preselected
-          // is because there is not sufficient room in old-gen to hold their to-be-promoted live objects.
-
-          if (add_region) {
+        if (cset->is_preselected(r->index())) {
+          continue;
+        }
+        if  (r->age() < InitialTenuringThreshold) {
+          size_t new_cset = cur_cset + r->get_live_data_bytes();
+          size_t region_garbage = r->garbage();
+          size_t new_garbage = cur_young_garbage + region_garbage;
+          bool add_regardless = (region_garbage > ignore_threshold) && (new_garbage < min_garbage);
+          assert(r->is_young(), "Only young candidates expected in the data array");
+          if ((new_cset <= max_cset) && (add_regardless || (region_garbage > garbage_threshold))) {
+#ifdef KELVIN_CSET
+            log_info(gc, ergo)("Adding region " SIZE_FORMAT ": new_cset: " SIZE_FORMAT ", max_cset: " SIZE_FORMAT
+                               ", region_garbage: " SIZE_FORMAT ", region live: " SIZE_FORMAT ", new_garbage: " SIZE_FORMAT,
+                               r->index(), new_cset, max_cset, region_garbage, r->get_live_data_bytes(), new_garbage);
+#endif
+            cur_cset = new_cset;
+            cur_young_garbage = new_garbage;
             cset->add_region(r);
           }
+#ifdef KELVIN_CSET
+          else {
+            log_info(gc, ergo)("Not adding region " SIZE_FORMAT ": new_cset: " SIZE_FORMAT ", max_cset: " SIZE_FORMAT
+                               ", region_garbage: " SIZE_FORMAT ", region live: " SIZE_FORMAT ", new_garbage: " SIZE_FORMAT,
+                               r->index(), new_cset, max_cset, region_garbage, r->get_live_data_bytes(), new_garbage);
+          }
+#endif
         }
+        // Note that we do not add aged regions if they were not pre-selected.  The reason they were not preselected
+        // is because there is not sufficient room in old-gen to hold their to-be-promoted live objects or because
+        // they are to be promted in place.
       }
     }
   } else {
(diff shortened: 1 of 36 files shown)


src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
```diff
@@ -174,37 +184,24 @@ void ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata(Shenand
...
-            }
-          }
-          // Note that we do not add aged regions if they were not pre-selected.  The reason they were not preselected
-          // is because there is not sufficient room in old-gen to hold their to-be-promoted live objects.
-
-          if (add_region) {
+        if (cset->is_preselected(r->index())) {
+          continue;
+        }
+        if  (r->age() < InitialTenuringThreshold) {
```
kdnilsen, line 190: why am i not checking for !r->is_old().  investigate.

kdnilsen: Because we do not add old regions to the candidate array during young and mixed evacuations.  Note that we have an assert below that enforces this region to be young.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
```diff
@@ -76,29 +87,147 @@ ShenandoahHeuristics::~ShenandoahHeuristics() {
...
-    ShenandoahHeapRegion* region = heap->get_region(i);
-    if (in_generation(region) && !region->is_empty() && region->is_regular() && (region->age() >= InitialTenuringThreshold)) {
-      size_t promotion_need = (size_t) (region->get_live_data_bytes() * ShenandoahEvacWaste);
-      if (old_consumed + promotion_need < old_available) {
+    ShenandoahHeapRegion* r = heap->get_region(i);
+    if (r->is_empty() || !r->has_live() || !r->is_young() || !r->is_regular()) {
+      continue;
+    }
+    if (r->age() >= InitialTenuringThreshold) {
+      r->save_top_before_promote();
```
kdnilsen, line 133: We have to save top before promote, even if we do not  promote in place, because evacuator asks "what was garbage before top was saved"?  Need to think about tis...

kdnilsen: the evacuator asks what was top at mark end?

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -215,6 +218,14 @@ class ShenandoahHeap : public CollectedHeap {
...
            size_t _initial_size;
            size_t _minimum_size;
+           size_t _promotion_potential;
+           size_t _promotion_in_place_potential;
+           size_t _pad_for_promote_in_place;    // bytes of filler
+           size_t _promotable_humongous_regions;
+           size_t _promotable_humongous_usage;
+           size_t _regular_regions_promoted_in_place;
+           size_t _regular_usage_promoted_in_place;
+
```
ysramakrishna, line 228: Although these names may in many cases be "self-explanatory" to those familiar with the code, I think they need at least a brief 1-line description each.

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -332,42 +345,21 @@ class ShenandoahHeap : public CollectedHeap {
...
   // effort.  If this happens, the requesting thread blocks until some other thread manages to evacuate the offending object.
   // Only after "all" threads fail to evacuate an object do we consider the evacuation effort to have failed.
 
-  intptr_t _alloc_supplement_reserve;  // Bytes reserved for young allocations during evac and update refs
   size_t _promoted_reserve;            // Bytes reserved within old-gen to hold the results of promotion
   volatile size_t _promoted_expended;  // Bytes of old-gen memory expended on promotions
 
+  // Allocation of old GCLABs (aka PLABs) assures that _old_evac_expended + request-size < _old_evac_reserved.  If the allocation
+  //  is authorized, increment _old_evac_expended by request size.  This allocation ignores old_gen->available().
+
```
ysramakrishna, line 362: This documentation comment seems like a stream-of-consciousness thing. I think we need this comment to move to the right spot.

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -517,11 +526,11 @@ class ShenandoahHeap : public CollectedHeap {
   void update_heap_references(bool concurrent);
   // Final update region states
   void update_heap_region_states(bool concurrent);
-  void rebuild_free_set(bool concurrent);
 
   void rendezvous_threads();
   void recycle_trash();
 public:
+  void rebuild_free_set(bool concurrent);
```
ysramakrishna, line 533: Please document the method's public specification so a caller is aware as to what they should expect the method to do.

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -698,6 +707,10 @@ class ShenandoahHeap : public CollectedHeap {
 // ---------- Allocation support
 //
 private:
+  // How many bytes to transfer between old and young after we have finished recycling collection set regions?
```
ysramakrishna, line 710: General comment (no action needed at this time) about region counts. It appears to me that when talking about quantities that might take positive (`surplus`) or negative (`deficit`) values that are suitably bounded, it might make sense, where possible, to have a single variable of type `ssize_t` and make use of signs for that one variable. Is this such an opportunity?

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -731,6 +744,12 @@ class ShenandoahHeap : public CollectedHeap {
 
   void set_young_lab_region_flags();
 
+  inline void set_old_region_surplus(size_t surplus) { _old_regions_surplus = surplus; };
```
ysramakrishna, line 747: I'd avoid the unnecessary inconsistency between `set_old_region_surplus` and `_old_regions_surplus`. I prefer both to use the singular form of `region`.

src/hotspot/share/gc/shenandoah/shenandoahControlThread.hpp
```diff
@@ -173,11 +173,11 @@ class ShenandoahControlThread: public ConcurrentGCThread {
   void prepare_for_graceful_shutdown();
   bool in_graceful_shutdown();
 
-  void service_concurrent_normal_cycle(const ShenandoahHeap* heap,
+  void service_concurrent_normal_cycle(ShenandoahHeap* heap,
```
ysramakrishna, line 176: Isn't this delta already there in another PR? I'd rebase this PR wrt that as parent, or land that and refresh this so as to avoid presenting the same diffs in two different PRs where possible.

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.hpp
```diff
@@ -50,16 +50,31 @@ class ShenandoahMmuTask;
  * MMU.
```
ysramakrishna, line 50: I believe these have already been split out into a separate PR. Let's drop these from this PR by producing a dependent PR to avoid having to look at these diffs again.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
```diff
@@ -97,7 +97,7 @@
           "collector accepts. In percents of heap region size.")            \
           range(0,100)                                                      \
                                                                             \
-  product(uintx, ShenandoahOldGarbageThreshold, 10, EXPERIMENTAL,           \
+  product(uintx, ShenandoahOldGarbageThreshold, 15, EXPERIMENTAL,           \
           "How much garbage an old region has to contain before it would "  \
           "be taken for collection.")                                       \
```
ysramakrishna, line 102:
```
"An old region must contain at least this percentage of garbage before being eligible for evacuation."
```
What happens when the user specifies a number like 0 or 100? Does it have the requisite effect? For 0, I would expect the region to be eligible for evacuation even when there is no garbage in it and, for 100, I would expect it to be collected only when it is fully garbage and not otherwise. Is this a reasonable expectation? Are there tests that check for this?

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
```diff
@@ -128,13 +128,6 @@
           "size.")                                                          \
           range(0,100)                                                      \
                                                                             \
-  product(uintx, ShenandoahOldMinFreeThreshold, 5, EXPERIMENTAL,            \
-          "Percentage of free old generation heap memory below which most " \
-          "heuristics trigger collection independent of other triggers. "   \
-          "Provides a safety margin for many heuristics. In percents of "   \
-          "(soft) max heap size.")                                          \
-          range(0,100)                                                      \
-                                                                            \
```
ysramakrishna, line 137: Would it be correct to assume that this behaviour is no longer present or, if present, no longer directly controllable in this manner?

It'd be great to have a PR comment against each of these knob changes motivating the changes to their default values, or of their no longer being available.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
```diff
@@ -210,8 +203,8 @@
           "Heuristics may trigger collections more frequently. Time is in " \
           "milliseconds. Setting this to 0 disables the feature.")          \
                                                                             \
-  product(uintx, ShenandoahGuaranteedYoungGCInterval, 5*60*1000, EXPERIMENTAL, \
-          "Run a collection of the young generation at least this often. "  \
+  product(uintx, ShenandoahGuaranteedYoungGCInterval, 30*1000,  EXPERIMENTAL,  \
```
ysramakrishna, line 206: Why was this made so small? Are there cases where, say, for a quiescent/idle service instance with little or no load, we must still, for some reason, trigger young gc's twice each minute?

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
```diff
@@ -378,7 +364,7 @@
           "to be less than this.")                                          \
           range(0, 100)                                                     \
                                                                             \
-  product(uintx, ShenandoahMaxYoungPercentage, 80, EXPERIMENTAL,            \
+  product(uintx, ShenandoahMaxYoungPercentage, 100, EXPERIMENTAL,           \
```
ysramakrishna, line 367: If a user sets MinYoung == MaxYoung == 100, does one get single-generation Shenandoah?  Or is that request rejected? What if both are set to 0? Are there sanity checks or tests that would fail today with such settings?

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
```diff
@@ -339,26 +335,16 @@
...
-                                                                            \
-  product(uintx, ShenandoahOldEvacRatioPercent, 12, EXPERIMENTAL,           \
+  product(uintx, ShenandoahOldEvacRatioPercent, 75, EXPERIMENTAL,           \
           "The maximum proportion of evacuation from old-gen memory, as "   \
-          "a percent ratio.  The default value 12 denotes that no more "    \
-          "than one eighth (12%) of the collection set evacuation "         \
-          "workload may be comprised of old-gen heap regions.  A larger "   \
-          "value allows a smaller number of mixed evacuations to process "  \
+          "a percent ratio.  The default value 75 denotes that no more "    \
+          "than 75% of the collection set evacuation "                      \
```
ysramakrishna, line 341: Is the underlying quantity the number of bytes eligible for evacuation or the number of regions eligible for evacuation? Perhaps make that more clear in the description string.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
```diff
@@ -300,17 +293,20 @@
...
+          "GC cycle.  Smaller values increase the risk of evacuation "      \
+          "failures, which will trigger stop-the-world Full GC passes.")    \
+          range(1.0,100.0)                                                  \
+                                                                            \
+  product(double, ShenandoahPromoEvacWaste, 1.2, EXPERIMENTAL,              \
+          "How much waste evacuations produce within the reserved space. "  \
+          "Larger values make evacuations more resilient against "          \
+          "evacuation conflicts, at expense of evacuating less on each "    \
+          "GC cycle.  Smaller values increase the risk of evacuation "      \
+          "failures, which will trigger stop-the-world Full GC passes.")    \
```
ysramakrishna, line 309: The terms `waste` and `conflicts` are not clear to me. Can we write this in a way that someone not familiar with the internals of the collector can understand without having to dig through the code to understand what `waste` and `evacuation conflict` are? Or include just enough for them to be able to use these knobs with some amount of instinct rather than knob-twiddling "because it was there".

It looks to me like a "magic number estimate" that the user produces out of thin air that the collector then uses in its accounting, and then real life unfolds in such a manner that this determines risk vs reward or efficiency of space usage. In that sense this is more like a "magic padding factor". If so, I'd describe in a line or two what makes the actual space usage more than the size of the evacuated objects, why it's difficult to accurately assess ahead of time, and why a larger value potentially wastes space by allowing more headroom while avoiding running out of space while a smaller value wastes less space while running a greater risk of running out of space in tight situations where it attempted an evacuation.

The talk of waste as related to "conflict" is a bit confusing to me at least without knowing what conflict is and why it must lead to wastage of space.

This probably needs more discussion to see why we may not be able to learn something better based on measurement as the collections unfold, based on an initial rough estimate to begin with and then refining it with a suitable padding for uncertainty over time.

test/hotspot/jtreg/ProblemList.txt
```diff
@@ -83,11 +83,16 @@ gc/stress/gclocker/TestGCLockerWithG1.java 8180622 generic-all
 gc/stress/TestJNIBlockFullGC/TestJNIBlockFullGC.java 8192647 generic-all
 gc/stress/TestStressG1Humongous.java 8286554 windows-x64
 
-gc/shenandoah/oom/TestThreadFailure.java 8306335 generic-all
-gc/shenandoah/oom/TestClassLoaderLeak.java 8306336 generic-all
-gc/shenandoah/TestDynamicSoftMaxHeapSize.java#generational 8306333 generic-all
-gc/stress/gclocker/TestGCLockerWithShenandoah.java#generational 8306341 generic-all
-
```
ysramakrishna, line 90: Nice!!

test/hotspot/jtreg/ProblemList.txt
```diff
@@ -83,11 +83,16 @@ gc/stress/gclocker/TestGCLockerWithG1.java 8180622 generic-all
...
+# kelvin wants to try enabling all tests
+# gc/shenandoah/oom/TestClassLoaderLeak.java 8306336 generic-all
+# gc/shenandoah/mxbeans/TestChurnNotifications.java#generational 8306337 generic-all
+# gc/stress/gcold/TestGCOldWithShenandoah.java#generational 8306339 generic-all
+# gc/stress/systemgc/TestSystemGCWithShenandoah.java#generational 8306340 generic-all
+# gc/stress/gclocker/TestGCLockerWithShenandoah.java#generational 8306341 generic-all
+# gc/shenandoah/oom/TestThreadFailure.java 8306335 generic-all
+# gc/shenandoah/oom/TestClassLoaderLeak.java 8306336 generic-all
+# gc/shenandoah/TestDynamicSoftMaxHeapSize.java#generational 8306333 generic-all
+# gc/stress/gclocker/TestGCLockerWithShenandoah.java#generational 8306341 generic-all
```
ysramakrishna, line 95: I'd just delete these when ready to land, and close the tickets as duplicates of the ticket for this PR.

ysramakrishna (changes requested):
Skimmed over some of the diffs and left some comments. Haven't gotten into the details yet, but will once the overall structure and shape of the changes becomes more clear. Most of the current comments are of form/structure rather than of the details which I will defer until Monday.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
kdnilsen: I will revert this change.  My test on the Extremem phased update workload reveals that this does cause some degenerated cycles that are avoided when we have the shorter interval.  Specifically, that workload has a wait-for-start-of-simulation pause that results in a 45 seconds of GC idle time, followed by a late trigger that degenerates during mark and then a back-to-back trigger for the next GC (which is dealing with pent-up demand following the "long" degenerated pause) and this second GC also degenerates during mark.  Following these two degenerated GCs, GenShen is able to recalibrate its heuristic triggers and stay ahead of pace throughout the remainder of the 20 minute simulation.

I believe the reason we are caught off guard here is because we have not yet "learned" the allocation rate.  The first (late) trigger after the 45s idle span is:  Free (3096M) is below minimum threshold (3101M)

Am reverting this change because I recognize that setting this to 30 seconds is arbitrary and not a robust fix to the bigger triggering problem.

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -354,6 +354,9 @@ class ShenandoahHeap : public CollectedHeap {
   // effort.  If this happens, the requesting thread blocks until some other thread manages to evacuate the offending object.
   // Only after "all" threads fail to evacuate an object do we consider the evacuation effort to have failed.
 
+  // How many full-gc cycles have been completed?
+  volatile size_t _completed_fullgc_cycles;
```
earthling-amzn, line 358: `ShenandoahCollectorPolicy` already counts full GC cycles (implicit and  explicit). We could add a `get_full_gc_count` there without duplicating any functionality.

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -390,6 +393,14 @@ class ShenandoahHeap : public CollectedHeap {
 
 public:
 
+  void finish_fullgc() {
+    Atomic::add(&_completed_fullgc_cycles, (size_t) 1);
```
earthling-amzn, line 397: `Atomic::inc`? Also - this will only be called on a safepoint at the end of full gc? Does it need atomic access?

earthling-amzn (changes requested):
I think it would be simpler/fewer changes to pull full gc count of the existing count in `ShenandoahCollectorPolicy`.

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
kdnilsen: Thanks for looking at this.  This is only written from within full gc.  Maybe declaring the variable volatile is sufficient to assure that the "many" mutator threads that access the variable will see any change written by full-gc control thread.  I followed the Atomic pattern out of an abundance of caution.

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
kdnilsen: good idea.  I searched around and couldn't figure out where this was.  I'll pursue this simplification.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -956,51 +959,142 @@ void ShenandoahHeapRegion::set_affiliation(ShenandoahAffiliation new_affiliation
...
+    heap->free_set()->add_old_collector_free_region(this);
+
+    old_gen->increment_affiliated_region_count();
+    old_gen->increase_used(promoted_used);
+
+    // Might as well make the free memory within newly promoted region available to hold promotions that we were not able
+    // to budget for previously.
+    heap->set_promoted_reserve(promo_reserve);
+
+    // TODO: adjust bounds in the free set
```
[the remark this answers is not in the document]
kdnilsen: This comment has been removed.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -956,51 +959,142 @@ void ShenandoahHeapRegion::set_affiliation(ShenandoahAffiliation new_affiliation
...
   }
-  heap->set_affiliation(this, new_affiliation);
+
+  assert(top() == tams, "Cannot promote regions in place if top has advanced beyond TAMS");
+
+  // Since this region may have served previously as OLD, it may hold obsolete object range info.
+  heap->card_scan()->reset_object_range(bottom(), end());
+  heap->card_scan()->mark_range_as_dirty(bottom(), top() - bottom());
+
```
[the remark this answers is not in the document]
kdnilsen: I've placed a TODO comment to replace this code with an existing coalesce-and-fill function.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -927,7 +930,7 @@ size_t ShenandoahHeapRegion::pin_count() const {
   return Atomic::load(&_critical_pins);
 }
 
-void ShenandoahHeapRegion::set_affiliation(ShenandoahAffiliation new_affiliation) {
+void ShenandoahHeapRegion::set_affiliation(ShenandoahAffiliation new_affiliation, bool defer_affiliated_region_count_updates) {
```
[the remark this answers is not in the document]
kdnilsen: We have refactored this code so that we no longer need the defer_affiliated_region_count_updates argument.  It has been removed.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
```diff
@@ -76,29 +87,147 @@ ShenandoahHeuristics::~ShenandoahHeuristics() {
...
+        promo_potential += region_live_data;
       }
-      // Note that we keep going even if one region is excluded from selection.
-      // Subsequent regions may be selected if they have smaller live data.
+      // We keep going even if one region is excluded from selection because we need to accumulate all eligible
+      // regions that are not preselected into promo_potential
     }
   }
+  heap->set_pad_for_promote_in_place(promote_in_place_pad);
+  heap->set_promotion_potential(promo_potential);
```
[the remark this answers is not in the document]
kdnilsen: We have changed the accumulation of promo_potential to not include regions that are "tenure_age - 1" if this is not an aging cycle.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
```diff
@@ -76,29 +87,147 @@ ShenandoahHeuristics::~ShenandoahHeuristics() {
...
+static int compare_by_aged_live(AgedRegionData a, AgedRegionData b) {
+  if (a._live_data < b._live_data)
+    return -1;
+  else if (a._live_data > b._live_data)
+    return 1;
+  else return 0;
+}
+
+// Returns bytes of old-gen memory consumed by selected aged regions
+size_t ShenandoahHeuristics::select_aged_regions(size_t old_available, size_t num_regions, bool preselected_regions[]) {
```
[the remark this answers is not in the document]
kdnilsen: I've changed the name of preselected_regions[] to candidate_regions_for_promotion_by_copy[].  I've also added a comment to explain (and motivate) the behavior of this function.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
```diff
@@ -166,13 +325,29 @@ void ShenandoahHeuristics::choose_collection_set(ShenandoahCollectionSet* collec
...
+          humongous_bytes_promoted += obj->size() * HeapWordSize;
+        }
       }
     } else if (region->is_trash()) {
       // Count in just trashed collection set, during coalesced CM-with-UR
       immediate_regions++;
       immediate_garbage += garbage;
     }
   }
+  heap->reserve_promotable_humongous_regions(humongous_regions_promoted);
```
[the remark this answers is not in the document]
kdnilsen: We may be able to simplify or remove these four reserve_promotable functions in the future.  But we need them for now.  They identify what work needs to be done during evacuation, and may feed into decisions about how to implement abbreviated gc cycles.  When I originally proposed to eliminate these functions, I was thinking we only needed this for updating generation sizes at the end of the GC cycle, but I was wrong.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
```diff
@@ -185,7 +360,9 @@ void ShenandoahHeuristics::choose_collection_set(ShenandoahCollectionSet* collec
   size_t immediate_percent = (total_garbage == 0) ? 0 : (immediate_garbage * 100 / total_garbage);
   collection_set->set_immediate_trash(immediate_garbage);
```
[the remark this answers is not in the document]
kdnilsen: Development on this improvement is underway.  The improvement will be incorporated after this PR is integrated.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.hpp
```diff
@@ -169,6 +173,8 @@ class ShenandoahHeuristics : public CHeapObj<mtGC> {
   virtual bool is_experimental() = 0;
   virtual void initialize();
 
+  virtual size_t evac_slack(size_t region_to_be_recycled);
```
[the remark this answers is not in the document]
kdnilsen: I'm renaming evac_slack() to bytes_of_allocation_runway_before_gc_trigger().  I'm renaming the argument young_regions_to_be_reclaimed.  I'm adding comments to improve the description of this function and its argument.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
```diff
@@ -324,6 +328,80 @@ static double saturate(double value, double min, double max) {
   return MAX2(MIN2(value, max), min);
 }
 
+// Return conservative estimate of how much memory can be allocated before we need to start GC
+size_t ShenandoahAdaptiveHeuristics::evac_slack(size_t young_regions_to_be_reclaimed) {
```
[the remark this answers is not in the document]
kdnilsen: It's not strictly just the amount of memory available.  It's the amount of memory that can be allocated before the next GC trigger.  The next GC trigger is supposed to happen with enough additional allocation runway that we can finish that GC before the runway is exhausted.  I'm changing the name of the function and adding some comments to hopefully make all of this more clear.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
```diff
@@ -348,137 +426,146 @@ bool ShenandoahAdaptiveHeuristics::should_start_gc() {
...
+                     _generation->name(), byte_size_in_proper_unit(promo_potential), proper_unit_for_byte_size(promo_potential));
+        return true;
+      } else if (promo_in_place_potential > 0) {
+        // Detect unsigned arithmetic underflow
+        assert(promo_in_place_potential < heap->capacity(), "Sanity");
+        log_info(gc)("Trigger (%s): expedite promotion in place of " SIZE_FORMAT "%s", _generation->name(),
+                     byte_size_in_proper_unit(promo_in_place_potential),
+                     proper_unit_for_byte_size(promo_in_place_potential));
+        return true;
+      } else if (mixed_candidates > 0) {
```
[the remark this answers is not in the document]
kdnilsen: Yes.  I think it's what we want.  It's basically our "compensation" for the fact that we do not have a truly concurrent old-gen collector.  If the old-collector threads were concurrent, they would be consuming all available idle cycles to take care of the mixed evacuations.

Certainly, this is an area that we can experiment with and auto-tune if appropriate.  My experiments with Phased-Updates workload suggested that if we have a bunch of mixed evacuation work to be done, we want to get it done as possible.  Otherwise, we can get caught flat footed if mutator demand for new allocations spikes.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -382,10 +406,42 @@ bool ShenandoahOldHeuristics::should_start_gc() {
     return false;
   }
 
-  // If there's been a promotion failure (and we don't have regions already scheduled for evacuation),
-  // start a new old generation collection.
-  if (_promotion_failed) {
-    log_info(gc)("Trigger: Promotion Failure");
+  if (_cannot_expand_trigger) {
+    ShenandoahHeap* heap = ShenandoahHeap::heap();
+    ShenandoahOldGeneration* old_gen = heap->old_generation();
```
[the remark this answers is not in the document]
kdnilsen: Thanks.  Replacing throughout.

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
kdnilsen: I've made this change as suggested by you.

ysramakrishna (approved):
Let's get this integrated as is, without doing any further extensive refactorings at this stage, and we can deal with more detailed or low-level review feedback in the fullness of time.

Such review comments may appear in the future and can be fixed later.

Thanks for the improvements in performance that these changes provided!

PR #250, opened 2023-04-09
Add generations to freeset

kdnilsen:
ShenandoahFreeSet has not yet been modified to deal efficiently with the combination of old-gen and young-gen collection set reserves.  This PR makes changes so that we can distinguish between collector_is_free, old_collector_is_free, and mutator_is_free.  Further, it endeavors to keep each set of free regions tightly packed, so the range of regions representing each set is small.

In its current form, this no longer fails existing regression tests (except for known problems that are being addressed independently)

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/250/head:pull/250` \
`$ git checkout pull/250`

`$ git checkout pull/250` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/250/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 250`

`$ git pr show -t 250`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/250.diff">https://git.openjdk.org/shenandoah/pull/250.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/250#issuecomment-1501243026)


src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -438,25 +505,69 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
...
+        HeapWord* fill_addr = r->top();
+        size_t fill_size = waste / HeapWordSize;
+        ShenandoahHeap::fill_with_object(fill_addr, fill_size);
+        r->set_top(r->end());
+	// Since we have filled the waste with an empty object, account for increased usage
+	_heap->increase_used(waste);
+	if (_heap->mode()->is_generational()) {
+	  _heap->generation_for(req.affiliation())->increase_used(waste);
+          if (req.affiliation() == ShenandoahRegionAffiliation::OLD_GENERATION) {
+            _heap->card_scan()->register_object(fill_addr);
```
kdnilsen, line 524: I'm asking myself whether this needs a lock.  This code is inside try_allocate_in(), which is called by allocate_single() and allocate_with_affiliation() and allocate_old_with_affiliation(), which are called by allocate(), which is called by ShenandoahHeap::allocate_memory_under_lock();  so I have the lock.

Other threads may be concurrently registering objects that belong to the same region without acquring a lock, but this would only happen for objects allocated from within a plab, and we know that none of those will touch the same memory as is affected by this operation, so no locking is necessary.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -42,51 +42,124 @@ ShenandoahFreeSet::ShenandoahFreeSet(ShenandoahHeap* heap, size_t max_regions) :
...
-         ", capacity: " SIZE_FORMAT ", num_bytes: " SIZE_FORMAT, _used, _capacity, num_bytes);
+inline bool ShenandoahFreeSet::peek_is_mutator_free(size_t idx) const {
+  return _mutator_free_bitmap.at(idx);
 }
 
-bool ShenandoahFreeSet::is_mutator_free(size_t idx) const {
+inline bool ShenandoahFreeSet::is_mutator_free(size_t idx) const {
   assert (idx < _max, "index is sane: " SIZE_FORMAT " < " SIZE_FORMAT " (left: " SIZE_FORMAT ", right: " SIZE_FORMAT ")",
           idx, _max, _mutator_leftmost, _mutator_rightmost);
+  assert(!_mutator_free_bitmap.at(idx) || (alloc_capacity(ShenandoahHeap::heap()->get_region(idx)) > 0),
```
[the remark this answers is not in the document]
shipilev: Also, for the benefit of fastdebug builds performance (= faster testing), this can be something like:

```
bool is_free = _mutator_free_bitmap.at(idx);
assert(!is_free || has_alloc_capacity(idx), "Mutator free set should contain useful regions");
return is_free;
```

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -437,29 +514,107 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
...
+        adjust_old_collector_bounds_if_touched(idx);
+      } else if (peek_is_collector_free(idx)) {
+        if ((waste > 0) && _heap->mode()->is_generational()) {
+          _heap->young_generation()->increase_used(waste);
+        }
+        // This applies to both generational and non-generational mode
+        clear_collector_free(idx);
+        assert(!is_mutator_free(idx) && !is_old_collector_free(idx), "Region cannot be in multiple free sets");
+        adjust_collector_bounds_if_touched(idx);
+      }
```
shipilev, line 559: Do we ever get to generic `else` branch here? If not, we should assert that we are not expected to be there.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -468,9 +623,56 @@ void ShenandoahFreeSet::recompute_bounds() {
...
+        old_collector_available_in_first_half += r->free();
+      }
+    }
+    for (size_t index = old_collector_middle; index <= _old_collector_rightmost; index++) {
+      if (is_old_collector_free(index)) {
+        ShenandoahHeapRegion* r = _heap->get_region(index);
+        old_collector_available_in_second_half += r->free();
+      }
+    }
+    _old_collector_search_left_to_right = (old_collector_available_in_second_half > old_collector_available_in_first_half);
```
shipilev, line 648: So, if there are more regions available in the "left" (first) half, then we search from *right* to left? Is this the intent? We need to drop a comment here why.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -468,9 +623,56 @@ void ShenandoahFreeSet::recompute_bounds() {
...
+  }
+}
+
+bool ShenandoahFreeSet::expand_collector_bounds_maybe(size_t idx) {
```
shipilev, line 652: For this and the following method, the return value is not used, can be dropped.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -688,67 +918,187 @@ void ShenandoahFreeSet::rebuild() {
...
+      } else if (is_mutator_free(i)) {
+        assert(!r->is_old(), "Old regions should not be in mutator_free set");
+        buffer[idx] = 'm';
+      } else if (is_collector_free(i)) {
+        assert(!r->is_old(), "Old regions should not be in collector_free set");
+        buffer[idx] = 'c';
+      } else if (is_old_collector_free(i)) {
+        buffer[idx] = 'C';
+      }
+      else if (r->is_humongous()) {
```
shipilev, line 1065:
```suggestion
      } else if (r->is_humongous()) {
```

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -317,7 +317,10 @@ class ShenandoahHeap : public CollectedHeap {
     WEAK_ROOTS_BITPOS  = 4,
 
     // Old regions are under marking, still need SATB barriers.
-    OLD_MARKING_BITPOS = 5
+    OLD_MARKING_BITPOS = 5,
+
+    // The evacuation reserves for old-gen and young-gen are available
+    VALID_EVACUATION_RESERVE_QUANTITIES_BITPOS = 6
```
shipilev, line 323: I think we only do these when we want to access the flags from generated code. If we only do the accesses from the GC code, this could be just a plan `bool` field.

shipilev (commented):
Cursory review follows

ysramakrishna (changes requested):
I thought I'd leave the comments so far, and finish the rest later today.

At a high level, I'd treat the old, young, etc. as being mutually exclusive subsets of the set of all free regions. Then usage of bitmaps for the set membership becomes an implementation detail that need not be exposed everywhere via verbage such as "set_x_free" "clear_x_free", etc. but rather treating them as set addition and removal operations.

So, `set_x_free` becomes `add_to_x`, `clear_x_free` becomes `remove_from_x`. In genera, abstracting the mathematical set membership concept above the details of its implementation makes the client code easier to read.

I'll look through the remainder of the code and complete the remainder of the review later today.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -34,24 +34,51 @@ class ShenandoahFreeSet : public CHeapObj<mtGC> {
...
+  bool _old_collector_search_left_to_right = true;
+
+
   void assert_bounds() const NOT_DEBUG_RETURN;
 
-  bool is_mutator_free(size_t idx) const;
-  bool is_collector_free(size_t idx) const;
+  inline bool is_mutator_free(size_t idx) const;
+  inline bool is_collector_free(size_t idx) const;
+  inline bool is_old_collector_free(size_t idx) const;
```
ysramakrishna, line 64: Would it be correct to assume (when we are stable between transitions in set membership) that the set of all free regions is tiled by its mutually exclusive subsets `collector_free`, `old_collector_free`, and `mutator_free` ?

Thus, when set membership is not being changed, I'd expect a free region to be in exactly one of these three subsets? If so, then (re `peek` methods below) one might be able to do membership assertion checks before and after transitions in membership (i.e. before removing from one subset and after adding to another subset). That would then avoid having to use `peek` versions of the membership querying methods.

I am trying to get a high level mental model so I can better understand the implementation and usage details. This is likely documented somewhere in `ShenandoahFreeSet`, but I am having trouble locating such documentation in the files I am looking at.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -34,24 +34,51 @@ class ShenandoahFreeSet : public CHeapObj<mtGC> {
   ShenandoahHeap* const _heap;
   CHeapBitMap _mutator_free_bitmap;
   CHeapBitMap _collector_free_bitmap;
+  // We keep the _old_collector regions separate from the young collector regions.  This allows us to pack the old regions
```
ysramakrishna, line 37: Looks like this comment block needs to move down to line 44.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -34,24 +34,51 @@ class ShenandoahFreeSet : public CHeapObj<mtGC> {
...
+  // even less frequently than the young survivor regions.  In generational mode, the young survivor regions are typically
+  // recycled after tenure age GC passes.
+  CHeapBitMap _old_collector_free_bitmap;
   size_t _max;
 
   // Left-most and right-most region indexes. There are no free regions outside
   // of [left-most; right-most] index intervals
   size_t _mutator_leftmost, _mutator_rightmost;
   size_t _collector_leftmost, _collector_rightmost;
+  size_t _old_collector_leftmost, _old_collector_rightmost;
```
ysramakrishna, line 48: Would suggest encapsulation in an interval struct with left & right for each of the three intervals.

At that point, it would make sense to talk about whether these intervals can overlap (because of interspersing of the range of indices in a specific subset), and whether there are invariants/properties of the union of the intervals that can be checked as invariants after these sub-intervals are adjusted.

ysramakrishna: I see that some invariants are checked in `assert_bounds` method. A high-level articulation of those invariants before that method would be useful.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -688,67 +918,187 @@ void ShenandoahFreeSet::rebuild() {
...
+        _capacity -= ac;
+        _old_capacity += ac;
+        log_debug(gc, free)("  Shifting region " SIZE_FORMAT " from mutator_free to old_collector_free", idx);
+      } else if (reserved < to_reserve) {
+        // Note: In a previous implementation, regions were only placed into the survivor space (collector_is_free) if
+        // they were entirely empty.  I'm not sure I understand the rational for that.  That alternative behavior would
+        // tend to mix survivor objects with ephemeral objects, making it more difficult to reclaim the memory for the
+        // ephemeral objects.  It also delays aging of regions, causing promotion in place to be delayed.
+        clear_mutator_free(idx);
+        set_collector_free(idx);
```
ysramakrishna, line 1002: Is there a situation when a region can be both in collector and in mutator free sets?

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -688,67 +918,187 @@ void ShenandoahFreeSet::rebuild() {
...
-      log_debug(gc, free)("  Shifting Region " SIZE_FORMAT " (Free: " SIZE_FORMAT "%s, Used: " SIZE_FORMAT "%s) to collector free set",
-                          idx, byte_size_in_proper_unit(region->free()), proper_unit_for_byte_size(region->free()),
-                               byte_size_in_proper_unit(region->used()), proper_unit_for_byte_size(region->used()));
+    ShenandoahHeapRegion* r = _heap->get_region(idx);
+    if (_mutator_free_bitmap.at(idx) && (alloc_capacity(r) > 0)) {
+      assert(!r->is_old(), "mutator_is_free regions should not be affiliated OLD");
+      // OLD regions that have available memory are already in the old_collector free set
+      if ((_old_capacity < to_reserve_old) && (r->is_trash() || (r->affiliation() == ShenandoahAffiliation::FREE))) {
+        clear_mutator_free(idx);
+        set_old_collector_free(idx);
```
ysramakrishna, line 991: this would read:

```
  remove_from_mutator(idx);
  add_to_old(idx);
```

or

```
    move_from_mutator_to_old(idx)
```

in the suggested set membership nomenclature.

Also, since we are always dealing with things in `ShenandoahFreeSet`, I'd dispense with the `free` adjective altogether and make the names more compact.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -688,67 +918,187 @@ void ShenandoahFreeSet::rebuild() {
...
+        size_t ac = alloc_capacity(r);
+        _capacity -= ac;
+        _old_capacity += ac;
```
ysramakrishna, line 994: This piecemeal adjustment of capacities following change of set membership indicates that the set abstraction should be pushed down to `old`, `mutator` and `young` (?) subsets, which then carry both membership & capacity that is adjusted at the point of membership changes.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -108,7 +144,7 @@ class ShenandoahFreeSet : public CHeapObj<mtGC> {
 
   void print_on(outputStream* out) const;
 
-  void reserve_regions(size_t to_reserve);
+  void reserve_regions(size_t young_reserve, size_t old_reserve);
```
ysramakrishna, line 147: Documentation comment, preferably here, but if not then in the implementation as a block header.

ysramakrishna (changes requested):
Some more incremental comments.

Will finish the remaining ones in the next batch very soon, but submitting these now as I need to step away briefly.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -688,67 +918,187 @@ void ShenandoahFreeSet::rebuild() {
   log_debug(gc, free)("Rebuilding FreeSet");
   for (size_t idx = 0; idx < _heap->num_regions(); idx++) {
     ShenandoahHeapRegion* region = _heap->get_region(idx);
+
+    // We move all young available regions into mutator_free set and then we take back the regions we need for our
+    // reserve.  This allows us to "compact" the collector_free (survivor) regions at the high end of the heap.
+    clear_mutator_free(idx);
```
ysramakrishna, line 924: The `clear_mutator_free()` isn't consistent with the comment `move all young available regions into mutator_free set`.

May be you can change the documentation comment to state:
```
We examine each region in turn and assign it to the appropriate unique subset based on the following criteria:
... list of criteria used to assign to one or other of the subsets.
```
Such a comment might help clarify the intent of the code below. Also, if the set abstraction were used with assertions moving into the set modification methods, then this method's logic and flow would be much easier to see. There is a leakage of abstraction happening here for the assertions because of the naked use of low-level constructs into the higher level logic of the algorithm for manipulating the set membership of the regions.

ysramakrishna: Actually, I see now that the `clear()` call further above have at line 916 already emptied all of the subsets (or at least that was the intent); thus, the individual removal from each of the subsets here isn't even necessary.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -688,67 +918,187 @@ void ShenandoahFreeSet::rebuild() {
...
+        log_debug(gc, free)(
+          "  Adding Region " SIZE_FORMAT " (Free: " SIZE_FORMAT "%s, Used: " SIZE_FORMAT "%s) to mutator free set",
           idx, byte_size_in_proper_unit(region->free()), proper_unit_for_byte_size(region->free()),
-               byte_size_in_proper_unit(region->used()), proper_unit_for_byte_size(region->used()));
+          byte_size_in_proper_unit(region->used()), proper_unit_for_byte_size(region->used()));
+      }
     }
   }
 
   // Evac reserve: reserve trailing space for evacuations
```
ysramakrishna, line 953: I'd move the individual region assignment loop above into its own work method, and the remainder of the work below that adjusts the reserves into its own work method, if appropriate? Not sure if such a refactor makes sense or not, but might make the code arguable more maintainable/readable?

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -688,67 +918,187 @@ void ShenandoahFreeSet::rebuild() {
...
+      old_reserve = _heap->get_promoted_reserve() + _heap->get_old_evac_reserve();
+    } else {
+      // We are rebuilding at end of GC, so we set aside budgets specified on command line (or defaults)
+      young_reserve = (_heap->young_generation()->max_capacity() * ShenandoahEvacReserve) / 100;
+      old_reserve = MAX2((_heap->old_generation()->max_capacity() * ShenandoahOldEvacReserve) / 100,
+                         ShenandoahOldCompactionReserve * ShenandoahHeapRegion::region_size_bytes());
+    }
   }
-
+  reserve_regions(young_reserve, old_reserve);
```
ysramakrishna, line 976: So I see now that this does a further iteration (below) over the regions until a specific size criteria is met, albeit iterating over the regions in the opposite order. While I have not understood the high level idea of the two loops, one thing to consider if whether the two iterations, one forward and one backward are needed, or if they could somehow be fused into a single iteration. Just tossing it out there as a question that you might be able to answer immediately, probably in the negative.

A brief couple of lines describing the idea of the two loops above and below would be useful.

ysramakrishna (commented):
Left a few more comments.

I'll make another pass over the free set dot cpp file once these comments have been addressed or otherwise resolved, probably in the next version/iteration of this PR.

src/hotspot/share/gc/shenandoah/mode/shenandoahGenerationalMode.hpp
```diff
@@ -30,7 +30,7 @@
 class ShenandoahGenerationalMode : public ShenandoahMode {
 public:
   virtual void initialize_flags() const;
-  virtual ShenandoahHeuristics* initialize_heuristics(ShenandoahGeneration* generation) const;
+  virtual ShenandoahHeuristics* initialize_heuristics(ShenandoahGeneration* generation) const override;
```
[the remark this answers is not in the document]
kdnilsen: Fixing.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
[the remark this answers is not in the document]
kdnilsen: done.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -42,51 +42,124 @@ ShenandoahFreeSet::ShenandoahFreeSet(ShenandoahHeap* heap, size_t max_regions) :
...
+                                                           ShenandoahAllocRequest& req, bool& in_new_region) {
+  size_t rightmost = _old_collector_rightmost;
+  size_t leftmost = _old_collector_leftmost;
+  if (_old_collector_search_left_to_right) {
+    // This mode picks up stragglers following a full GC
+    for (size_t c = leftmost; c <= rightmost; c++) {
+      if (is_old_collector_free(c)) {
+        ShenandoahHeapRegion* r = _heap->get_region(c);
+        assert(r->is_trash() || (r->affiliation() == ShenandoahAffiliation::FREE)
+               || (r->affiliation() == ShenandoahAffiliation::OLD_GENERATION),
```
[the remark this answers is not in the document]
kdnilsen: replaced throughout.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -205,7 +276,11 @@ HeapWord* ShenandoahFreeSet::allocate_single(ShenandoahAllocRequest& req, bool&
         }
         if (allow_new_region) {
           // Then try a free region that is dedicated to GC allocations.
-          result = allocate_with_affiliation(FREE, req, in_new_region);
+          if (req.affiliation() == ShenandoahAffiliation::OLD_GENERATION) {
```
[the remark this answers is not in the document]
kdnilsen: done.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: I've added a comment.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: fixed.

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
```diff
@@ -661,10 +661,15 @@ inline bool ShenandoahHeap::in_collection_set_loc(void* p) const {
   return collection_set()->is_in_loc(p);
 }
 
+
```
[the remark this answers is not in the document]
kdnilsen: thanks.  fixed.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -63,26 +90,35 @@ class ShenandoahFreeSet : public CHeapObj<mtGC> {
...
-  bool has_no_alloc_capacity(ShenandoahHeapRegion *r);
+  bool can_allocate_from(ShenandoahHeapRegion *r) const;
+  size_t alloc_capacity(ShenandoahHeapRegion *r) const;
+  bool has_no_alloc_capacity(ShenandoahHeapRegion *r) const;
 
 public:
   ShenandoahFreeSet(ShenandoahHeap* heap, size_t max_regions);
 
-  // Number of regions dedicated to GC allocations (for evacuation or promotion) that are currently free
+  // Number of regions dedicated to GC allocations (for evacuation) that are currently free
```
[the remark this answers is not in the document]
kdnilsen: In all three cases, this is the number of regions dedicated to "the cause" that are at "partially free".  These regions are not necessarily "fully free".  I've modified the comments to reflect such.

The name "collector_free" is historical.  This predates generational Shenandoah and it is still used for "non-generational mode" of Shenandoah.  I'm inclined to leave it as is.  Though I am quite willing to make the change if that is the "consensus" position.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -63,26 +90,35 @@ class ShenandoahFreeSet : public CHeapObj<mtGC> {
...
+  void flip_to_old_gc(ShenandoahHeapRegion* r);
 
   void recompute_bounds();
   void adjust_bounds();
+  bool adjust_mutator_bounds_if_touched(size_t idx);
+  bool adjust_collector_bounds_if_touched(size_t idx);
+  bool adjust_old_collector_bounds_if_touched(size_t idx);
   bool touches_bounds(size_t num) const;
+  bool expand_collector_bounds_maybe(size_t idx);
+  bool expand_old_collector_bounds_maybe(size_t idx);
```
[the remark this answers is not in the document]
kdnilsen: Done.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -34,24 +34,51 @@ class ShenandoahFreeSet : public CHeapObj<mtGC> {
...
+  inline void clear_mutator_free(size_t idx);
+  inline void clear_collector_free(size_t idx);
+  inline void clear_old_collector_free(size_t idx);
 
   HeapWord* try_allocate_in(ShenandoahHeapRegion* region, ShenandoahAllocRequest& req, bool& in_new_region);
   HeapWord* allocate_with_affiliation(ShenandoahAffiliation affiliation, ShenandoahAllocRequest& req, bool& in_new_region);
-  HeapWord* allocate_with_old_affiliation(ShenandoahAllocRequest& req, bool& in_new_region);
+  HeapWord* allocate_old_with_affiliation(ShenandoahAffiliation affiliation, ShenandoahAllocRequest& req, bool& in_new_region);
 
   // While holding the heap lock, allocate memory for a single object which is to be entirely contained
```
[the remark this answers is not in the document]
kdnilsen: I've added the asserts.  It's a good improvement.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -34,24 +34,51 @@ class ShenandoahFreeSet : public CHeapObj<mtGC> {
...
-  bool is_mutator_free(size_t idx) const;
-  bool is_collector_free(size_t idx) const;
+  inline bool is_mutator_free(size_t idx) const;
+  inline bool is_collector_free(size_t idx) const;
+  inline bool is_old_collector_free(size_t idx) const;
+
+  // Routines that do not assert non-empty free, for use in assertions and during state transitions
+  inline bool peek_is_mutator_free(size_t idx) const;
+  inline bool peek_is_collector_free(size_t idx) const;
+  inline bool peek_is_old_collector_free(size_t idx) const;
```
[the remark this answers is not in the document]
kdnilsen: I understand.  I'm adding some comments to try to clarify.  (maybe a bigger fix can come later to refactor the implementation of how set membership is represented.)

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -34,24 +34,51 @@ class ShenandoahFreeSet : public CHeapObj<mtGC> {
...
+
+  inline void set_mutator_free(size_t idx);
+  inline void set_collector_free(size_t idx);
+  inline void set_old_collector_free(size_t idx);
+
+  inline void clear_mutator_free(size_t idx);
+  inline void clear_collector_free(size_t idx);
+  inline void clear_old_collector_free(size_t idx);
```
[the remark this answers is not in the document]
kdnilsen: I agree with the ideal of improved abstraction, as discussed in other context.  I'm inclined to keep the existing API as is for now, with a todo statement describing possible future refactoring to improve abstraction.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: There are some inconsistencies, I agree.  We do not keep track of capacity for the collector_set, just the old_collector_set and the mutator_set.  This is historical also.  I'm inclined to leave as is for now.  Maybe we can address this when/if we refactor to improve the "set" abstraction.

For now, I'll place a comment in the source code.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
kdnilsen: Added documentation.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -688,67 +918,187 @@ void ShenandoahFreeSet::rebuild() {
...
+                       " collector free range [" SIZE_FORMAT ".." SIZE_FORMAT "], "
+                       "old collector free range [" SIZE_FORMAT ".." SIZE_FORMAT "] allocates from %s",
+                       _mutator_leftmost, _mutator_rightmost, _collector_leftmost, _collector_rightmost,
+                       _old_collector_leftmost, _old_collector_rightmost,
+                       _old_collector_search_left_to_right? "left to right": "right to left");
+    for (uint i = 0; i < _heap->num_regions(); i++) {
+      ShenandoahHeapRegion *r = _heap->get_region(i);
+      uint idx = i % 64;
+      if ((i != 0) && (idx == 0)) {
+        log_info(gc, ergo)(" %6u: %s", i-64, buffer);
```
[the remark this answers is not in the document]
kdnilsen: Thanks.  Done.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: I've tidied this code and added/improved comments to explain its behavior.  I've moved the first loop into its own work method.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: I've added comments to explain rationale for the backward loop.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
kdnilsen: I've added documentation to make this more clear, I hope.

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -309,7 +309,7 @@ class ShenandoahHeap : public CollectedHeap {
     WEAK_ROOTS_BITPOS  = 4,
 
     // Old regions are under marking, still need SATB barriers.
-    OLD_MARKING_BITPOS = 5
+    OLD_MARKING_BITPOS = 5,
```
earthling-amzn, line 312: Are these trailing commas intentional?

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -377,15 +377,17 @@ class ShenandoahHeap : public CollectedHeap {
 
   bool _upgraded_to_full;
 
+  bool _has_evacuation_reserve_quantities;
```
earthling-amzn, line 380: Could we have a description of this new field in the header?

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -42,51 +42,118 @@ ShenandoahFreeSet::ShenandoahFreeSet(ShenandoahHeap* heap, size_t max_regions) :
...
+  assert(!in_mutator_set(idx) && !in_collector_set(idx), "Freeset membership is mutually exclusive");
+  _old_collector_free_bitmap.set_bit(idx);
+}
+
+inline void ShenandoahFreeSet::remove_from_old_collector_set(size_t idx) {
+  _old_collector_free_bitmap.clear_bit(idx);
+}
+
+
+HeapWord* ShenandoahFreeSet::allocate_old_with_affiliation(ShenandoahAffiliation affiliation,
```
earthling-amzn, line 125: Will `affiliation` always be `old` here? Should we remove this parameter?

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -437,29 +523,109 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
...
-    // Touched the bounds? Need to update:
-    if (touches_bounds(num)) {
-      adjust_bounds();
+      assert(probe_mutator_set(idx), "Must be mutator free: " SIZE_FORMAT, idx);
+      remove_from_mutator_set(idx);
+      assert(!in_collector_set(idx) && !in_old_collector_set(idx), "Region cannot be in multiple free sets");
+      adjust_mutator_bounds_if_touched(idx);
+    } else if (r->free() < PLAB::min_size() * HeapWordSize) {
+      // Permanently retire this region if there's room for a fill object
+      size_t waste = r->free();
```
earthling-amzn, line 540: Are we writing a filler object here to keep this region out of the old collector freeset when it's rebuilt?

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -437,29 +523,109 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
...
+      _old_collector_rightmost--;
+    }
+    return true;
+  } else {
+    return false;
+  }
+}
+
+// If idx represents a collector bound, recompute the collector bounds, returning true iff bounds were adjusted.
+bool ShenandoahFreeSet::adjust_collector_bounds_if_touched(size_t idx) {
```
earthling-amzn, line 610: This really feels like there should be a `RegionFreeSet` class with instances for `mutator`, `collector`, and `old_collector`. There's a lot of code duplication here. A refactoring for another PR.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -535,9 +531,9 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
         generation->increase_allocated(waste);
         _heap->notify_mutator_alloc_words(waste >> LogHeapWordSize, true);
       }
-      assert(peek_is_mutator_free(idx), "Must be mutator free: " SIZE_FORMAT, idx);
-      clear_mutator_free(idx);
-      assert(!is_collector_free(idx) && !is_old_collector_free(idx), "Region cannot be in multiple free sets");
+      assert(probe_mutator_set(idx), "Must be mutator free: " SIZE_FORMAT, idx);
+      remove_from_mutator_set(idx);
+      assert(!in_collector_set(idx) && !in_old_collector_set(idx), "Region cannot be in multiple free sets");
       adjust_mutator_bounds_if_touched(idx);
```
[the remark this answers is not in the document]
ysramakrishna: If you are doing the `expand_` and `adjust_` at each insertion or removal from sets, I'd encourage you to think about encapsulating intervals as a performance heuristic into the set abstraction, and always keep them updated, or have an API call that adjusts them if you are batching the interval adjustments to amortize the cost of these adjustments over several insertions and deletions.

ysramakrishna (commented):
I'll leave these comments here for now and finish the rest of the review in a bit.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: yes.  Putting the region back into the old_collector set at rebuild time just makes work for us, because then we'll have to rediscover that it needs to be retired, and will need to retire it, and will need to adjust the bounds again.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: Affiliation might be FREE.  First, we try to allocate from within an old-collector-region that is_old().  If that fails, we'll try to allocate from an old_collector_region that is not afilliated.

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
kdnilsen: my mistake.  I had another constant in this list, and then removed it.

kdnilsen: fixed.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
ysramakrishna: I think the .hpp file declaration documents the intent here:

```
  // Satisfy allocation request req by finding memory that matches affiliation, which either equals req.affiliation
  // or FREE. We know req.is_old().
```

So I think you are right that this interface can probably be simplified and cleaned up (or abstracted and consolidated).

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
kdnilsen: Comment has been added.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: I'm adding a comment to explain this.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -683,82 +930,218 @@ void ShenandoahFreeSet::clear() {
...
+    } else {
+      // We are rebuilding at end of GC, so we set aside budgets specified on command line (or defaults)
+      young_reserve = (_heap->young_generation()->max_capacity() * ShenandoahEvacReserve) / 100;
+      old_reserve = MAX2((_heap->old_generation()->max_capacity() * ShenandoahOldEvacReserve) / 100,
+                         ShenandoahOldCompactionReserve * ShenandoahHeapRegion::region_size_bytes());
+    }
   }
-
+  reserve_regions(young_reserve, old_reserve);
   recompute_bounds();
```
ysramakrishna, line 1016: If the `reserve_regions()` case is the only one where `recompute_bounds()` is called to batch the adjustments following mass movement and reshuffling of free regions into the three types of free sets, then what might work better (cf my previous comment about doing interval end-point adjustments incrementally whenever a set loses or receives a member rather than first add/remove one member and then do an `adjust_bounds_if_touched()`) would be to here have special versions of `add_` and `remove_` which affect the membership change but leave the intervals unchanged (as you do now), but only in the case of these mass adjustments of membership during a rebuild, then do a single sweep to set up the interval extrema at the end of the rebuild.

ysramakrishna (commented):
Another couple of comments.

Will finish review of remaining detailed stuff in the  free set dot cpp in next round.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: I'm adding a comment to the code to clarify.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: Agree that refactoring would be nice in the future.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -55,88 +55,84 @@ inline void ShenandoahFreeSet::increase_used(size_t num_bytes) {
...
 }
 
-inline bool ShenandoahFreeSet::peek_is_mutator_free(size_t idx) const {
+inline bool ShenandoahFreeSet::probe_mutator_set(size_t idx) const {
   return _mutator_free_bitmap.at(idx);
 }
 
-inline bool ShenandoahFreeSet::is_mutator_free(size_t idx) const {
+inline bool ShenandoahFreeSet::in_mutator_set(size_t idx) const {
+  bool is_mutator_free = _mutator_free_bitmap.at(idx);
```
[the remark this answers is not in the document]
kdnilsen: Thanks.  I'm inverting the order of assert and bitmap queries and adding the assert to the probe_ methods.

I agree with future refactoring.  Thanks for the suggestions.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -33,51 +33,102 @@ class ShenandoahFreeSet : public CHeapObj<mtGC> {
...
-  // recycled after tenure age GC passes.
+  // even less frequently than the young survivor regions.
   CHeapBitMap _old_collector_free_bitmap;
   size_t _max;
 
-  // Left-most and right-most region indexes. There are no free regions outside
-  // of [left-most; right-most] index intervals
+  // Left-most and right-most region indexes. There are no free regions outside of [left-most; right-most] index intervals.
+  // The sets are not necessarily contiguous.  It is common for collector_is_free regions to reside within the mutator_is_free
+  // range, and for _old_collector_is_free regions to reside within the collector_is_free range.
```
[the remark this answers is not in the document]
kdnilsen: Yes.  It is possible that all three can overlap.  I'll fix this description.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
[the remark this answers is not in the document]
kdnilsen: I think I'll go ahead and make suggested changes and take a shot at a poor man's set abstraction.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
earthling-amzn: Should we have a more restrictive test when rebuilding the freeset? Rather than including regions with non-zero allocation capacity, include only regions with capacity greater than some minimum? (Minimum LAB size, e.g.)

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: What you suggest might be slightly more efficient, but the code is probably cleaner if we just remove "recompute_bounds()" and compute the bounds incrementally as part of the rebuild() and reserve() process.  This happens only twice per GC cycle (at start of evac, and after update-refs).  In the next revision of this PR, I will offer the "simplified" API.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -33,51 +33,102 @@ class ShenandoahFreeSet : public CHeapObj<mtGC> {
...
+  inline bool in_collector_set(size_t idx) const;
+  inline bool in_old_collector_set(size_t idx) const;
+
+  // The following three probe routines mimic the behavior is in_mutator_set(), in_collector_set() and in_old_collector_set()
+  // but do not assert that the regions have allocation capacity.  These probe routines are used in assertions enforced
+  // during certain state transitions.
+  inline bool probe_mutator_set(size_t idx) const;
+  inline bool probe_collector_set(size_t idx) const;
+  inline bool probe_old_collector_set(size_t idx) const;
+
```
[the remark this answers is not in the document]
kdnilsen: done.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: I've gone partway there, but more thorough refactoring can happen in a future PR.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -535,9 +531,9 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
         generation->increase_allocated(waste);
         _heap->notify_mutator_alloc_words(waste >> LogHeapWordSize, true);
       }
-      assert(peek_is_mutator_free(idx), "Must be mutator free: " SIZE_FORMAT, idx);
-      clear_mutator_free(idx);
-      assert(!is_collector_free(idx) && !is_old_collector_free(idx), "Region cannot be in multiple free sets");
+      assert(probe_mutator_set(idx), "Must be mutator free: " SIZE_FORMAT, idx);
```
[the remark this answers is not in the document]
kdnilsen: There is a comment that describes the difference in the description of the probe_ membership functions.  Is this sufficient?

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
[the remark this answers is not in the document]
kdnilsen: Latest PR reduces replication of code.  It does not go all the way to a new ShenandoahFreeRegionSet abstraction.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: Actually, I did not remove recompute_bounds() because we currently depend on a side effect of this function, which is to decide whether to allocate old_colllector regions left-to-right or right-to-left.  I've placed a comment in the code to clarify a future TODO action.  Is it ok to leave as is for now?  (Or should I try to sort this out right now?)

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -29,8 +29,15 @@
 #include "gc/shenandoah/shenandoahHeapRegionSet.hpp"
 #include "gc/shenandoah/shenandoahHeap.hpp"
 
+enum MemoryReserve {
+  Mutator,
+  Collector,
+  OldCollector
+};
+
```
ysramakrishna, line 37: Did you consider avoiding templates and having a FreeSet class with the relevant methods `in_set(idx)`, `add_to_set(idx)`, etc. that you instantiate three of, a mutator, collector, and old_collector. Then, instead of `in_set<Mutator>(idx)`, etc. you would do `mutator.in_set(idx)`, etc. The downside is that we you can't then do assertion checks about membership of other sets (i.e. mutual exclusion checks) inside the methods that add to a specific set, so would need some wrapper machinery for that outside in ShenandoahFreeSet.

However, the advantage is that such a FreeSet can hide your interval bounds adjustment and checks nicely. Also, that kind of encapsulation allows your `assert_bounds()` to just do `mutator.assert_bounds(); collector.assert_bounds(); ...` etc.

One question was why the addition of a single element doesn't just do the interval adjustment. 

```
   add(x) {
       if (x < leftmost) {
           leftmost = x;
       } else if (x > rightmost) {
           rightmost = x;
       } // else isn't extremal
       set_bit(x);
    }
```

Similarly, the removal of an element also doing the interval adjustment, although more expensive:

```
    remove(x) {
        if (x == leftmost) {
            for (i = leftmost+1; i <= rightmost; i++) {
               if (is_set_bit(i)) {
                  leftmost = i;
                  break;
               }
            }
            assert(leftmost > x, "Error");
         } else if (x == rightmost) {
               for (i = rightmost - 1; i >= leftmost; i--) {
                   if (is_set_bit(i)) {
                      rightmost = i;
                      break;
                   }
               }
               assert(rightmost < x, "Error");
            }
        }
        unset_bit(x);
    }
```

kdnilsen: I've been reluctant to introduce a FreeSet class, because I think a FreeSets class provides better abstraction in that it allows us to encapsulate that as we move regions from one set to another, we want to automate the transfer of capacity between them.

In the most recent commits, I have tried to incorporate some of your suggestions into an implementation of the FreeSets abstraction.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: In the very latest review, I've removed recompute bounds entirely.  The only thing we do at the end of rebuild is decide whether to use right-to-left or left-to-right bias for allocation from OldCollector set.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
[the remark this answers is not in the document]
kdnilsen: I've changed this abstraction.  in_free_set() asserts that the region has memory available to be allocated.  membership() reports which free set a region is in, but does not assert available memory.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: In the very latest revision committed, I have removed the fill object and have taken your suggestion to only install regions into a freeset if available memory is greater than PLAB::min_size().  Thanks.  I think this is cleaner.

kdnilsen:
It occurs to me that this is a "big stretch" for a single PR.  It may be preferable to integrate this in two phases:
1. The first phase could commit up to: https://github.com/openjdk/shenandoah/pull/250/commits/202ee74766e31ebaba3d87770ff40bb40ccbea58
2. The second phase, could integrate the refactoring that followed that stable commit

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -29,30 +29,129 @@
...
+enum MemoryReserve : uint8_t {
+  NotFree,
+  Mutator,
+  Collector,
+  OldCollector,
+  NumFreeSets
+};
+
+class ShenandoahSetsOfFree {
+  friend class ShenandoahFreeSet;
```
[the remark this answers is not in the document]
ysramakrishna: Hmm, wait, the only non-public method is `clear_internal` which is anyway exposed via `clear_all`. 

In that case, is the friend accessing a private data member field? 

Or perhaps the friend annotation is now obsolete from an earlier round of development and must now be removed.

ysramakrishna (commented):
Approach looks good; I left a few initial comments, but will work through the details later.

In my opinion I think the whole thing can be in the same PR, no need to break it up (to your musings somewhere).

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -29,30 +29,129 @@
...
+  inline MemoryReserve membership(size_t idx) const;
 
-  void assert_bounds() const NOT_DEBUG_RETURN;
+  // Returns true iff region idx is in the test_set free_set.  Before returning true, asserts that the free
+  // set is not empty.  Requires that test_set != NotFree or NumFreeSets.
+  inline bool in_free_set(size_t idx, MemoryReserve which_set) const;
 
-  bool is_mutator_free(size_t idx) const;
-  bool is_collector_free(size_t idx) const;
+  // Each of the following four methods returns _max to indicate absence of requested region.
```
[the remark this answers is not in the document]
kdnilsen: Thanks for this question.  The comment is old and wrong.  I'm updating the comment.

Your proposed wording is correct except that the only left_most returns _max whenever which_set is empty.  right_most returns 0 when which_set is empty.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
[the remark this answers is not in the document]
kdnilsen: You are correct that I no longer need the friend class.  In a previous version of the code, I had an assert on ShenandoahFreeSets::clear_all() that was not appropriate for all calls (in particular the initialization call) from ShenandoahFreeSet::clear_internal().  I'll make this change. (hmmm. Or should I put that assert_heaplocked()) back into ShenandoahFreeSets::clear_all()?)

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -29,30 +29,129 @@
 #include "gc/shenandoah/shenandoahHeapRegionSet.hpp"
 #include "gc/shenandoah/shenandoahHeap.hpp"
 
-class ShenandoahFreeSet : public CHeapObj<mtGC> {
+enum MemoryReserve : uint8_t {
```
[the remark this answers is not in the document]
kdnilsen: That's probably better.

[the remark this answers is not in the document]
kdnilsen: My preference for expedience is to keep as is.  But if you feel strongly, I'll separate it out.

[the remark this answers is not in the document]
ysramakrishna: That' fine. Let's leave as is for now, for the reason you state.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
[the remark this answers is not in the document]
ysramakrishna: ok.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -29,30 +29,136 @@
...
+
+  // Return true iff regions for allocation from this set should be peformed left to right.  Otherwise, allocate
+  // from right to left.
+  inline bool alloc_from_left_bias(FreeMemoryType which_set);
+
+  // Determine whether we prefer to allocate from left to right or from right to left for this free-set.
+  void establish_alloc_bias(FreeMemoryType which_set);
+
+  // Assure leftmost, rightmost, leftmost_empty, and rightmost_empty bounds are valid for all free sets.
+  // valid bounds honor all of the following (where max is the number of heap regions):
```
ysramakrishna, line 124: Valid (upper-case V).

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -29,30 +29,136 @@
...
+  // set is not empty.  Requires that test_set != NotFree or NumFreeSets.
+  inline bool in_free_set(size_t idx, FreeMemoryType which_set) const;
+
+  // The following four methods return the left-most and right-most bounds on ranges of regions representing
+  // the requested set.  The _empty variants represent bounds on the range that holds completely empty
+  // regions, which are required for humongous allocations and desired for "very large" allocations.  A
+  // return value of -1 from left_most() or left_most_empty() denotes that the corresponding set is empty.
+  // In other words:
+  //   if the requested which_set is empty:
+  //     left_most() and left_most_empty() return _max, right_most() and right_most_empty() return 0
```
ysramakrishna, line 92: I think you an elide the "_" underscore in left_most (leftmost is an English word). It also makes `leftmost_empty` easier to read.

Would be great if you can make that global substitution in your comments etc. as well before checking in. (Or make it all have the underscore consistently in comments too if you prefer the variant with the underscore.)

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -38,55 +38,410 @@
...
+      ShenandoahHeapRegion* r = heap->get_region(index);
+      available_in_second_half += r->free();
+    }
+  }
+
+  // We desire to first consume the sparsely distributed regions in order that the remaining regions are densely packed.
+  // Densely packing regions reduces the effort to search for a region that has sufficient memory to satisfy a new allocation
+  // request.  Regions become sparsely distributed following a Full GC, which tends to slide all regions to the front of the
+  // heap rather than allowing survivor regions to remain at the high end of the heap where we intend for them to congregate.
+  // In the future, we may modify Full GC so that it slides old objects to the end of the heap and young objects to the start
```
ysramakrishna, line 272: Mark with a `TODO:` ?

ysramakrishna (approved):
I didn't go through all of the details of the allocation strategies to reduce entropy of memory types across the range of regions etc., but this looks good otherwise.

Would you be able to share some statistics or performance numbers that indicate the improvement here, either using range statistics or just raw allocation performance (say using a hyperalloc/extremem workload to compare)?

Thanks for patiently responding to my nitpickings during the review process!

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -29,30 +29,136 @@
 #include "gc/shenandoah/shenandoahHeapRegionSet.hpp"
 #include "gc/shenandoah/shenandoahHeap.hpp"
 
-class ShenandoahFreeSet : public CHeapObj<mtGC> {
+enum FreeMemoryType : uint8_t {
```
earthling-amzn, line 32: Globally visible type should `Shenandoah` prefix.

earthling-amzn (changes requested):

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: Thanks.  Done.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
kdnilsen: Thanks.  Done.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
kdnilsen: Thanks.  Done.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
kdnilsen: Thanks.  Done.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -29,30 +29,129 @@
...
+};
+
+class ShenandoahSetsOfFree {
+  friend class ShenandoahFreeSet;
+
 private:
-  ShenandoahHeap* const _heap;
-  CHeapBitMap _mutator_free_bitmap;
-  CHeapBitMap _collector_free_bitmap;
   size_t _max;
```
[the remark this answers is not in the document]
kdnilsen: Adding a comment: num_regions

ysramakrishna (approved):

earthling-amzn (approved):

PR #253, opened 2023-04-12
Restore event descriptions immortality

shipilev:
Upstream code has event descriptions as immortal strings for a reason: they are passed inside the logging classes, and so the references to them may outlive the scope of the current method. We have seen logs corruption because of that before, e.g. in hs_err-s `Events` section. This PR restores the immortality of event descriptions, producing const strings with macros.

This also stops emitting "(GLOBAL)" for global collections, which I think is what non-generational Shenandoah mode prints. Please tell me if we still need to print "(GLOBAL)" in generational mode.

Tangentially, this also improves performance for compilers that choose to initialize the automatic storage for `char msg[1024]`, see https://isocpp.org/files/papers/P2723R0.html.

Additional testing:
 - [x] Eyeballing the GC logs

---------
### Progress

### Reviewers
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [a550586e](https://git.openjdk.org/shenandoah/pull/253/files/a550586e06296608fedd11e9bbda0ab61fe5aaeb)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/253/head:pull/253` \
`$ git checkout pull/253`

`$ git checkout pull/253` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/253/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 253`

`$ git pr show -t 253`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/253.diff">https://git.openjdk.org/shenandoah/pull/253.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/253#issuecomment-1505048182)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahUtils.hpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahUtils.hpp
@@ -44,18 +44,21 @@
 class GCTimer;
 class ShenandoahGeneration;
 
-#define SHENANDOAH_RETURN_EVENT_MESSAGE(generation_type, prefix, postfix) \
-  switch (generation_type) {                                              \
-    case GLOBAL:                                                          \
-      return prefix " (GLOBAL)" postfix;                                  \
-    case YOUNG:                                                           \
-      return prefix " (YOUNG)" postfix;                                   \
-    case OLD:                                                             \
-      return prefix " (OLD)" postfix;                                     \
-    default:                                                              \
-      ShouldNotReachHere();                                               \
-      return prefix " (?)" postfix;                                       \
-  }                                                                       \
+#define SHENANDOAH_RETURN_EVENT_MESSAGE(heap, generation_type, prefix, postfix) \
+  if (!heap->mode()->is_generational()) {                                       \
+    return prefix "" postfix;                                                   \
+  }                                                                             \
+  switch (generation_type) {                                                    \
+    case GLOBAL:                                                                \
+      return prefix " (GLOBAL)" postfix;                                        \
+    case YOUNG:                                                                 \
+      return prefix " (YOUNG)" postfix;                                         \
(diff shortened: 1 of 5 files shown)


kdnilsen:
We do distinguish GLOBAL from YOUNG and OLD in generational mode.  A GLOBAL collection is when we collect both OLD and YOUNG at the same time.  These are generally rare.  They show up, for example, if someone explicitly invokes System.gc().

src/hotspot/share/gc/shenandoah/shenandoahUtils.hpp
```diff
@@ -44,6 +44,19 @@
 class GCTimer;
 class ShenandoahGeneration;
 
+#define SHENANDOAH_RETURN_EVENT_MESSAGE(generation_type, prefix, postfix) \
+  switch (generation_type) {                                              \
+    case GLOBAL:                                                          \
+      return prefix "" postfix;                                           \
```
kdnilsen, line 50: Suggest reporting GLOBAL here.

kdnilsen (approved):

src/hotspot/share/gc/shenandoah/shenandoahUtils.hpp
shipilev: Ok, now reporting `GLOBAL` for generational mode. This macro is now also checking the generational mode, so we drop any mention of generation in non-generational mode. This keeps the logs the same as upstream for non-generational mode.

ysramakrishna (approved):

shipilev:
Thank you!

/integrate

PR #255, opened 2023-04-13
Account for humongous object waste

earthling-amzn:
The unused portion of the last region in a humongous object is now tracked and logged.

---------
### Progress

### Reviewers
 * [Y. Srinivas Ramakrishna](https://openjdk.org/census#ysr) (@ysramakrishna - Author) ⚠️ Review applies to [e146bf7d](https://git.openjdk.org/shenandoah/pull/255/files/e146bf7d9d47eeece1ca5d06362135e095340601)
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [e146bf7d](https://git.openjdk.org/shenandoah/pull/255/files/e146bf7d9d47eeece1ca5d06362135e095340601)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/255/head:pull/255` \
`$ git checkout pull/255`

`$ git checkout pull/255` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/255/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 255`

`$ git pr show -t 255`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/255.diff">https://git.openjdk.org/shenandoah/pull/255.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/255#issuecomment-1507695526)


shipilev (commented):
Do we need to record this as waste? Can be just "used"?

I suddenly remembered I wanted to fix this a long time ago :)
https://bugs.openjdk.org/browse/JDK-8232057

src/hotspot/share/gc/shenandoah/shenandoahGeneration.hpp
```diff
@@ -52,6 +52,7 @@ class ShenandoahGeneration : public CHeapObj<mtGC> {
 protected:
   // Usage
   size_t _affiliated_region_count;
+  size_t _humongous_waste;      // how much waste for padding in humongous objects
```
ysramakrishna, line 55: Could you document here whether this is included in `_used` below, or is explicitly excluded from it. From my cursory reading of the code doing the adjustments, it looked like `_used` includes `_humongous_waste`.

it would also be good to include a line of documentation about where this statistic might be useful (or, if it is, where it is used today).

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
```diff
@@ -347,14 +347,17 @@ class ShenandoahVerifyOopClosure : public BasicOopIterateClosure {
...
 public:
-  ShenandoahCalculateRegionStatsClosure() : _used(0), _committed(0), _garbage(0), _regions(0) {};
+  ShenandoahCalculateRegionStatsClosure() : _used(0), _committed(0), _garbage(0), _regions(0), _humongous_waste(0) {};
 
   void heap_region_do(ShenandoahHeapRegion* r) {
     _used += r->used();
     _garbage += r->garbage();
     _committed += r->is_committed() ? ShenandoahHeapRegion::region_size_bytes() : 0;
+    if (r->is_humongous()) {
+      _humongous_waste += r->free();
```
ysramakrishna, line 359: Wouldn't this be the case only when `_used > 0` (i.e. the HR is in use and isn't free? Otherwise, is there an implicit assumption of the regions over which this closure is working that wants to appear as an assertion check here?)

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
```diff
@@ -417,6 +421,15 @@ class ShenandoahGenerationStatsClosure : public ShenandoahHeapRegionClosure {
               label, generation->name(), stats.regions(),
               byte_size_in_proper_unit(capacity), proper_unit_for_byte_size(capacity));
 
+    if (!generation->is_global()) {
+      // Humongous waste is currently only tracked for young and old generations
```
ysramakrishna, line 425: Why?

(May be this gets back to my question about where and how this statistic is being used or being planned to be used.)

ysramakrishna (approved):
LGTM. (left a few questions that could help w/clarifying documentation of the "waste" statistic.)

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
earthling-amzn: I'm not sure I follow. This closure visits every region. We don't want to count the `free` portions of regular/in-use regions as humongous waste.

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
earthling-amzn: I will try to change this. Ultimately, I want the "global" generation to subsume the usage metrics currently tracked in `ShenandoahHeap`.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.hpp
earthling-amzn: It isn't included in `used`, but is deducted from `available`. This gives the heuristics a more accurate view of how much memory remains for allocations. I will update the comment.

kdnilsen (approved):
Thanks for cleaning this up.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.hpp
ysramakrishna: Makes sense. Thanks for updating the documentation.

ysramakrishna (approved):
Looks good modulo previous comment re documentation clarification(s). 🚢

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
earthling-amzn: > a fortiori
>: with greater reason or more convincing force —used in drawing a conclusion that is inferred to be even more certain than another

TIL (Today I Learned)

PR #260, opened 2023-04-17
Usage tracking cleanup

earthling-amzn:
There are many nuances to tracing memory utilization. Shenandoah track's usage, waste by humongous objects, padding for promotion LABs alignment and all this is also tracked by generation, the heap and feeds into the heuristics and the pacer. The code to update all of these values and route them to the right places was spread across the allocation call stack. This change consolidates all of the logic into one method, invoked near the end of the allocation.


src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -710,6 +712,54 @@ void ShenandoahHeap::decrease_committed(size_t bytes) {
...
+    // only actual size counts toward usage for mutator allocations
+    generation->increase_used(actual_bytes);
+    increase_used(actual_bytes);
+
+    // notify pacer of both actual size and waste
+    notify_mutator_alloc_words(req.actual_size(), req.waste());
+
+    if (wasted_bytes > 0) {
+      if (req.actual_size() > ShenandoahHeapRegion::humongous_threshold_words()) {
+        // generation->increase_humongous_waste(wasted_bytes);
```
kdnilsen, line 757: why is this commented out?

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -710,6 +712,54 @@ void ShenandoahHeap::decrease_committed(size_t bytes) {
...
+//   free memory left in that region is considered waste. It does not contribute
+//   to the usage, but it _does_ contribute to allocation rate.
+// * The bottom of a PLAB must be aligned on card size. In some cases this will
+//   require padding in front of the PLAB (a filler object). Because this padding
+//   is included in the region's used memory we include the padding in the accounting.
+// * Mutator allocations are used to compute an allocation rate. They are also
+//   sent to the Pacer for those purposes.
+// * There are three sources of waste:
+//  * The padding used to align a PLAB on card size
+//  * Region's free is less than minimum TLAB size and is retired
```
kdnilsen, line 732: I think there are two notions "retirement" in existing code.  There is a temporary retirement that applies only until we rebuild the free set (at end of this gc, or start of next evac).  And there is a permanent retirement that places a fill object into the remnant of free memory so that the memory is no longer free.  (maybe this distinction is not on the master branch.)  The permanent retirement lasts until this region is collected.  The accounting is different.  In both cases, we increase ShenandoahFreeSet::used.  In the permanent case, we also increase the relevant generation:used.

kdnilsen: I'm not sure that we really need both approaches.  The rationale for permanent retirement is that we don't want to repeatedly discover and re-retire old-gen regions that are scattered throughout the heap.  This is less of a problem with young-gen regions, because regions that are retired once will either soon be promoted or they will be collected.

kdnilsen (approved):
This looks like a good improvement.  Thanks for sorting through these details.  I assume we pass all the existing GenShen regression tests.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
earthling-amzn: I'm not familiar with the 'permanent' retirement - I don't think that is on `master`. We do fill in LABs with objects when they are retired, but that memory was already 'used' as soon as the LAB was allocated. These changes don't actually touch how the freeset tracks its own usage. What's been hoisted out is only the accounting for generation/heap usage, allocation rate and what the pacer uses. If the 'permanent' retirement includes the filler object in the `actual_size` of the allocation then this new function will account for it as expected. It _should_ do this so that sum of region::used equals the used for a generation/heap.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
earthling-amzn: I will uncomment it after humongous waste tracking is in.

earthling-amzn:
There are no new regressions here.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -246,13 +246,45 @@ HeapWord* ShenandoahFreeSet::allocate_single(ShenandoahAllocRequest& req, bool&
...
+      usable_free -= CardTable::card_size();
+    } else {
+      assert(usable_free == 0, "usable_free is a multiple of card_size and card_size > min_fill_size");
+    }
+  }
+
+  return usable_free / HeapWordSize;;
+}
+
+HeapWord* ShenandoahFreeSet::allocated_aligned_plab(size_t size, ShenandoahAllocRequest& req, ShenandoahHeapRegion* r) {
```
ysramakrishna, line 266: Let's write a documentation comment:
```
// Given a size argument, which is a multiple of card size, a request struct
// for a PLAB, and an old region, return a pointer to the allocated space for
// a PLAB which is card-aligned and where any remaining shard in the region
// has been suitably filled by a filler object.
// It is assumed (and assertion-checked) that such an allocation is always possible.
```

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -246,13 +246,45 @@ HeapWord* ShenandoahFreeSet::allocate_single(ShenandoahAllocRequest& req, bool&
...
+  assert(_heap->mode()->is_generational(), "PLABs are only for generational mode");
+  assert(r->is_old(), "All PLABs reside in old-gen");
+  assert(!req.is_mutator_alloc(), "PLABs should not be allocated by mutators.");
+  assert(size % CardTable::card_size_in_words() == 0, "size must be multiple of card table size, was " SIZE_FORMAT, size);
+
+  HeapWord* result = r->allocate_aligned(size, req, CardTable::card_size());
+  assert(result != nullptr, "Allocation cannot fail");
+  assert(r->top() <= r->end(), "Allocation cannot span end of region");
+  assert(req.actual_size() % CardTable::card_size_in_words() == 0, "PLAB size must be multiple of card size");
```
ysramakrishna, line 275: We've already checked at line 270 that `size % CardTable::card_size_in_words() == 0`. I think we can check a stronger condition here, i.e.:
```
assert(req.actual_size() == size, "Unexpected actual_size()");
```

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -246,13 +246,45 @@ HeapWord* ShenandoahFreeSet::allocate_single(ShenandoahAllocRequest& req, bool&
...
+  return usable_free / HeapWordSize;;
+}
+
+HeapWord* ShenandoahFreeSet::allocated_aligned_plab(size_t size, ShenandoahAllocRequest& req, ShenandoahHeapRegion* r) {
+  assert(_heap->mode()->is_generational(), "PLABs are only for generational mode");
+  assert(r->is_old(), "All PLABs reside in old-gen");
+  assert(!req.is_mutator_alloc(), "PLABs should not be allocated by mutators.");
+  assert(size % CardTable::card_size_in_words() == 0, "size must be multiple of card table size, was " SIZE_FORMAT, size);
+
+  HeapWord* result = r->allocate_aligned(size, req, CardTable::card_size());
```
ysramakrishna, line 272: Note that `allocate_aligned` takes a bunch of safety measures that are, in this particular caller's case, redundant ad wasted. Since allocation paths must be made as efficient as possible, we should (in the future) try and get these to be more efficient and avoid such redundancy where possible.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -288,112 +319,46 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
...
-          _heap->old_generation()->increase_used(padding);
-          // For verification consistency, we need to report this padding to _heap
-          _heap->increase_used(padding);
-        }
+        result = allocated_aligned_plab(size, req, r);
       }
       // Otherwise, leave result == nullptr because the adjusted size is smaller than min size.
     } else {
       // This is a GCLAB or a TLAB allocation
+      size_t size = req.size();
```
ysramakrishna, line 338: I prefer `adjusted_size` here too, but ok if you want to leave it this way (there are adjustments that happen below).

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -288,112 +319,46 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
...
       size_t free = align_down(r->free() >> LogHeapWordSize, MinObjAlignment);
```
ysramakrishna, line 339: Not code that you wrote, but still ...

Why do we need this alignment downward? It sounds like this would want to be an assertion instead:
```
size_t free_words = r->free() >> LogHeapWordSize;
assert(free_words == align_down(free_words, MinObjAlignment), "Should always be min obj aligned");
```

Or may be I am missing something here. In which case a suitable documentation comment would be useful.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
ysramakrishna, line 266: Did you really want the method to be called `allocated_...` or `allocate_...`.

I'd think the imperative verb form works better here, and I assume the extra `d` was an unintended typo?

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -288,112 +319,46 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
...
-        if (free > usable_free) {
-          // Account for the alignment padding
-          size_t padding = (free - usable_free) * HeapWordSize;
-          increase_used(padding);
-          assert(r->is_old(), "All PLABs reside in old-gen");
-          _heap->old_generation()->increase_used(padding);
-          // For verification consistency, we need to report this padding to _heap
-          _heap->increase_used(padding);
-        }
+        result = allocated_aligned_plab(size, req, r);
```
ysramakrishna, line 333: `allocate_..`

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -288,112 +319,46 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
...
-      if (free > usable_free) {
-        // Account for the alignment padding
-        size_t padding = (free - usable_free) * HeapWordSize;
-        increase_used(padding);
-        assert(r->is_old(), "All PLABs reside in old-gen");
-        _heap->old_generation()->increase_used(padding);
-        // For verification consistency, we need to report this padding to _heap
-        _heap->increase_used(padding);
-      }
+      result = allocated_aligned_plab(size, req, r);
```
ysramakrishna, line 358: `allocate_..`

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -288,112 +319,46 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
...
-        }
-      }
-      free /= HeapWordSize;
-      usable_free /= HeapWordSize;
-      size_t remnant = size % CardTable::card_size_in_words();
-      if (remnant > 0) {
-        // Since we have Elastic TLABs, align size up.  This is consistent with aligning min_size up.
-        size = size - remnant + CardTable::card_size_in_words();
-      }
+      // Since we have Elastic TLABs, align size up.  This is consistent with aligning min_size up.
```
ysramakrishna, line 323: I's just say:
// Since we are using ElasticTLABs, we align the allocation up to card sizes/boundaries.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
ysramakrishna: Isn't this likely to fail -- at least in the general case -- the alignment checks in assertions in `allocate_aligned_plab()`?

I am not sure I understand the difference between `ShenandoahElasticPLAB` and not, since the documentation comment isn't very helpful:
```
  product(bool, ShenandoahElasticTLAB, true, DIAGNOSTIC,                    \
          "Use Elastic TLABs with Shenandoah")                              \
```

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -405,9 +370,7 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
   if (result != nullptr) {
     // Allocation successful, bump stats:
     if (req.is_mutator_alloc()) {
-      assert(req.is_young(), "Mutator allocations always come from young generation.");
-      generation->increase_used(size * HeapWordSize);
-      increase_used(size * HeapWordSize);
+      increase_used(req.actual_size() * HeapWordSize);
```
ysramakrishna, line 373: There is leakage of abstraction happening here. I'd prefer if the adjustments of heap use occurred when the space was allocated by the region. That avoids accounting errors because of piecemeal adjustments scattered along the allocation paths here.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -405,9 +370,7 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
   if (result != nullptr) {
     // Allocation successful, bump stats:
     if (req.is_mutator_alloc()) {
-      assert(req.is_young(), "Mutator allocations always come from young generation.");
-      generation->increase_used(size * HeapWordSize);
```
ysramakrishna, line 409: Where is the generation's use being adjusted now?

ysramakrishna (changes requested):
I have left a few comments. Despite the improvements that you have made by some of the current refactoring, I am concerned about the organization of the allocation paths and believe that there are opportunities for cleaning these up, and specializing and slimming these allocation paths a bit more.

That can wait for later, but I'd like to take another pass at this once you have had a chance to read the review comments and in case those lead to any changes.

Thanks!

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -224,6 +224,7 @@ class ShenandoahHeap : public CollectedHeap {
   shenandoah_padding(1);
 
 public:
+  void increase_used(const ShenandoahAllocRequest& req);
   void increase_used(size_t bytes);
```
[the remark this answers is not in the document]
earthling-amzn: I would like to remove `increase_used(size_t bytes)` from `ShenandoahHeap` and use `ShenandoahGlobalGeneration` as the source of truth for "whole heap" metrics.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -710,6 +712,54 @@ void ShenandoahHeap::decrease_committed(size_t bytes) {
   _committed -= bytes;
 }
 
+// For tracking usage based on allocations, it should be the case that:
+// * The sum of regions::used == heap::used
+// * The sum of a generation's regions::used == generation::used
+// * The sum of a generation's humongous regions::free == generation::humongous_waste
+// These invariants are checked by the verifier on safepoints.
```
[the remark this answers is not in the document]
earthling-amzn: Sure, will do.

ysramakrishna:
> I have left a few comments. Despite the improvements that you have made by some of the current refactoring, I am concerned about the organization of the allocation paths and believe that there are opportunities for cleaning these up, and specializing and slimming these allocation paths a bit more.
> 
> That can wait for later, but I'd like to take another pass at this once you have had a chance to read the review comments and in case those lead to any changes.
> 
> Thanks!

This is a good clean up, now that I see the big picture of the consolidation of the adjustments. I'll approve once you have had a chance to consider my review comments.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
earthling-amzn: When `ShenandoahElasticTLAB` is enabled, the allocator is allowed to return a TLAB smaller than the requested size (but greater than the `min_size`).

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
earthling-amzn: Oops, good catch.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
earthling-amzn: Right - your withdrawn comment is just what I was going for with this change. `ShenandoahFreeSet` maintains its own notion of `used`, which has similar but subtly different rules for the accounting that happens for generations and the heap. Perhaps they could be merged at some point.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
earthling-amzn: Waste, generation/heap used, bytes allocated and the concerns for the pacer are all consolidated into the overloaded `increase_used` method in `ShenandoahHeap`.

earthling-amzn:
Thanks for the review @ysramakrishna . I'll make changes based on your suggestions.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -710,6 +712,54 @@ void ShenandoahHeap::decrease_committed(size_t bytes) {
...
+//
+// Additional notes:
+// * When a mutator's allocation request causes a region to be retired, the
+//   free memory left in that region is considered waste. It does not contribute
+//   to the usage, but it _does_ contribute to allocation rate.
+// * The bottom of a PLAB must be aligned on card size. In some cases this will
+//   require padding in front of the PLAB (a filler object). Because this padding
+//   is included in the region's used memory we include the padding in the accounting.
```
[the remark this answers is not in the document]
earthling-amzn: We don't really get waste in the strictest sense for non-humongous objects. However, the _bottom_ of PLABs need to be aligned to card size (and this alignment itself needs to be large enough to hold a "filler" object). When this happens, this filler object is included in the "used" memory for the region (and so, is included in the "used" memory for the old generation). I will clarify this comment.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -723,17 +773,10 @@ void ShenandoahHeap::decrease_used(size_t bytes) {
...
-    increase_used(bytes);
-  }
-
+void ShenandoahHeap::notify_mutator_alloc_words(size_t words, size_t waste) {
   if (ShenandoahPacing) {
     control_thread()->pacing_notify_alloc(words);
-    if (waste) {
-      pacer()->claim_for_alloc(words, true);
-    }
+    pacer()->claim_for_alloc(waste, true);
```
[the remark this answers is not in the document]
earthling-amzn: I don't think there was a bug in the original code, I also think this change preserves the original semantics. After this change, the pacer is being informed of the allocation _and_ the waste in the same method call. Previously, `waste` was a boolean and this function would be called twice per allocation - once with `waste` equal to `false` (for the `actual_size`) and again with `waste` equal to `true`. Now, `waste` is just the amount of waste (`size_t`) and this method is called with both `actual_size` and `waste` at the same time.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -246,13 +246,45 @@ HeapWord* ShenandoahFreeSet::allocate_single(ShenandoahAllocRequest& req, bool&
   return nullptr;
 }
 
+size_t get_usable_free_words(size_t free) {
```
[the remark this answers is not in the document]
earthling-amzn: Yes, I agree. I would like something like a strong typedef here to prevent mixing bytes and words in the same expression. If you look closely at the [original code](https://github.com/openjdk/shenandoah/pull/260/files#diff-46de849c458b5bfca09d1c8134e9d71d131faac923b39d44db2fe27d168941aeL370), the path for in-elastic TLABs mixed up words and bytes.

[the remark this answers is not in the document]
earthling-amzn: I wrote it this way to reduce duplication in the callers.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
earthling-amzn: Hmm... I'm not sure about that. In the case when elastic TLABs is enabled, we will have already decreased size to fit in the region and so we don't expect to `allocate_aligned` to further decrease the size. However, in the case when elastic TLABs is disabled we can't be sure that `allocate_aligned` will not set `actual_size` to something smaller than the requested `size`. This arguably violates the intention behind disabling elastic TLABs, but I think this is a matter for a different PR?

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
earthling-amzn: Some of these checks are still required when ElasticTLABs is disabled.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
earthling-amzn: Never mind, I see that when elastic TLABs are disabled and we only `allocate_aligned` if we know the request will fit.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
earthling-amzn: Is this just making sure there is enough free space for a minimum sized object?

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -710,6 +712,54 @@ void ShenandoahHeap::decrease_committed(size_t bytes) {
...
+//  * The unused portion of memory in the last region of a humongous object
+void ShenandoahHeap::increase_used(const ShenandoahAllocRequest& req) {
+  size_t actual_bytes = req.actual_size() * HeapWordSize;
+  size_t wasted_bytes = req.waste() * HeapWordSize;
+  ShenandoahGeneration* generation = generation_for(req.affiliation());
+
+  if (req.is_gc_alloc()) {
+    assert(wasted_bytes == 0 || req.type() == ShenandoahAllocRequest::_alloc_plab, "Only PLABs have waste");
+    generation->increase_used(actual_bytes + wasted_bytes);
+    increase_used(actual_bytes + wasted_bytes);
```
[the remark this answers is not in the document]
earthling-amzn: Yes, but waste is only non-zero for PLABs because only PLABs have their "waste" _in front of_ the usable parts of the LAB.

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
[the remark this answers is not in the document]
earthling-amzn: I've updated the pull request to track whole heap usage in the global generation.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
ysramakrishna: I realize now that if the user runs with a larger than needed value for "MinObjAlignment" by setting it on the command-line, this will respect that request. This code is fine. Apologies for my confusion. I had somehow assumed that if MinObjAignment is set then the heap (regions) always maintain the invariant that object space is allocated in chunks that are always aligned to that request. Perhaps that is not true in general. It's been a while since I have been in this code (although I am sure @rkennke has been recently on account of Lilliput!)

ysramakrishna (approved):
Thanks for your patience with my review comments! 

This looks good to me. 🚢

src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
```diff
@@ -1350,9 +1350,20 @@ class ShenandoahPostCompactClosure : public ShenandoahHeapRegionClosure {
...
+      _heap->old_generation()->establish_usage(_old_regions, _old_usage, _old_humongous_waste);
+      _heap->young_generation()->establish_usage(_young_regions, _young_usage, _young_humongous_waste);
+    } else {
+      assert(_old_regions == 0, "Old regions only expected in generational mode");
+      assert(_old_usage == 0, "Old usage only expected in generational mode");
+      assert(_old_humongous_waste == 0, "Old humongous waste only expected in generational mode");
+    }
+
+    // In generational mode, global usage should be the sum of young and old. This is also true
+    // for non-generational modes except that there are no old regions.
```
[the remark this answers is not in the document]
earthling-amzn: Yes - the assertions on 1357-1359 assert that these values are all zero for old generation in the non-generational modes.

PR #268, opened 2023-04-27
Cleanup NumberSeq additions

shipilev:
I have been looking into how to decrease our upstream exposure for `NumberSeq` changes, and it seems to be not solvable in an easy way. Normally, I would add `AbsSeq::add(AbsSeq& other)` methods and get to that from `HdrSeq::add(HdrSeq& other)`, and that almost works. The remaining bit is figuring out the decaying average/variance thingie. 

I spent an hour trying to come up with easily provable equations for _combining_ the decaying average/variance. The decaying average is somewhat easy, although there are boundary problems (decaying average treats 0-th element without adjustments, which requires remembering the `_last` element to compensate). Decaying variance looks remarkably scary in the few of my approaches.

If we enter with current code upstream, it would be the uphill battle to implemented decays and prove they work. So instead, I took the alternative approach of doing everything inside Shenandoah code, except the decays. This reverts `numberSeq.*` to upstream state.

Additional testing:
 - [x] Linux x86_64 fastdebug `hotspot_gc_shenandoah`
 - [x] Linux AArch64 fastdebug `hotspot_gc_shenandoah`
 - [x] GitFarm

---------
### Progress

### Reviewers
 * [Roman Kennke](https://openjdk.org/census#rkennke) (@rkennke - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/268/head:pull/268` \
`$ git checkout pull/268`

`$ git checkout pull/268` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/268/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 268`

`$ git pr show -t 268`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/268.diff">https://git.openjdk.org/shenandoah/pull/268.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/268#issuecomment-1525945644)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahNumberSeq.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahNumberSeq.cpp
@@ -149,8 +149,10 @@ void HdrSeq::add(HdrSeq& other) {
 void HdrSeq::clear() {
   for (int mag = 0; mag < MagBuckets; mag++) {
     int* bucket = _hdr[mag];
-    for (int c = 0; c < ValBuckets; c++) {
-      bucket[c] = 0;
+    if (bucket != nullptr) {
+      for (int c = 0; c < ValBuckets; c++) {
+        bucket[c] = 0;
+      }
     }
   }
 }
--- a/src/hotspot/share/gc/shenandoah/shenandoahNumberSeq.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahNumberSeq.cpp
@@ -121,7 +121,7 @@ double HdrSeq::percentile(double level) const {
   return maximum();
 }
 
-void HdrSeq::add(HdrSeq& other) {
+void HdrSeq::add(const HdrSeq& other) {
   if (other.num() == 0) {
     // Other sequence is empty, return
     return;
--- a/src/hotspot/share/gc/shenandoah/shenandoahNumberSeq.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahNumberSeq.cpp
@@ -165,6 +165,7 @@ void HdrSeq::add(HdrSeq& other) {
 }
 
 void HdrSeq::clear() {
+  // Clear the storage
   for (int mag = 0; mag < MagBuckets; mag++) {
     int* bucket = _hdr[mag];
     if (bucket != nullptr) {
@@ -173,6 +174,15 @@ void HdrSeq::clear() {
       }
     }
   }
+
+  // Clear other fields too
+  _last = 0;
+  _maximum = 0;
+  _sum = 0;
+  _sum_of_squares = 0;
+  _num = 0;
+  _davg = 0;
+  _dvariance = 0;
 }
 
 BinaryMagnitudeSeq::BinaryMagnitudeSeq() {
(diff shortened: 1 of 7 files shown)


src/hotspot/share/gc/shenandoah/shenandoahNumberSeq.cpp
```diff
@@ -121,40 +121,58 @@ double HdrSeq::percentile(double level) const {
...
+  _sum += other._sum;
+  _sum_of_squares += other._sum_of_squares;
+  _num += other._num;
+
+  // Until JDK-8298902 is fixed, we taint the decaying statistics
+  _davg = NAN;
+  _dvariance = NAN;
+}
+
+void HdrSeq::clear() {
```
ysramakrishna, line 167: Don't the _num, _sum, _maximum, etc. need to be cleared as well? May be extend the tests to check for clearing if not already the case.

Otherwise looks good; will take a closer look in a bit.

shipilev: Ah yes, of course. Cleared now.

src/hotspot/share/gc/shenandoah/shenandoahNumberSeq.cpp
```diff
@@ -121,40 +121,68 @@ double HdrSeq::percentile(double level) const {
   return maximum();
 }
 
-// Merge this HdrSeq into hdr2: clear optional and on-by-default
-// Note: this method isn't intrinsically MT-safe; callers must take care
-// of any mutual exclusion as necessary.
-void HdrSeq::merge(HdrSeq& hdr2, bool clear_this) {
+void HdrSeq::add(HdrSeq& other) {
```
earthling-amzn, line 124: Can we make `other` be a `const` reference? just to make it clear who is getting added to who.

earthling-amzn (changes requested):

ysramakrishna (approved):
LGTM, modulo @earthling-amzn 's suggestion and ideally extending the test to cover clear.

src/hotspot/share/gc/shenandoah/shenandoahNumberSeq.cpp
shipilev: Yes, added!

ysramakrishna (approved):
LGTM. 🚢

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
```diff
@@ -966,7 +966,8 @@ template<typename RememberedSet>
 void ShenandoahScanRemembered<RememberedSet>::merge_worker_card_stats_cumulative(
   HdrSeq* worker_stats, HdrSeq* cum_stats) {
   for (int i = 0; i < MAX_CARD_STAT_TYPE; i++) {
-    worker_stats[i].merge(cum_stats[i]);
+    cum_stats[i].add(worker_stats[i]);
```
rkennke, line 969: While you are here, maybe rename the cum_stats variable to something else ;-)

rkennke (approved):
Looks good to me. Maybe rename that variable. ;-)

rkennke (approved):

ysramakrishna (approved):

PR #270, opened 2023-05-02
Use distinct "end of cycle" message for each Shenandoah pause

earthling-amzn:
This change allows the `GarbageCollectionNotificationInfo::gcAction` to be overridden. Shenandoah now uses this to distinguish the notification raised for each of the pauses during the concurrent cycle. This will make it easier for monitoring software to identify which phases may be having longer than expected pauses.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/270/head:pull/270` \
`$ git checkout pull/270`

`$ git checkout pull/270` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/270/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 270`

`$ git pr show -t 270`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/270.diff">https://git.openjdk.org/shenandoah/pull/270.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/270#issuecomment-1532258497)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahVMOperations.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahVMOperations.cpp
@@ -49,55 +49,36 @@ void VM_ShenandoahReferenceOperation::doit_epilogue() {
 }
 
 void VM_ShenandoahInitMark::doit() {
-  ShenandoahGCPauseMark mark(_gc_id, SvcGCMarker::CONCURRENT);
+  ShenandoahGCPauseMark mark(_gc_id, "Init Mark", SvcGCMarker::CONCURRENT);
   _gc->entry_init_mark();
 }
 
 void VM_ShenandoahFinalMarkStartEvac::doit() {
-  ShenandoahGCPauseMark mark(_gc_id, SvcGCMarker::CONCURRENT);
+  ShenandoahGCPauseMark mark(_gc_id, "Final Mark", SvcGCMarker::CONCURRENT);
   _gc->entry_final_mark();
 }
 
 void VM_ShenandoahFullGC::doit() {
-  ShenandoahGCPauseMark mark(_gc_id, SvcGCMarker::FULL);
+  ShenandoahGCPauseMark mark(_gc_id, "Full GC", SvcGCMarker::FULL);
   _full_gc->entry_full(_gc_cause);
 }
 
 void VM_ShenandoahDegeneratedGC::doit() {
-  ShenandoahGCPauseMark mark(_gc_id, SvcGCMarker::CONCURRENT);
+  ShenandoahGCPauseMark mark(_gc_id, "Degenerated GC", SvcGCMarker::CONCURRENT);
   _gc->entry_degenerated();
 }
 
 void VM_ShenandoahInitUpdateRefs::doit() {
-  ShenandoahGCPauseMark mark(_gc_id, SvcGCMarker::CONCURRENT);
+  ShenandoahGCPauseMark mark(_gc_id, "Init Update Refs", SvcGCMarker::CONCURRENT);
   _gc->entry_init_updaterefs();
 }
 
 void VM_ShenandoahFinalUpdateRefs::doit() {
-  ShenandoahGCPauseMark mark(_gc_id, SvcGCMarker::CONCURRENT);
+  ShenandoahGCPauseMark mark(_gc_id, "Final Update Refs", SvcGCMarker::CONCURRENT);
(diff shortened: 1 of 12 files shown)


src/hotspot/share/gc/shenandoah/shenandoahVMOperations.cpp
```diff
@@ -49,55 +49,36 @@ void VM_ShenandoahReferenceOperation::doit_epilogue() {
...
 }
 
 void VM_ShenandoahFinalUpdateRefs::doit() {
-  ShenandoahGCPauseMark mark(_gc_id, SvcGCMarker::CONCURRENT);
+  ShenandoahGCPauseMark mark(_gc_id, "Final Update Refs", SvcGCMarker::CONCURRENT);
   _gc->entry_final_updaterefs();
 }
 
 void VM_ShenandoahFinalRoots::doit() {
-  ShenandoahGCPauseMark mark(_gc_id, SvcGCMarker::CONCURRENT);
```
kdnilsen, line 82: Why are we disabling the aging of regions?  And why are we disabling the aging cycle?

kdnilsen: In my initial review, I missed that this functionality has been moved to ShenandoahConcurrentGC::entry_final_roots()

earthling-amzn: I moved it into `ShenandoahConcurrentGC::op_final_roots` to fit the pattern of other phases (`VM_Operation_XYZ` and `entry_XYZ` should just have GC tracking boiler plate). Also, with this approach, we don't have to marshal the `is_aging_cycle` around.

kdnilsen: Thanks for clarifying.

earthling-amzn: As I think on this a bit more... we should probably also age regions during final update refs. The "final roots" phase only happens on abbreviated cycles.

kdnilsen (approved):

src/hotspot/share/gc/shenandoah/shenandoahVMOperations.cpp
kdnilsen: Thanks for catching that.  We need it in both places or some code path shared between the different cycles.

src/hotspot/share/gc/shenandoah/shenandoahVMOperations.cpp
```diff
@@ -49,55 +49,36 @@ void VM_ShenandoahReferenceOperation::doit_epilogue() {
 }
 
 void VM_ShenandoahInitMark::doit() {
-  ShenandoahGCPauseMark mark(_gc_id, SvcGCMarker::CONCURRENT);
+  ShenandoahGCPauseMark mark(_gc_id, "Init Mark", SvcGCMarker::CONCURRENT);
```
ysramakrishna, line 52: Not your code, but I am curious how the 3rd argument "reason_type" `{MINOR, FULL, CONCURRENT}` is even ever used. I saw it dutifully passed around and eventually get used only to check whether it was `== FULL` or not, down in `VM_GC_Operation::notify_gc_begin`.

src/hotspot/share/services/memoryManager.cpp
```diff
@@ -293,7 +293,8 @@ void GCMemoryManager::gc_end(bool recordPostGCUsage,
     }
 
     if (is_notification_enabled()) {
-      GCNotifier::pushNotification(this, _gc_end_message, GCCause::to_string(cause));
+      const char* message = notificationMessage != nullptr ? notificationMessage : _gc_end_message;
+      GCNotifier::pushNotification(this, message, GCCause::to_string(cause));
```
ysramakrishna, line 297: This looks great to me. I am assuming all collectors that do their work in multiple pieces (at least ZGC and possibly G1 too, may be?) might want to make use of this functionality.

I am guessing this might benefit from an upstream review at some point.

ysramakrishna (approved):
LGTM.

src/hotspot/share/services/memoryManager.cpp
earthling-amzn: Yes, I reckon so. I will target these changes for upstream as well.

PR #275, opened 2023-05-09
Increase card offset mask

earthling-amzn:
On 32-bit platforms, a 512 byte card covers 128 words. On 64-bit platforms, a 1024 byte card (the maximum allowed) also covers 128 words. This change increases the size of the mask we use when encoding object start offsets to handle these situations.

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/275/head:pull/275` \
`$ git checkout pull/275`

`$ git checkout pull/275` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/275/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 275`

`$ git pr show -t 275`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/275.diff">https://git.openjdk.org/shenandoah/pull/275.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/275#issuecomment-1540538836)


kdnilsen (approved):

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
```diff
@@ -391,13 +391,14 @@ class ShenandoahCardCluster: public CHeapObj<mtGC> {
   // a particular card region.  We pack this bit into start byte under assumption that start byte is accessed less
   // frequently than last byte.  This is true when number of clean cards is greater than number of dirty cards.
   static const uint16_t ObjectStartsInCardRegion = 0x80;
-  static const uint16_t FirstStartBits           = 0x3f;
```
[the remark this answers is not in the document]
earthling-amzn: Hmm, we are definitely using the `last` field in the `cross_map` struct. I'd have to study the code to understand how we could remove it.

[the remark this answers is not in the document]
kdnilsen: You are right that we are still "maintaining" the last field and there are many references to it in the code.  I am thinking that with Ramki's changes to remembered set scanning, it is no longer used in scanning, but I may be remembering this incorrectly.

I agree this is not a subtle change and would require some "study" to make sure it is done properly, if at all.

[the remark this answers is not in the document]
ysramakrishna: @kdnilsen : `object_starts` uses the xmap info, which is still used in card scanning in the `block_start` method. The use of a traditional BOT was deferred to future work when I redid the RS scanning earlier this year. There's an open ticket for that work/investigation.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.hpp
```diff
@@ -391,13 +391,14 @@ class ShenandoahCardCluster: public CHeapObj<mtGC> {
...
+  static const uint16_t FirstStartBits           = 0x7f;
 
   crossing_info *object_starts;
 
 public:
   // If we're setting first_start, assume the card has an object.
   inline void set_first_start(size_t card_index, uint8_t value) {
+    assert(value < FirstStartBits, "Offset into card would be truncated");
```
ysramakrishna, line 401: `value <= FirstStartBits` ?

Can/should replace (or augment, if you prefer) with a static assert using the maximum offset computed using card size and oop size?

ysramakrishna: (dynamic) may be also in `set_last_start`?

earthling-amzn: I like the idea of a static assert, will look into that.

earthling-amzn: Shouldn't it be `<`? If the offset were `=` to `FirstStartBits` it should be offset `0` in the next card?

PR #278, opened 2023-05-10
Assert that region usage accounting is only accessed in generational mode

earthling-amzn:
There was one unrelated test failure in the github actions (running G1).

---------
### Progress

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/278/head:pull/278` \
`$ git checkout pull/278`

`$ git checkout pull/278` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/278/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 278`

`$ git pr show -t 278`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/278.diff">https://git.openjdk.org/shenandoah/pull/278.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/278#issuecomment-1542786597)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahGlobalGeneration.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahGlobalGeneration.cpp
@@ -30,7 +30,6 @@
 #include "gc/shenandoah/shenandoahUtils.hpp"
 #include "gc/shenandoah/shenandoahVerifier.hpp"
 #include "gc/shenandoah/shenandoahYoungGeneration.hpp"
-#include "gc/shenandoah/heuristics/shenandoahHeuristics.hpp"
 
 const char* ShenandoahGlobalGeneration::name() const {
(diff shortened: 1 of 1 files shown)


kdnilsen (approved):

ysramakrishna:
Is the idea of the assert then to catch the failure to debug it? Should the `global generation` instance not even exist (i.e. c'tor not be called) in non-generational mode, or can such an object exist, but this and only this method may not be called on it in non-generational mode? Wondering if this is just a question of running the failing GHA test to reproduce the issue?

earthling-amzn:
Yes, it's the latter scenario. The global generation exists in all modes, but the implementation of `used_regions` is only correct for the generational mode. The assert is intended to show that this specific method is only called in generational mode.

PR #286, opened 2023-06-14
8309874: NMethod barriers may remain armed when regions are promoted in place

earthling-amzn:
Genshen runs in-place promotions during the evacuation phase. If the cycle is _only_ performing in-place promotions, it should not arm NMethod barriers. The barriers are now only armed and disarmed if the collection set is not empty.

---------
### Progress

### Issue
 * [JDK-8309874](https://bugs.openjdk.org/browse/JDK-8309874): NMethod barriers may remain armed when regions are promoted in place (**Bug** - P4)

### Reviewers
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [fac642a3](https://git.openjdk.org/shenandoah/pull/286/files/fac642a3a3c72c55eecab7d17ffe820016b1d5d5)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/286/head:pull/286` \
`$ git checkout pull/286`

`$ git checkout pull/286` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/286/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 286`

`$ git pr show -t 286`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/286.diff">https://git.openjdk.org/shenandoah/pull/286.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/286#issuecomment-1591945172)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
@@ -766,7 +766,6 @@ void ShenandoahConcurrentGC::op_final_mark() {
     heap->prepare_concurrent_roots();
 
     if (heap->mode()->is_generational()) {
-      ShenandoahGeneration* young_gen = heap->young_generation();
       size_t humongous_regions_promoted = heap->get_promotable_humongous_regions();
       size_t regular_regions_promoted_in_place = heap->get_regular_regions_promoted_in_place();
       if (!heap->collection_set()->is_empty() || (humongous_regions_promoted + regular_regions_promoted_in_place > 0)) {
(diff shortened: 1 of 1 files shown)


kdnilsen (approved):
Thanks for fixing this.

ysramakrishna:
Is this a performance improvement, or does not fixing this lead to buggy behaviour?

It sounded like in the case where the collection set isn't empty the nmethod barriers are disarmed as a consequence of some actions that would be skipped when the collection set is empty, thus leaving the nmethod barriers armed past that point? Could you point out the paths where the disarming happens in that case, and add it as a comment perhaps in the `arm` and `disarm` methods, so that this protocol is clearly documented?

Thanks!

earthling-amzn:
This addresses an assert on our 17 branches, but I suspect it is a performance improvement as well.

ysramakrishna (approved):
Thanks for adding the comment!

ysramakrishna:
> This addresses an assert on our 17 branches (the assert was removed along with the nmethod sweeper), but I suspect it fixes a performance regression as well.

Is there a natural place for the assert? (Can it be asserted in an appropriate heap verification perhaps?) Doesn't need to be done here, but wondering about how to make this future-proof via an assertion/check.

PR #289, opened 2023-06-15
8311883: [Genshen] Adaptive tenuring threshold

ysramakrishna:
JDK-8311883 [GenShen] Adaptive tenuring

I am opening this previously draft PR for formal preliminary review. It has already benefited from review feedback from a code walkthrough of an earlier version of the code. Most of that feedback and the corrections thereof are to be found in the comments in this PR. I have addressed a large majority of those comments, and am working on the last one that I plan to address as part of this PR. For the ones that I don't plan to address in this PR, I will create follow up tickets. Those will be added in the responses for the remaining feedback comments recorded in this PR's conversation.

Preliminary testing w/SPECjbb didn't yield reliable performance data from which to infer any performance improvements stemming from enabling adaptive tenuring. I believe that was because of the way SPECjbb is run, which causes excessive degenerate and full gc's. I plan to collect SPECjbb numbers with a fixed lower max HBIR so as to be able to discern performance differences from this change, as well as Extremem workloads. Those will be added here once ready over the next few days.

---------
### Progress

### Issue
 * [JDK-8311883](https://bugs.openjdk.org/browse/JDK-8311883): [Genshen] Adaptive tenuring threshold (**Enhancement** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/289/head:pull/289` \
`$ git checkout pull/289`

`$ git checkout pull/289` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/289/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 289`

`$ git pr show -t 289`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/289.diff">https://git.openjdk.org/shenandoah/pull/289.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/289#issuecomment-1631260827)


src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
```diff
@@ -532,6 +542,20 @@ void ShenandoahHeap::increase_object_age(oop obj, uint additional_age) {
...
+  // This is impossible to do unless we "freeze" ABA-type oscillations
+  // With Lilliput, we can do this more easily.
+  markWord w = obj->mark();
+  // We can do better for objects with inflated monitor
+  if (w.is_being_inflated() || w.has_displaced_mark_helper()) {
+    // Informs caller that we aren't able to determine the age
+    return markWord::max_age + 1; // sentinel
+  }
+  assert(w.age() <= markWord::max_age, "Impossible!");
   return w.age();
```
kdnilsen, line 559: Do we need to add region->age() to w.age()?

src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp
```diff
@@ -89,24 +99,27 @@ void ShenandoahEvacuationTracker::print_evacuations_on(outputStream* st,
...
+          old_region_ages.add(r->age(), r->get_live_data_words());
+        }
       }
     }
+    uint tenuring_threshold = heap->age_census()->tenuring_threshold();
+    st->print("Young regions: ");
+    young_region_ages.print_on(st, tenuring_threshold);
+    st->cr();
+    st->print("Old regions: ");
+    old_region_ages.print_on(st, tenuring_threshold);
```
ysramakrishna, line 121: `tenuring_threshold` in this context doesn't make any sense for old regions.

src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp
```diff
@@ -89,24 +99,27 @@ void ShenandoahEvacuationTracker::print_evacuations_on(outputStream* st,
...
+  if (_generational) {
+    AgeTable young_region_ages(false), old_region_ages(false);
+    for (uint i = 0; i < heap->num_regions(); ++i) {
+      ShenandoahHeapRegion* r = heap->get_region(i);
+      if (r->age() > 0 && r->age() < AgeTable::table_size) {
+        if (r->is_young()) {
+          young_region_ages.add(r->age(), r->get_live_data_words());
+        } else {
+          assert(r->is_old(), "Sanity");
+          old_region_ages.add(r->age(), r->get_live_data_words());
```
ysramakrishna, line 112: We currently don't do anything with this. Perhaps the idea is to use it in some way in the future?

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
```diff
@@ -438,8 +439,8 @@ void ShenandoahControlThread::process_phase_timings(const ShenandoahHeap* heap)
       ResourceMark rm;
       LogStream ls(lt);
       heap->phase_timings()->print_cycle_on(&ls);
-      ShenandoahEvacuationTracker::print_evacuations_on(&ls, &evac_stats.workers,
-                                                        &evac_stats.mutators);
+      evac_tracker->print_evacuations_on(&ls, &evac_stats.workers,
```
ysramakrishna, line 442: Does this need to be per instance.

src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp
```diff
@@ -53,14 +57,18 @@ void ShenandoahEvacuationStats::accumulate(const ShenandoahEvacuationStats* othe
   _bytes_completed += other->_bytes_completed;
   _evacuations_attempted += other->_evacuations_attempted;
   _bytes_attempted += other->_bytes_attempted;
-
-  _age_table.merge(&other->_age_table);
+  if (_generational && GenShenCensusAtEvac) {
+    // Discard const
+    _age_table.merge(((ShenandoahEvacuationStats*)other)->age_table());
```
ysramakrishna, line 62: Is the cast necessary?

src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp
```diff
@@ -23,27 +23,31 @@
...
   _bytes_attempted += bytes;
 }
 
-void ShenandoahEvacuationStats::end_evacuation(size_t bytes, uint age) {
+void ShenandoahEvacuationStats::end_evacuation(size_t bytes, bool young_gen, uint age) {
   ++_evacuations_completed;
   _bytes_completed += bytes;
-  if (age > 0) {
+  if (_generational && GenShenCensusAtEvac && young_gen) {
+    assert(age > 0, "Error");
```
ysramakrishna, line 50: Check that age is recorded _after_ incrementing at evacuation.

src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
```diff
@@ -458,14 +459,19 @@ class ShenandoahPrepareForGenerationalCompactionObjectClosure : public ObjectClo
...
       _empty_regions_pos(0),
       _old_to_region(old_to_region),
       _young_to_region(young_to_region),
       _from_region(nullptr),
       _old_compact_point((old_to_region != nullptr)? old_to_region->bottom(): nullptr),
       _young_compact_point((young_to_region != nullptr)? young_to_region->bottom(): nullptr),
-      _worker_id(worker_id) {}
+      _worker_id(worker_id) {
+    if (_heap->mode()->is_generational()) {
+      _tenuring_threshold = _heap->age_census()->tenuring_threshold();
```
ysramakrishna, line 472: Is this updated in the most recent stop-world marking, or is it using the one from the most recent concurrent collection? Could an interrupted concurrent marking cause a bogus census to be used here?

Partial censuses should be expunged.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
```diff
@@ -477,6 +477,11 @@ void ShenandoahGeneration::prepare_regions_and_collection_set(bool concurrent) {
         preselected_regions[i] = false;
       }
 
+      if (!GenShenCensusAtEvac) {
+        // Age table updates
+        heap->update_epoch();
```
ysramakrishna, line 482: A different name for the method to avoid confusion.

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
```diff
@@ -471,7 +479,9 @@ inline oop ShenandoahHeap::try_evacuate_object(oop p, Thread* thread, Shenandoah
   oop result = ShenandoahForwarding::try_update_forwardee(p, copy_val);
   if (result == copy_val) {
     // Successfully evacuated. Our copy is now the public one!
-    _evac_tracker->end_evacuation(thread, size * HeapWordSize, ShenandoahHeap::get_object_age(copy_val));
+    // If copying to the old generation, we don't care about recording
+    // object age in the census stats.
+    _evac_tracker->end_evacuation(thread, size * HeapWordSize, young_gen, young_gen ? ShenandoahHeap::get_object_age(copy_val) : 0);
```
ysramakrishna, line 484: Revisit this to see why we record age 0 for the case of old gen objects.

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
```diff
@@ -532,6 +542,20 @@ void ShenandoahHeap::increase_object_age(oop obj, uint additional_age) {
 
 uint ShenandoahHeap::get_object_age(oop obj) {
   markWord w = obj->has_displaced_mark() ? obj->displaced_mark() : obj->mark();
+  assert(w.age() <= markWord::max_age, "Impossible!");
+  return w.age();
+}
+
+uint ShenandoahHeap::get_object_age_concurrent(oop obj) {
```
ysramakrishna, line 549: Pla ce documentation comments distinguishing between the two methods here too.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
```diff
@@ -35,6 +35,28 @@
                             range,                                          \
                             constraint)                                     \
                                                                             \
+  product(bool, GenShenCensusAtEvac, true, EXPERIMENTAL,                    \
```
ysramakrishna, line 38: GenShen -> ShenandoahGenerational, perhaps.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.hpp
```diff
@@ -460,10 +461,29 @@ class ShenandoahHeapRegion {
...
-  void decrement_age() { if (_age-- == 0) { _age = 0; } }
-  void reset_age()     { _age = 0; }
+
+  void decrement_age() {
+    if (_age-- == 0) {
+      _age = 0;
+    } else {
+      CENSUS_NOISE(_youth++;)
+    }
+  }
```
ysramakrishna, line 479: Unused: remove.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.hpp
```diff
@@ -460,10 +461,29 @@ class ShenandoahHeapRegion {
 
   void set_affiliation(ShenandoahAffiliation new_affiliation);
 
-  uint age()           { return _age; }
+  // Region ageing and rejuvenation
+  uint age() { return _age; }
+  uint youth() {
+    return CENSUS_NOISE(_youth)
+           NO_CENSUS_NOISE(0);
```
ysramakrishna, line 468: May nit need in product build.

src/hotspot/share/gc/shenandoah/shenandoahMark.inline.hpp
```diff
@@ -57,8 +57,8 @@ void ShenandoahMark::dedup_string(oop obj, StringDedup::Requests* const req) {
   }
 }
 
-template <class T, StringDedupMode STRING_DEDUP>
-void ShenandoahMark::do_task(ShenandoahObjToScanQueue* q, T* cl, ShenandoahLiveData* live_data, StringDedup::Requests* const req, ShenandoahMarkTask* task) {
+template <class T, ShenandoahGenerationType GENERATION, StringDedupMode STRING_DEDUP>
```
ysramakrishna, line 60: Consider extracting the GENERATION type from T (closure's type).

src/hotspot/share/gc/shenandoah/shenandoahMark.inline.hpp
```diff
@@ -99,19 +99,33 @@ void ShenandoahMark::do_task(ShenandoahObjToScanQueue* q, T* cl, ShenandoahLiveD
...
-  size_t size = obj->size();
+template <ShenandoahGenerationType GENERATION>
+inline void ShenandoahMark::count_liveness(ShenandoahLiveData* live_data, oop obj, uint worker_id) {
+  const ShenandoahHeap* const heap = ShenandoahHeap::heap();
+  const size_t region_idx = heap->heap_region_index_containing(obj);
+  ShenandoahHeapRegion* const region = heap->get_region(region_idx);
+  const size_t size = obj->size();
+
+  // Age census for young objects
+  if (GENERATION == YOUNG && !GenShenCensusAtEvac) {
```
ysramakrishna, line 118: Fold the two together.

src/hotspot/share/gc/shenandoah/shenandoahMark.inline.hpp
```diff
@@ -99,19 +99,33 @@ void ShenandoahMark::do_task(ShenandoahObjToScanQueue* q, T* cl, ShenandoahLiveD
     // Avoid double-counting objects that are visited twice due to upgrade
     // from final- to strong mark.
     if (task->count_liveness()) {
-      count_liveness(live_data, obj);
+      count_liveness<GENERATION>(live_data, obj, worker_id);
```
ysramakrishna, line 102: Experiment with pulling count_liveness into the closure `cl` which already knows the generation type.

src/hotspot/share/gc/shenandoah/shenandoahPacer.hpp
```diff
@@ -27,6 +27,7 @@
 
 #include "gc/shenandoah/shenandoahNumberSeq.hpp"
 #include "gc/shenandoah/shenandoahPadding.hpp"
+#include "gc/shenandoah/shenandoahSharedVariables.hpp"
```
ysramakrishna, line 30: Check this again.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
```diff
@@ -35,6 +35,28 @@
...
+          "Object age census at evacuation, not mearking")                  \
```
ysramakrishna, line 39: "marking"

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
```diff
@@ -35,6 +35,28 @@
...
+                                                                            \
+  product(bool, GenShenAdaptiveTenuring, false, EXPERIMENTAL,               \
+          "Adaptive tenuring threshold")                                    \
+                                                                            \
+  product(uintx, GenShenMinTenuringThreshold, 1, EXPERIMENTAL,              \
+          "Floor for adaptive tenuring threshold")                          \
+                                                                            \
+  product(uintx, GenShenMaxTenuringThreshold, 16, EXPERIMENTAL,             \
```
ysramakrishna, line 47: Why not 15?

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
```diff
@@ -35,6 +35,28 @@
...
+          "Ceiling for adaptive tenuring threshold")                        \
+                                                                            \
+  product(double, GenShenTenuringMortalityRateThreshold, 0.3, EXPERIMENTAL, \
```
ysramakrishna, line 50: Experiment with different values here (0.1 may be more appropriate perhaps?).

Add guidance about which way (lower) to move this parm to keep objects in young gen longer.

src/hotspot/share/gc/shenandoah/shenandoahAgeCensus.cpp
```diff
@@ -0,0 +1,271 @@
...
+  size_t total= 0;
+  for (uint i = 1; i < MAX_COHORTS; i++) {
+    const size_t prev_pop = prev_pv->sizes[i-1];  // (i-1) OK because i >= 1
+    const size_t cur_pop  = cur_pv->sizes[i];
+    double mr = mortality_rate(prev_pop, cur_pop);
+    // Suppress printing when everything is zero
+    if (prev_pop + cur_pop > 0) {
+      log_info(gc, age)(UINTX_FORMAT "\t\t " SIZE_FORMAT "\t\t " SIZE_FORMAT "\t\t %.2f " ,
+                        (uintx)i, prev_pop, cur_pop, mr);
+    }
```
ysramakrishna, line 249: vertical alignment via tabbing. Specify units for volume.

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
```diff
@@ -480,6 +484,12 @@ class ShenandoahHeap : public CollectedHeap {
   inline size_t set_young_evac_reserve(size_t new_val);
   inline size_t get_young_evac_reserve() const;
 
+  // Methods related to age tables in generational mode for Young Gen
+  inline ShenandoahAgeCensus* age_census() const;
+  AgeTable* get_age_table();
+  void update_epoch();
+  void reset_epoch();
```
ysramakrishna, line 491: `epoch` overload/confusion with prior use. Consider if this is really needed; if so, find better names to avoid confusion with other uses of the term `epoch` or methods thereof.

src/hotspot/share/gc/shenandoah/shenandoahAgeCensus.cpp
```diff
@@ -0,0 +1,271 @@
...
+    " (min " UINTX_FORMAT ", max " UINTX_FORMAT")",
+    (uintx) _tenuring_threshold[_epoch], GenShenMinTenuringThreshold, GenShenMaxTenuringThreshold);
+}
+
+uint ShenandoahAgeCensus::compute_tenuring_threshold_work() {
+  // Starting with the largest non-zero population by age cohort
+  // and working down in age of cohorts,find the lowest age such
+  // that all higher ages have a mortality rate that is below a
+  // pre-specified threshold. We consider this to be the adaptive
+  // tenuring age to be used for the next cohort.
```
ysramakrishna, line 174: ... for the next _cycle_.

src/hotspot/share/gc/shenandoah/shenandoahAgeCensus.cpp
```diff
@@ -0,0 +1,271 @@
...
+void ShenandoahAgeCensus::update_epoch() {
+  assert(_epoch < MAX_SNAPSHOTS, "Out of bounds");
+  if (++_epoch >= MAX_SNAPSHOTS) {
+    _epoch=0;
+  }
+  // Merge data from local age tables into the global age table for the epoch,
+  // clearing the local tables.
+  _global_age_table[_epoch]->clear();
+  CENSUS_NOISE(_global_noise[_epoch].clear();)
+  if (!GenShenCensusAtEvac) {
```
ysramakrishna, line 120: In the case of `GenShenCensusAtEvac` this work is done by the calling method explicitly via `ingest` and `compute_...` 

This should probably be done in a separate method/API rather to keep the API/spec of this method clean.

ysramakrishna:
For the case of `!GenShenCensusAtEvac`, there was discussion regarding the appearance of age cohorts older than tenuring threshold for cohorts older than computed tenuring threshold. While objects older than tenuring threshold should be expected in the young generation because of being excluded from the collection set, it was surprising to have found that population in the census if it were not in the collection set. On the other hand, if it were in the collection set, one would have expected those instances to be promoted and not appear in subsequent censuses.

Check what's going on here and fix the problem, or find an explanation for the observed data.

Finally, in this case, since the data is available via region census, would it make sense to add it to the age table? We believe not, since the data may not be accurate because of undercounting the age. (Total this up in the noise vector to see what it looks like?)

src/hotspot/share/gc/shenandoah/shenandoahAgeCensus.cpp
```diff
@@ -0,0 +1,271 @@
...
+    assert(i > 0, "Error");
+    // Population & mortality rate of current cohort
+    const size_t cur_pop = cur_pv->sizes[i];
+    const size_t prev_pop = prev_pv->sizes[i-1];
+    const double mr = mortality_rate(prev_pop, cur_pop);
+    // We ignore any cohorts that had a very low population count, or
+    // that have a lower mortality rate than we care to age in young; these
+    // cohorts are considered eligible for tenuring when all older
+    // cohorts are.
+    if (i > 1 && (prev_pop < GenShenTenuringCohortPopulationThreshold ||
```
ysramakrishna, line 196: Consider getting a more accurate population for cohort 0 (by adding the population in each region above TAMS), so that we may consider the case of `i = 1` also above (i.e. `i >=1`). This would potentially allo tenuring threshold to go as low as 1 in some cases. This choice will need to be carefully considered.

ysramakrishna:
For the case of `!GenShenCensusAtEvac`, when older cohorts above the tenuring threshold in a previous epoch are finally included in the collection set and promoted, the mortality rate shows up as`1.0`, causing the new tenuring threshold to shoot up. This may prevent timely promotion of a cohort that might otherwise have been considered eligible for promotion. We should see about avoiding this effect while still allowing the survivor ageing window to expand when necessary.

ysramakrishna:
Consider using the census information collected at marking in the current cycle (and perhaps at evacuation in the previous cycle?) to inform the generation sizing/budgeting for the evacuation cycle to follow. Talk with Kelvin to discuss a suitable API and place to hook this into.

src/hotspot/share/gc/shenandoah/shenandoahHeap.hpp
ysramakrishna: Done.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
ysramakrishna: Done.

src/hotspot/share/gc/shenandoah/shenandoahAgeCensus.cpp
ysramakrishna: Done

src/hotspot/share/gc/shenandoah/shenandoahAgeCensus.cpp
ysramakrishna: Done.

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
ysramakrishna: It's a bit awkward to keep it static, and the cost of calling the instance method no worse. (We actually just make use of a single instance field `_generational` in the method. We could make that field static and leave the method static, but I am not sure it's worth doing.

src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp
ysramakrishna: Yes, it is on the path where we evacuate. It isn't for the case where it's on the marking path. (This is of course the evacuation path here.)

src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp
ysramakrishna: Avoided the cast by moving age table out of the structure (which makes sense anyway since it's not used in some cases).

src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp
ysramakrishna: Changed to just use the max tenuring threshold here, although that doesn't really make sense here either. We can rethink this in the future, or elide that part. Leaving as is (using max tenuring threshold) for the moment.

src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp
ysramakrishna: Leaving as is for now for informational purposes; we can deal with this suitably in the future if/when we use it in some way. We might also decide to elide it until such time.

src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp
```diff
@@ -53,14 +57,18 @@ void ShenandoahEvacuationStats::accumulate(const ShenandoahEvacuationStats* othe
...
+    // Discard const
+    _age_table.merge(((ShenandoahEvacuationStats*)other)->age_table());
+  }
 }
 
 void ShenandoahEvacuationStats::reset() {
   _evacuations_completed = _evacuations_attempted = 0;
   _bytes_completed = _bytes_attempted = 0;
-  _age_table.clear();
+  if (_generational && GenShenCensusAtEvac) {
```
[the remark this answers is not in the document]
ysramakrishna: Done for this and other flags beginning "GenShen*" -> "ShenandoahGenerational*". 

Please feel free to suggest better (and/or shorter) names.

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
ysramakrishna: To answer the question, "why 0?" : it was because I didn't want to extract the age for the old_gen case, because it isn't used in that case (the census doesn't record old gen object ages). I'll see about cleaning this up a bit.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
ysramakrishna: Done for all these "GenShen" flags.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
ysramakrishna: Done.

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
ysramakrishna: Cleaned this up to avoid this pattern. Hopefully, it reads a bit cleaner and less clever (and error-prone) now.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.hpp
ysramakrishna: Removed.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.hpp
ysramakrishna: Indeed. Fixed.

src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp
ysramakrishna: It turns out that the assertion `(age > 0)` is still too strong because it might not be an aging cycle, even though we might be evacuating age 0 objects (without, in this case, incrementing their age).

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
ysramakrishna: Done.

src/hotspot/share/gc/shenandoah/shenandoahMark.inline.hpp
ysramakrishna: Done.

src/hotspot/share/gc/shenandoah/shenandoahAgeCensus.cpp
ysramakrishna: Done.

src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp
```diff
@@ -89,24 +106,20 @@ void ShenandoahEvacuationTracker::print_evacuations_on(outputStream* st,
   mutators->print_on(st);
   st->cr();
 
-  AgeTable young_region_ages(false), old_region_ages(false);
   ShenandoahHeap* heap = ShenandoahHeap::heap();
-  for (uint i = 0; i < heap->num_regions(); ++i) {
-    ShenandoahHeapRegion* r = heap->get_region(i);
-    if (r->age() > 0 && r->age() < AgeTable::table_size) {
```
ysramakrishna, line 96: It turns out that this was previously needed because one could not bound the region age. For now, I am dropping this test and instead bounding region ages, clamping them at `markWord::max_age`; see `ShenandoahHeapRegion::increment_age()`.

Should we reach a situation where we need to use larger ages than this, we'll need to revisit this code.

At this time, region ages pose a number of issues including injection of noise into object demographics and mortality rate calculations which drive our adaptive tenuring algorithm.

src/hotspot/share/gc/shared/ageTable.hpp
```diff
@@ -52,17 +53,18 @@ class AgeTable {
...
   // clear table
   void clear();
+  // check whether it's clear
+  bool is_clear() PRODUCT_RETURN0;
 
   // add entry
   inline void add(oop p, size_t oop_size);
 
   void add(uint age, size_t oop_size) {
-    assert(age > 0 && age < table_size, "invalid age of object");
```
ysramakrishna, line 60: We now also include age 0 objects in the census for the case of concurrent collectors in the young generation. This slot would have `sizes[0] == 0` for our traditional use cases for generational stop-world collectors.

I'll upstream this change separately in a sibling PR/JBS. TBD.

src/hotspot/share/gc/shenandoah/shenandoahPacer.hpp
ysramakrishna: Needed for `ShenandoahSharedFlag` (at least). A previous include somewhere else satisfied all previous uses which unravelled when some of the files (notably shenandoah thread local data's .cpp file was moved to its own file).

Leaving as is.

src/hotspot/share/gc/shenandoah/shenandoahMark.inline.hpp
ysramakrishna: See comment at line 60 further above. This will be addressed separately as a refactor to follow to avoid complicating this PR with more changes.

src/hotspot/share/gc/shenandoah/shenandoahMark.inline.hpp
ysramakrishna: This might be possible via the suggestion/comment further down at line 102. However, this might require a bit more refactoring of other code, which I'd prefer to do not here in this ticket, but separately.

I'll defer that to a separate PR for which I'll open a JBS in anticipation and link it here.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
ysramakrishna: We'll need to discuss this. Allowing this to be 16 (i.e. 1 greater than max_age) allows for a potential "never tenure" behavior. Otherwise, age 15 objects would always become eligible for tenuring. I think allowing it to be above 15 (i.e. allowing 16, and allowing objects to stay in young till and including age 15) would seem to make sense.

I'll set the default to 15 (thus never allowing objects that , but allow users to set it to 16, simulating "never tenure", if both min and max are set to 16.

ysramakrishna: Provide guidance here on how to use these min and max values to obtain different behavior, while picking sensible defaults (i.e. 0 and 15).

ysramakrishna:
> Consider using the census information collected at marking in the current cycle (and perhaps at evacuation in the previous cycle?) to inform the generation sizing/budgeting for the evacuation cycle to follow. Talk with Kelvin to discuss a suitable API and place to hook this into.

Tracked in a separate PR/JBS.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
ysramakrishna: Done. Updated guidance in documentation.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
ysramakrishna: Done; changed down to 0.1, added documentation guidance. Experiments pending to establish what might work well most of the time.

src/hotspot/share/gc/shenandoah/shenandoahAgeCensus.cpp
ysramakrishna: Done. Default ceiling is set by min age which defaults to 1.

src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
ysramakrishna: It is true that cancellations and partial censuses could lead to bogus partial censuses being used. I'll locate the right place in the cancellation path(s) where partial censuses will  need to be expunged. Essentially, these would be the points where concurrent marking is canceled (when censuses are conducted during the marking phase) and where concurrent evacuation is canceled (when censuses are conducted during the evacuation phase). All other cancellations are fine because the census data will be valid despite cancellation in those cases.

ysramakrishna: In the event of degeneration, if a census-collection phase is interrupted mid-stream and restarted afresh, the partial census will need to be expunged. It'll then be replaced by the new census (in e.g. degenerate GC's mark or evac). In the case of a degenerate STW GC being cancelled and degenerating further to a full STW GC, during _its_ census-collection phase the partial census will be expunged. Any completed censuses from any census taking phase that persisted will be used as before. Currently, all calculations are independent of the (start and end) time of the (completed) censuses being used, but a future enhancement might potentially consider a temporal dimension. We'll capture the times in the census today, but those are not used by the algorithm today.

src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp
```diff
@@ -23,44 +23,59 @@
...
 #include "runtime/threadSMR.inline.hpp"
 #include "runtime/thread.hpp"
 
-ShenandoahEvacuationStats::ShenandoahEvacuationStats()
+ShenandoahEvacuationStats::ShenandoahEvacuationStats(bool generational)
   : _evacuations_completed(0), _bytes_completed(0),
     _evacuations_attempted(0), _bytes_attempted(0),
-    _age_table(false) {}
+    _generational(generational) {
+  if (generational && (ShenandoahGenerationalCensusAtEvac || !ShenandoahGenerationalAdaptiveTenuring)) {
```
earthling-amzn, line 38: Could we drop `Generational` from these options? It's clear enough (at least to me) that these apply to the generational mode.

src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp
```diff
@@ -23,44 +23,59 @@
...
+    _age_table = new AgeTable(false);
+  }
+}
+
+AgeTable* ShenandoahEvacuationStats::age_table() const {
+  assert(_generational && (ShenandoahGenerationalCensusAtEvac || !ShenandoahGenerationalAdaptiveTenuring), "Don't call");
```
earthling-amzn, line 44: Maybe hoist expression `_generational && (ShenandoahGenerationalCensusAtEvac || !ShenandoahGenerationalAdaptiveTenuring` into a `_use_age_table` (or such) field?

src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp
```diff
@@ -121,7 +134,8 @@ class ShenandoahStatAggregator : public ThreadClosure {
 };
 
 ShenandoahCycleStats ShenandoahEvacuationTracker::flush_cycle_to_global() {
-  ShenandoahEvacuationStats mutators, workers;
+  bool gen_mode = ShenandoahHeap::heap()->mode()->is_generational();
```
earthling-amzn, line 137: Why not use the `_generational` member here?

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
```diff
@@ -464,13 +465,30 @@ void ShenandoahGeneration::prepare_regions_and_collection_set(bool concurrent) {
...
   }
 
+  // Tally the census counts and compute the adaptive tenuring threshold
+  if (is_generational && ShenandoahGenerationalAdaptiveTenuring && !ShenandoahGenerationalCensusAtEvac) {
+    // Objects above TAMS weren't included in the age census. Since they were all
+    // allocated in this cycle they belong in the age 0 cohort. We walk over all
+    // young regions and sum the volume of objects between TAMS and top.
+    ShenandoahUpdateCensusZeroCohortClosure age0_cl(complete_marking_context());
+    heap->young_generation()->heap_region_iterate(&age0_cl);
+    size_t age0_pop = age0_cl.get_pop();
```
earthling-amzn, line 475: `get_population`?

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
```diff
@@ -461,7 +467,9 @@ inline oop ShenandoahHeap::try_evacuate_object(oop p, Thread* thread, Shenandoah
 
   oop copy_val = cast_to_oop(copy);
 
+  bool young_gen = false;
   if (mode()->is_generational() && target_gen == YOUNG_GENERATION && is_aging_cycle()) {
+    young_gen = true;
```
earthling-amzn, line 472: `young_gen` looks unused?

kdnilsen:
I've run this through some Extremem workloads.  Good news is I see no regressions.  On the other hand, I am not yet seeing huge benefit.  (I may be seeing a decrease in degenerated cycles, but need to run a few more tests to be sure.)

One concern is that we are still improperly identifying promotions as mortality.  I'm attaching a log with some of my comments preceded by ;; at the start of lines.  See line 3421 of the log, for example.  IN GC(16), we chose tenure age 2.
Then we promoted in place 420 regions.  This caused us to believe there was high mortality of ages 3-6, but really there was no mortality and only promotions.  In GC(17), we should have stayed with tenure age 2.

I think the way to fix this is to only scan from 1 to the current tenure age when you select a new tenure age.  If there is no mortality at the current tenure age, then we can set the new tenure age to 1 + current tenure age.
[auto-tenure.out.txt](https://github.com/openjdk/shenandoah/files/12031253/auto-tenure.out.txt)

kdnilsen:
I'm thinking we would want to make -XX:+ShenandoahGenerationalAdaptiveTenuring -XX:-ShenandoahGenerationalCensusAtEvac the default behavior in this PR.  Are there reasons not to?

ysramakrishna:
> 

I agree that that's the desired ideal end goal as I see it. I'd like to have good, reliable performance data that justifies it first though, or at least shows that it's performance neutral at this time.

I'll make the remaining changes stemming from your tests and run some more performance tests as well.

ysramakrishna:
> I'm thinking we would want to make -XX:+ShenandoahGenerationalAdaptiveTenuring -XX:-ShenandoahGenerationalCensusAtEvac the default behavior in this PR. Since we do not yet have a large community of GenShen production users, are there reasons not to make these the defaults?

I agree that that's the desired ideal end goal as I see it. I'd like to have good, reliable performance data that justifies it first though, or at least shows that it's performance neutral at this time.

I'll make the remaining changes stemming from your tests and run some more performance tests as well.

ysramakrishna:
> I've run this through some Extremem workloads. Good news is I see no regressions. On the other hand, I am not yet seeing huge benefit. (I may be seeing a decrease in degenerated cycles, but need to run a few more tests to be sure.)
> 
> One concern is that we are still improperly identifying promotions as mortality. I'm attaching a log with some of my comments preceded by ;; at the start of lines. See line 3421 of the log, for example. IN GC(16), we chose tenure age 2. Then we promoted in place 420 regions. This caused us to believe there was high mortality of ages 3-6, but really there was no mortality and only promotions. In GC(17), we should have stayed with tenure age 2.
> 
> I think the way to fix this is to only scan from 1 to the current tenure age when you select a new tenure age. If there is no mortality at the current tenure age, then we can set the new tenure age to 1 + current tenure age. [auto-tenure.out.txt](https://github.com/openjdk/shenandoah/files/12031253/auto-tenure.out.txt)

Thanks for helping me reason through this. My instinct was that the temporary wide-opening of the survivor window (or conversely the raising of the tenuring bar) would be harmless because the next census would realize that the oldest cohort of young survivors in this gc did not exhibit sufficient mortality and quickly lower the bar again before the next subsequent collection. However, this does indeed result in the aforementioned survivor cohort staying in the young gen for one GC longer which could be bad from a performance perspective.

My second idea of leaving that window open (or the bar raised, whichever your metaphor of preference is) and looking at all of the cohorts in the young generation every time was that in the event that other criteria (e.g. low garbage density) had us keeping a region in the young gen longer where we were already including that cohort in our census, it would make sense for us to consider its mortality each time we encountered it.

However, upon further reflection, it might make sense to decouple these effects and ignore cohorts that the algorithm would have "logically" promoted, but which were physically were not because of other considerations. Such cohorts then would constitute a case of "delayed tenure cohorts", but whose tenure isn't readjudicated at each epoch just because they were held back for other considerations. Thinking in terms of this academic tenuring metaphor makes me feel that your proposal is in fact the correct one (and would be considered fairer in an academic setting than my revisionist original algorithm :-)

I'll make that change and gather performance data. Many thanks for the performance data that you obtained and shared with me, and for the related discussion!

src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp
ysramakrishna: The code can be called also in non-generational (single generation) mode, and records the same stats. Obviously we want to elide the age census that we do in generational mode in that situation. Hence the option.

Let me know if you had something else in mind here for avoiding the `_generational` attribute or of skinning this cat in a different way.

src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp
ysramakrishna: Good idea. Done!

src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp
ysramakrishna: Done.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
ysramakrishna: Done.

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
ysramakrishna: Deleted.

ysramakrishna:
> I'm thinking we would want to make -XX:+ShenandoahGenerationalAdaptiveTenuring -XX:-ShenandoahGenerationalCensusAtEvac the default behavior in this PR. Since we do not yet have a large community of GenShen production users, are there reasons not to make these the defaults?

Done.

ysramakrishna:
> I've run this through some Extremem workloads. Good news is I see no regressions. On the other hand, I am not yet seeing huge benefit. (I may be seeing a decrease in degenerated cycles, but need to run a few more tests to be sure.)
> 
> One concern is that we are still improperly identifying promotions as mortality. I'm attaching a log with some of my comments preceded by ;; at the start of lines. See line 3421 of the log, for example. IN GC(16), we chose tenure age 2. Then we promoted in place 420 regions. This caused us to believe there was high mortality of ages 3-6, but really there was no mortality and only promotions. In GC(17), we should have stayed with tenure age 2.
> 
> I think the way to fix this is to only scan from 1 to the current tenure age when you select a new tenure age. If there is no mortality at the current tenure age, then we can set the new tenure age to 1 + current tenure age. [auto-tenure.out.txt](https://github.com/openjdk/shenandoah/files/12031253/auto-tenure.out.txt)

Done.

ysramakrishna:
> JDK-8311883 [GenShen] Adaptive tenuring
> 
...
> Preliminary testing w/SPECjbb didn't yield reliable performance data from which to infer any performance improvements stemming from enabling adaptive tenuring. I believe that was because of the way SPECjbb is run, which causes excessive degenerate and full gc's. I plan to collect SPECjbb numbers with a fixed lower max HBIR so as to be able to discern performance differences from this change, as well as Extremem workloads. Those will be added here once ready over the next few days.


Update on performance efforts:

I've run SPECjbb and Extremem under various configurations of the benchmarks and of Generational Shenandoah in order to separate the performance of reference and specimen. Unfortunately, it appears as if the triggering mechanisms often cause degenerate collections which has been a bit of a challenge to completely eliminate. This appears to inject enough variability into the results that any differences in performance between the two is difficult to discern. 

SPECjbb configurations included fixed preset IR among others, and Extremem configurations (thanks to Kelvin for advice) included various load characteristics. Generational Shenandoah configurations included guaranteed collection interval triggers (variously and in combination for young & old), and disabling garbage density as a criterion for inclusion in the collection set.

The only configuration in which I was able to fully eliminate degenerate GCs for these workloads was Extremem "hefty load" with a 45 g heap. This produced exactly one degenerate GC in each case during the early part of the run. Setting a guaranteed collection interval eliminated this degenerate collection.

A comparison was then made for the time in each of the concurrent phases, both total and average over all cycles. The differences were a wash.

The performance numbers for adaptive tenuring were based on the default settings to be found in
src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp.

While no positive or negative performance impacts have yet been measured, the framework is now in place to allow further experimentation in the future in order to investigate and potentially extract any potential benefits from adaptive tenuring.

I'm happy to integrate this now, or do further future experiments, including coupling this with the size budgeting for generations to demonstrate impact before this goes in. I am open to any and all suggestions.

I'll attach to the ticket the design document upon which the current adaptive tenuring is based, and where we discuss some alternatives and some pitfalls we encountered along the way.

kdnilsen (approved):
Thanks.  This was a lot of work.  Good to see it integrated.

src/hotspot/share/gc/shared/ageTable.hpp
ysramakrishna: https://bugs.openjdk.org/browse/JDK-8314329 will track this and other changes to the age table that need to be separately udstreamed.

src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
ysramakrishna: The time tracking of the census is a potential future enhancement.

Today each new census clears any old partial or interrupted censuses, but we do not track the start and end times of completed censuses.

src/hotspot/share/gc/shenandoah/shenandoahMark.inline.hpp
ysramakrishna: Tracked in task ticket https://bugs.openjdk.org/browse/JDK-8314331

ysramakrishna:
/summary

8311883: [Genshen] Adaptive tenuring threshold

Generational Shenandoah currently has the notion of a tenuring threshold but it isn't dynamically adapted, but rather kept fixed at 7.

We now adapt the tenuring threshold based on object demographics as determined by a recent GC that visits objects in the young generation.

We keep track of age-cohort populations at each minor GC epoch and use the historical data to determine if it would be a good idea to tenure or not based on measured mortality rates. A few tunable (experimental) knobs are exposed to play with these to determine some good settings in the future.

The object census is conducted by default at marking, and is subject to noise on account of objects whose age could not be determined because of displaced header. In this case, the computed tenuring threshold is used in the following evacuation of the same cycle. Optionally, the census can be conducted at evacuation time, but sees only objects that are in the collection set. In this case, the tenuring threshold that is computed is used for tenuring decisions in the next evacuation cycle. 

Other variants are possible, and may be implemented / tested in the future as opportunity and data permit.

The computed tenuring threshold has not yet been coupled with size budgeting, but will be in a followup PR.

At this time, performance measurements have not shown any benefit, but we believe that with the framework now in place, we may be able to find an adaptive tenuring algorithm that works better than the current one and provide performance benefits. Adaptive tenuring is enabled by default, but can be optionally disabled to mimic previous behavior.

ysramakrishna:
/summary

Generational Shenandoah currently has the notion of a tenuring threshold but it isn't dynamically adapted, but rather kept fixed at 7.

We now adapt the tenuring threshold based on object demographics as determined by a recent GC that visits objects in the young generation.

We keep track of age-cohort populations at each minor GC epoch and use the historical data to determine if it would be a good idea to tenure or not based on measured mortality rates. A few tunable (experimental) knobs are exposed to play with these to determine some good settings in the future.

The object census is conducted by default at marking, and is subject to noise on account of objects whose age could not be determined because of displaced header. In this case, the computed tenuring threshold is used in the following evacuation of the same cycle. Optionally, the census can be conducted at evacuation time, but sees only objects that are in the collection set. In this case, the tenuring threshold that is computed is used for tenuring decisions in the next evacuation cycle.

Other variants are possible, and may be implemented / tested in the future as opportunity and data permit.

The computed tenuring threshold has not yet been coupled with size budgeting, but will be in a followup PR.

At this time, performance measurements have not shown any benefit, but we believe that with the framework now in place, we may be able to find an adaptive tenuring algorithm that works better than the current one and provide performance benefits. Adaptive tenuring is enabled by default, but can be optionally disabled to mimic previous behavior.

ysramakrishna:
I am integrating these changes with the GHA failures https://github.com/ysramakrishna/shenandoah/actions/runs/5898305273 that are tracked in https://bugs.openjdk.org/browse/JDK-8311843

/integrate

PR #290, opened 2023-06-22
8310680: GenShen: In-place region promotions may fail

earthling-amzn:
There are a few small changes here:
* In `op-update-refs-final`, only active regions are aged.
* In `op-final-roots`, a region's age may be reset outside of an aging cycle.
* When the field used to save `top` before promotion is set, it indicates the region has been chosen for in-place-promotion.

---------
### Progress

### Issue
 * [JDK-8310680](https://bugs.openjdk.org/browse/JDK-8310680): GenShen: In-place region promotions may fail (**Bug** - P4)

### Reviewers
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [0c335c33](https://git.openjdk.org/shenandoah/pull/290/files/0c335c339a3e1935706b9631e204f68652fe809e)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/290/head:pull/290` \
`$ git checkout pull/290`

`$ git checkout pull/290` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/290/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 290`

`$ git pr show -t 290`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/290.diff">https://git.openjdk.org/shenandoah/pull/290.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/290#issuecomment-1603419075)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -1718,9 +1718,11 @@ private:
           // doing this work during a safepoint.  We cannot put humongous regions into the collection set because that
           // triggers the load-reference barrier (LRB) to copy on reference fetch.
           r->promote_humongous();
-        } else if (r->is_regular() && (r->get_top_before_promote() != nullptr)
-                  && (r->garbage_before_padded_for_promote() < old_garbage_threshold)
-                  && (r->get_top_before_promote() == tams)) {
+        } else if (r->is_regular() && (r->get_top_before_promote() != nullptr)) {
+          assert(r->garbage_before_padded_for_promote() < old_garbage_threshold,
+                 "Region " SIZE_FORMAT " has too much garbage for promotion", r->index());
+          assert(r->get_top_before_promote() == tams,
+                 "Region " SIZE_FORMAT " has been used for allocations before promotion", r->index());
           // Likewise, we cannot put promote-in-place regions into the collection set because that would also trigger
           // the LRB to copy on reference fetch.
           r->promote_in_place();
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -1718,7 +1718,9 @@ private:
           // doing this work during a safepoint.  We cannot put humongous regions into the collection set because that
           // triggers the load-reference barrier (LRB) to copy on reference fetch.
           r->promote_humongous();
-        } else if (r->is_regular() && (r->garbage_before_padded_for_promote() < old_garbage_threshold) && (r->get_top_before_promote() == tams)) {
+        } else if (r->is_regular() && (r->get_top_before_promote() != nullptr)
+                  && (r->garbage_before_padded_for_promote() < old_garbage_threshold)
+                  && (r->get_top_before_promote() == tams)) {
           // Likewise, we cannot put promote-in-place regions into the collection set because that would also trigger
           // the LRB to copy on reference fetch.
           r->promote_in_place();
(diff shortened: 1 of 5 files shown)


src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -1718,7 +1718,9 @@ class ShenandoahGenerationalEvacuationTask : public WorkerTask {
           // doing this work during a safepoint.  We cannot put humongous regions into the collection set because that
           // triggers the load-reference barrier (LRB) to copy on reference fetch.
           r->promote_humongous();
-        } else if (r->is_regular() && (r->garbage_before_padded_for_promote() < old_garbage_threshold) && (r->get_top_before_promote() == tams)) {
+        } else if (r->is_regular() && (r->get_top_before_promote() != nullptr)
+                  && (r->garbage_before_padded_for_promote() < old_garbage_threshold)
```
kdnilsen, line 1722: I think my earlier guidance may have been wrong here.  I think we can rewrite this as:
else if (r->is_regular() && (r->get_top_before_promote() != nullptr)) {
  assert(r->garbage_before_padded_for_promote() < old_garbage_threshold, ...)
  assert (r->get_top_before_promote() == tams, ...)

kdnilsen (approved):
Thanks.  Looks good to me.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
earthling-amzn: Okay, will make that change.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -1718,7 +1718,11 @@ class ShenandoahGenerationalEvacuationTask : public WorkerTask {
           // doing this work during a safepoint.  We cannot put humongous regions into the collection set because that
           // triggers the load-reference barrier (LRB) to copy on reference fetch.
           r->promote_humongous();
-        } else if (r->is_regular() && (r->garbage_before_padded_for_promote() < old_garbage_threshold) && (r->get_top_before_promote() == tams)) {
+        } else if (r->is_regular() && (r->get_top_before_promote() != nullptr)) {
+          assert(r->garbage_before_padded_for_promote() < old_garbage_threshold,
+                 "Region " SIZE_FORMAT " has too much garbage for promotion", r->index());
+          assert(r->get_top_before_promote() == tams,
```
ysramakrishna, line 1724: Where was this second condition established when identifying a region as eligible for promotion-in-place? As far as I can see the second condition is tested _after_ the top has been saved.

ysramakrishna (approved):
LGTM except for the second assert that was moved from the test into the assert.

It would be good if the condition for promote-in-place were extracted into a method to avoid inconsistency of use in different places, but I am not sure how that can be done easily or if it's overkill to do so, since the condition is tested and then we set a sentinel that signals that condition for the later step.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
earthling-amzn: That's a good catch. We do check this condition, but we do it after we set the sentinel that indicates this region should be promoted. I'll fix this.

earthling-amzn:
Right - the sentinel value is meant to remove duplication of the logic. I added a comment explaining this in the code.

ysramakrishna (approved):
LGTM! 🚢

PR #291, opened 2023-06-22
8310574 GenShen: Should not update-references for in-place-promotions

earthling-amzn:
When a cycle _only_ promotes regions in place, it is not necessary to update refs.

---------
### Progress

### Issue
 * [JDK-8310574](https://bugs.openjdk.org/browse/JDK-8310574): GenShen: Should not update-references for in-place-promotions (**Bug** - P4)

### Reviewers
 * [Y. Srinivas Ramakrishna](https://openjdk.org/census#ysr) (@ysramakrishna - Committer) ⚠️ Review applies to [2d0460e8](https://git.openjdk.org/shenandoah/pull/291/files/2d0460e80836916e5c8d0b69ce98e325ceb2de69)
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [2d0460e8](https://git.openjdk.org/shenandoah/pull/291/files/2d0460e80836916e5c8d0b69ce98e325ceb2de69)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/291/head:pull/291` \
`$ git checkout pull/291`

`$ git checkout pull/291` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/291/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 291`

`$ git pr show -t 291`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/291.diff">https://git.openjdk.org/shenandoah/pull/291.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/291#issuecomment-1603445076)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
@@ -127,11 +127,15 @@ bool ShenandoahConcurrentGC::collect(GCCause::Cause cause) {
 
     // Concurrent mark roots
     entry_mark_roots();
-    if (check_cancellation_and_abort(ShenandoahDegenPoint::_degenerated_roots)) return false;
+    if (check_cancellation_and_abort(ShenandoahDegenPoint::_degenerated_roots)) {
+      return false;
+    }
 
     // Continue concurrent mark
     entry_mark();
-    if (check_cancellation_and_abort(ShenandoahDegenPoint::_degenerated_mark)) return false;
+    if (check_cancellation_and_abort(ShenandoahDegenPoint::_degenerated_mark)) {
+      return false;
+    }
   }
 
   // Complete marking under STW, and start evacuation
@@ -196,18 +200,24 @@ bool ShenandoahConcurrentGC::collect(GCCause::Cause cause) {
   if (heap->is_evacuation_in_progress()) {
     // Concurrently evacuate
     entry_evacuate();
-    if (check_cancellation_and_abort(ShenandoahDegenPoint::_degenerated_evac)) return false;
+    if (check_cancellation_and_abort(ShenandoahDegenPoint::_degenerated_evac)) {
+      return false;
+    }
   }
 
   if (heap->has_forwarded_objects()) {
     // Perform update-refs phase.
     vmop_entry_init_updaterefs();
     entry_updaterefs();
-    if (check_cancellation_and_abort(ShenandoahDegenPoint::_degenerated_updaterefs)) return false;
+    if (check_cancellation_and_abort(ShenandoahDegenPoint::_degenerated_updaterefs)) {
+      return false;
+    }
 
     // Concurrent update thread roots
     entry_update_thread_roots();
-    if (check_cancellation_and_abort(ShenandoahDegenPoint::_degenerated_updaterefs)) return false;
+    if (check_cancellation_and_abort(ShenandoahDegenPoint::_degenerated_updaterefs)) {
+      return false;
+    }
 
     vmop_entry_final_updaterefs();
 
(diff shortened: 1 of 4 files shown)


src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
```diff
@@ -782,18 +784,21 @@ void ShenandoahConcurrentGC::op_final_mark() {
         if (ShenandoahVerify) {
           heap->verifier()->verify_before_evacuation();
         }
-        // TODO: we do not need to run update-references following evacuation if collection_set->is_empty().
 
         heap->set_evacuation_in_progress(true);
```
ysramakrishna, line 788: The new code at lines 796-801 makes sense. However, then why do we set evacuation in progress at line 788 above if there might be nothing to evacuate in the event that the collection set is empty, in which case the boolean parm should be like at line 799 below? If so, wouldn't the two conditions be synonymous? If not, what is the information in one that is not in the other. In that case, could the `set_evacuation_in_progress` method be renamed for greater clarity?

src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
```diff
@@ -829,15 +834,16 @@ void ShenandoahConcurrentGC::op_final_mark() {
...
 
         // Verify before arming for concurrent processing.
         // Otherwise, verification can trigger stack processing.
         if (ShenandoahVerify) {
           heap->verifier()->verify_during_evacuation();
         }
 
+        // From here on, we need to update references.
+        heap->set_has_forwarded_objects(true);
+
```
ysramakrishna, line 846: At both places, this makes sense since the collection set isn't empty, apropos of my earlier comment.

What is the reason to not have `set_has_forwarded_objects` being true at verification time? I notice that you made the same change in the earlier part as well. Would it be worth a documentation comment?

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
```diff
@@ -800,7 +800,7 @@ void ShenandoahVerifier::verify_at_safepoint(const char* label,
         break;
       case _verify_gcstate_evacuation:
         enabled = true;
-        expected = ShenandoahHeap::HAS_FORWARDED | ShenandoahHeap::EVACUATION;
+        expected = ShenandoahHeap::EVACUATION;
```
ysramakrishna, line 803: May be the semantics of these states should be explained via a state diagram where these are defined using "may" and "must" modalities, so the semantics are clear and there isn't ambiguity. ("has" to me feels like a "must" modality, in the modal logic sense; `EVACUATION` could similarly be clarified with a `MAY` or `MUST` modal sense.)

I am rambling, but hopefully you get the idea of what I am trying to articulate :-)

ysramakrishna (commented):
I'll go back and re-read this so I understand the higher level shape of this, but I don't think I understand the transitions and states sufficiently well here to "approve". Hopefully my comments & questions help a bit in terms of a review of sorts.

src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
earthling-amzn: I'll add a comment to clarify. The in-place promotions happen during the evacuation phase, _even when the collection set is empty_. We've discussed pulling this action into a separate/distinct phase, or _only_ allowing in-place-promotions  when the collection set is _not_ empty (thus eliminating the special case of cycles that _only_ perform in-place-promotions.

src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
earthling-amzn: Previously, the verifier treated "evacuation in progress" and "heap has forwarded objects" as the same state. We now have a case where "evacuation"  (promote-in-place) may happen without forwarded objects.

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
earthling-amzn: I think the `expected` state here is effectively a `MUST`.

kdnilsen (approved):

ysramakrishna (approved):

PR #304, opened 2023-08-04
8313574: GenShen: Completing a global mark should also complete an old mark

earthling-amzn:
The mark context is shared, but each generation maintains a separate flag (`ShenandoahGeneration::_is_marking_complete`) to record the status of the mark. The remembered set scan will use the mark bitmap only if the old generation mark is complete. Otherwise, it walks the old regions and requires them to be parseable.

Presently, we make the heap parseable after global marking finishes, but this work can be deferred to the start of an old mark if we had global mark record completion of old marking.

---------
### Progress

### Issue
 * [JDK-8313574](https://bugs.openjdk.org/browse/JDK-8313574): GenShen: Completing a global mark should also complete an old mark (**Bug** - P3)

### Reviewers
 * [Y. Srinivas Ramakrishna](https://openjdk.org/census#ysr) (@ysramakrishna - Committer) ⚠️ Review applies to [2bf50f3f](https://git.openjdk.org/shenandoah/pull/304/files/2bf50f3fe3fc2a25a8679c23f3870a1f13361446)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/304/head:pull/304` \
`$ git checkout pull/304`

`$ git checkout pull/304` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/304/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 304`

`$ git pr show -t 304`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/304.diff">https://git.openjdk.org/shenandoah/pull/304.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/304#issuecomment-1666013873)


kdnilsen (approved):
Looks good.  Thanks for prompt fix.

ysramakrishna (approved):
LGTM.

Please add a stack retrace of the failure you'd typically see (and perhaps the failing test) that identified this issue into the associated JBS ticket.

Fix looks good.

ysramakrishna:
> LGTM.
> 
> Please add a stack retrace of the failure you'd typically see (and perhaps the failing test) that identified this issue into the associated JBS ticket.
> 
> Fix looks good.

Also I am guessing the JBS issue should be a "bug" rather than just called an enhancement (which I agree it is also :-). Or may be the bug is tracked in a different ticket and will be closed as a duplicate of this enhancement?

earthling-amzn:
Converted to draft because of failures in Dacapo suite.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.hpp
```diff
@@ -145,7 +145,8 @@ class ShenandoahOldHeuristics : public ShenandoahHeuristics {
   // last_old_region_index() entries, or memory may be corrupted when this function overwrites the
   // end of the array.
   unsigned int get_coalesce_and_fill_candidates(ShenandoahHeapRegion** buffer);
-  bool has_coalesce_and_fill_candidates() { return _last_old_region > 0; }
+  bool has_coalesce_and_fill_candidates() const { return _last_old_region > 0; }
+  size_t coalesce_and_fill_candidates() const { return _last_old_region; }
```
ysramakrishna, line 149: The return value from `coalesce_and_fill_candidates()` (which by the way could be confused for an imperative action verb) lready available in `last_old_region_index()` although might be a good idea to rename appropriately? Also fine to leave as is, but it sounds to me like one name might work for both uses.

src/hotspot/share/gc/shenandoah/shenandoahDegeneratedGC.cpp
```diff
@@ -96,13 +96,6 @@ void ShenandoahDegenGC::op_degenerated() {
 
 #ifdef ASSERT
   if (heap->mode()->is_generational()) {
-    if (_generation->is_global()) {
-      // We can only get to a degenerated global cycle _after_ a concurrent global cycle
-      // has been cancelled. In which case, we expect the concurrent global cycle to have
-      // cancelled the old gc already.
-      assert(!heap->is_old_gc_active(), "Old GC should not be active during global cycle");
-    }
-
```
ysramakrishna, line 105: Is this not true?

You could rearrange the test into the assertion check though:

```
    assert(!_generation->is_global() || !heap->is_old_gc_active, "Old GC should not be active during global cycle");
```
(or its de-Morganization, if you prefer).

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
```diff
@@ -380,46 +380,48 @@ void ShenandoahOldGeneration::transition_to(State new_state) {
...
+//           |   |   |    |     +-----------------+     +----------------------+
+//           |   |   |    |     |    MARKING      | <-> |       YOUNG GC       |
+//           |   |   +----|-----|                 |     | (RSet Parses Region) |
+//           |   |        |     +-----------------+     +----------------------+
+//           |   |        |       |
+//           |   |        |       | Has Candidates
+//           |   |        |       v
+//           |   |        |     +-----------------+
+//           |   |        |     |    WAITING FOR  |
+//           |   +--------|---> |    EVACUATIONS  |
```
ysramakrishna, line 417: This transition has arrows at both ends (IDLE and WAITING FOR EVACUATIONS).

ysramakrishna (commented):
Left a few comments, and going over all of the changes again to make sure I understand the changes completely again before I sign off on the review.

ysramakrishna (approved):
Changes LGTM, modulo a few minor comments in previous set of comments.

src/hotspot/share/gc/shenandoah/shenandoahDegeneratedGC.cpp
earthling-amzn: With this change, if the global cycle degenerates after final mark there could now be regions waiting to be coalesced and filled, so the old gc state is not idle.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
earthling-amzn: That is intentional, I shall also add an arrow at both ends for IDLE and WAITING FOR FILL. Global cycles do not progress the old generation through marking/bootstrapping so they are allowed to move the old generation from IDLE to one of these states at the end of global mark.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.hpp
earthling-amzn: I agree, one (good) name for both is best.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
ysramakrishna: I see, so a bidirectional arrow?

It would of course be ideal (in the fullness of time) if there were events labeling the transitions, even if this required a table for the event alphabet.

PR #307, opened 2023-08-16
8309639: GenShen: Regression in LRU cache benchmark

earthling-amzn:
Shenandoah's adaptive heuristic begins in a state where it is "learning" the application. In this state, it triggers collections somewhat more aggressively in an effort to "learn" the allocation rate and cycle times for the application.

Shenandoah also has the ability to "skip" evacuation phases when marking finds sufficient garbage in regions with no live objects. We call these "abbreviated" cycles. During the course of development for the generational mode, we found that such cycles tend to decrease the average cycle time tracked by the heuristic. This in turn leads to the heuristic triggering a cycle "too late" - because it expects the cycle to take less time than a normal (i.e., not abbreviated) cycle requires. To address this, we modified the heuristic to not include abbreviated cycle times in the average cycle time.

Unfortunately, this change also made the heuristic not count abbreviated cycles as "learning" cycles. Ordinarily, this is not a concern because the application will usually run plenty of "normal" cycles. However, we've discovered that the certain benchmarks and jtreg tests only run abbreviated cycles. This causes the heuristic to stay in the "learning" state for the duration of the benchmark. In this state, the GC is run more aggressively and performance suffers (15-20%).

---------
### Progress

### Issue
 * [JDK-8309639](https://bugs.openjdk.org/browse/JDK-8309639): GenShen: Regression in LRU cache benchmark (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/307/head:pull/307` \
`$ git checkout pull/307`

`$ git checkout pull/307` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/307/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 307`

`$ git pr show -t 307`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/307.diff">https://git.openjdk.org/shenandoah/pull/307.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/307#issuecomment-1681399994)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
@@ -234,13 +234,13 @@ void ShenandoahHeuristics::adjust_penalty(intx step) {
 void ShenandoahHeuristics::record_success_concurrent(bool abbreviated) {
   _degenerated_cycles_in_a_row = 0;
   _successful_cycles_in_a_row++;
-
-  if (!(abbreviated && ShenandoahAdaptiveIgnoreShortCycles)) {
-    _gc_cycle_time_history->add(elapsed_cycle_time());
-    _gc_times_learned++;
-  }
+  _gc_times_learned++;
 
   adjust_penalty(Concurrent_Adjust);
+
+  if (_gc_times_learned <= ShenandoahLearningSteps || !(abbreviated && ShenandoahAdaptiveIgnoreShortCycles)) {
+    _gc_cycle_time_history->add(elapsed_cycle_time());
+  }
 }
 
 void ShenandoahHeuristics::record_success_degenerated() {


rkennke:
Should this go straight to upstream jdk repo?

earthling-amzn:
> Should this go straight to upstream jdk repo?

No, this is only in the shenandoah repo. I will change the tag in the title to `GenShen`.

kdnilsen (approved):
This looks like a good compromise.  Since this new code isn't in the "isolation audit", I don't want to lose track of the change.  Should we make this conditional upon is_generational, and revert the non-generational mode to original behavior?

src/hotspot/share/gc/shenandoah/heuristics/shenandoahHeuristics.cpp
```diff
@@ -234,13 +234,13 @@ void ShenandoahHeuristics::adjust_penalty(intx step) {
 void ShenandoahHeuristics::record_success_concurrent(bool abbreviated) {
   _degenerated_cycles_in_a_row = 0;
   _successful_cycles_in_a_row++;
+  _gc_times_learned++;
 
-  if (!(abbreviated && ShenandoahAdaptiveIgnoreShortCycles)) {
+  adjust_penalty(Concurrent_Adjust);
+
+  if (_gc_times_learned <= ShenandoahLearningSteps || !(abbreviated && ShenandoahAdaptiveIgnoreShortCycles)) {
```
ysramakrishna, line 241: Can you leave a documentation comment here recording the intent in English?

As I read it, you include the time in the recorded history if any of the following hold:
1. we haven't reached the required number of learning steps, or
2. we are not an abbreviated cycle, or
3. we haven't been asked to ignore short cycles

To the extent that this is so, I think the test would read easier if you de-Morganized the conditional into:

```
if (_gc_times_learned <= ShenandoahLearningSteps || !abbreviated || !ShenandoahAdaptiveIgnoreShortCycles)) {
```

Anyway, something for us to remember why we have this condition would be useful as well; your PR explanation is excellent. May be a condensed form of that would be appropriate as a documentation comment here?

earthling-amzn:
> Should we make this conditional upon is_generational, and revert the non-generational mode to original behavior?

I'm more inclined to test if this feature (i.e., heuristic ignores abbreviated cycles) improves performance on the upstream branch. I will test for a measurable benefit for this feature. For now, I think we should fix this bug. If we can show the performance benefit of this feature, I think we should upstream it. Otherwise, we should just back it out altogether.

ysramakrishna (approved):
LGTM. In addition to the code documentation remark, can I also request that the observed change in performance with your favorite benchmark or test be included either here in the PR or in the JBS ticket?

I am just curious how the learning data is used. If that is entirely abbreviated cycles, say, wouldn't you have the same issue? And skipping abbreviated cycle counting after the learning steps would leave you in the same state of under-predicting when a normal cycle does occur. Does the small number of recorded samples then save you because of the potentially larger padding because of the potentially larger standard deviation than if you counted all of the abbreviated cycles?

Adversarial workload scenarios are easy to imagine, but may be real applications don't behave like I was thinking of adversarial scenarios, so the performance data that you share would be useful to allay such fears.

Allow me to do some thinking out loud here. May be there's a flaw in my thinking below.

In general, my instinct would be that including abbreviated cycles would be fine; if they are frequent then the infrequent normal cycles increase the standard deviation (even if it might subsequently decay). Conversely, if abbreviated cycles are infrequent then we should be fine. May be thinking of the benchmarks/tests that gained from the change in heuristic would allow us to determine why the scenarios in which this helped and why. In particular, was the load presented by the benchmark very spiky, and this allowed us to tread in the direction of being conservative. Still, why doesn't a degenerate collection swing us back into the right spot? Does it do that but do we swing back again to complacency? What was the difference between the durations for abbreviated and normal cycles in the cases where we saw this improvement? I'd have expected the standard deviation to have provided a buffer/cushion.

Thanks!

PR #310, opened 2023-08-24
8314972: GenShen: promote_in_place needs to prepare remembered set before it enables old allocations within region

kdnilsen:
An error in the original implementation left us vulnerable to a race between a worker thread which is promoting a region in place and another worker or mutator thread who is allocating memory for old objects within the same region.

This PR prevents this race by requiring the worker thread that promotes a region in place to finish the preparation of the region's remembered set before it makes the region available to other threads for allocation of old-gen memory.

---------
### Progress

### Issue
 * [JDK-8314972](https://bugs.openjdk.org/browse/JDK-8314972): GenShen: promote_in_place needs to prepare remembered set before it enables old allocations within region (**Bug** - P2)

### Reviewers
 * [Y. Srinivas Ramakrishna](https://openjdk.org/census#ysr) (@ysramakrishna - Committer) ⚠️ Review applies to [96ca2bea](https://git.openjdk.org/shenandoah/pull/310/files/96ca2bea0cb078d66f75683c484c75e2d8f65f35)
 * [William Kemper](https://openjdk.org/census#wkemper) (@earthling-amzn - Committer) ⚠️ Review applies to [96ca2bea](https://git.openjdk.org/shenandoah/pull/310/files/96ca2bea0cb078d66f75683c484c75e2d8f65f35)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/310/head:pull/310` \
`$ git checkout pull/310`

`$ git checkout pull/310` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/310/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 310`

`$ git pr show -t 310`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/310.diff">https://git.openjdk.org/shenandoah/pull/310.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/310#issuecomment-1693622806)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
@@ -1017,12 +1017,17 @@ void ShenandoahHeapRegion::promote_in_place() {
 
   assert(get_top_before_promote() == tams, "Cannot promote regions in place if top has advanced beyond TAMS");
 
-  // Since this region may have served previously as OLD, it may hold obsolete object range info.
+  // Rebuild the remembered set information and mark the entire range as DIRTY.  We do NOT scan the content of this
+  // range to determine which cards need to be DIRTY.  That would force us to scan the region twice, once now, and
+  // once during the subsequent remembered set scan.  Instead, we blindly (conservatively) mark everything as DIRTY
+  // now and then sort out the CLEAN pages during the next remembered set scan.
+  //
+  // Rebuilding the remembered set consists of clearing all object registrations (reset_object_range()) here,
+  // then registering every live object and every coalesced range of free objects in the loop that follows.
   heap->card_scan()->reset_object_range(bottom(), end());
   heap->card_scan()->mark_range_as_dirty(bottom(), get_top_before_promote() - bottom());
 
-  // TODO: use an existing coalesce-and-fill function rather than
-  // replicating the code here.
+  // TODO: use an existing coalesce-and-fill function rather than replicating the code here.
   HeapWord* obj_addr = bottom();
   while (obj_addr < tams) {
     oop obj = cast_to_oop(obj_addr);
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
@@ -1015,6 +1015,35 @@ void ShenandoahHeapRegion::promote_in_place() {
   ShenandoahYoungGeneration* young_gen = heap->young_generation();
   size_t region_size_bytes = ShenandoahHeapRegion::region_size_bytes();
 
+  assert(top() == tams, "Cannot promote regions in place if top has advanced beyond TAMS");
+
+  // Since this region may have served previously as OLD, it may hold obsolete object range info.
+  heap->card_scan()->reset_object_range(bottom(), end());
+  heap->card_scan()->mark_range_as_dirty(bottom(), top() - bottom());
+
+  // TODO: use an existing coalesce-and-fill function rather than
+  // replicating the code here.
+  HeapWord* obj_addr = bottom();
+  while (obj_addr < tams) {
+    oop obj = cast_to_oop(obj_addr);
+    if (marking_context->is_marked(obj)) {
+      assert(obj->klass() != NULL, "klass should not be NULL");
+      // This thread is responsible for registering all objects in this region.  No need for lock.
+      heap->card_scan()->register_object_without_lock(obj_addr);
+      obj_addr += obj->size();
+    } else {
+      HeapWord* next_marked_obj = marking_context->get_next_marked_addr(obj_addr, tams);
+      assert(next_marked_obj <= tams, "next marked object cannot exceed tams");
+      size_t fill_size = next_marked_obj - obj_addr;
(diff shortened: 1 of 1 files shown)


kdnilsen:
I've got some debugging to complete on this PR.

kdnilsen:
Failed pre-submit tests are due to:

1. riscv64 not currently supported by GenShen
2. https://bugs.openjdk.org/browse/JDK-8311843

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -1015,6 +1015,35 @@ void ShenandoahHeapRegion::promote_in_place() {
   ShenandoahYoungGeneration* young_gen = heap->young_generation();
   size_t region_size_bytes = ShenandoahHeapRegion::region_size_bytes();
 
+  assert(get_top_before_promote() == tams, "Cannot promote regions in place if top has advanced beyond TAMS");
+
+  // Since this region may have served previously as OLD, it may hold obsolete object range info.
```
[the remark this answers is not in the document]
ysramakrishna: May be modify the comment to state something like:
// Clear any obsolete object range information; we'll register the live objects in the loop below,
// as well as registering and stamping free block fillers over dead objects.
// We also mark the entire range of the card table dirty; it'll be updated when we scan it next time
// for young pointers.

ysramakrishna (approved):
LGTM. 🚢

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -1015,6 +1015,35 @@ void ShenandoahHeapRegion::promote_in_place() {
...
+      // This thread is responsible for registering all objects in this region.  No need for lock.
+      heap->card_scan()->register_object_without_lock(obj_addr);
+      obj_addr += obj->size();
+    } else {
+      HeapWord* next_marked_obj = marking_context->get_next_marked_addr(obj_addr, tams);
+      assert(next_marked_obj <= tams, "next marked object cannot exceed tams");
+      size_t fill_size = next_marked_obj - obj_addr;
+      assert(fill_size >= ShenandoahHeap::min_fill_size(), "previously allocated objects known to be larger than min_size");
+      ShenandoahHeap::fill_with_object(obj_addr, fill_size);
+      heap->card_scan()->register_object_without_lock(obj_addr);
```
ysramakrishna, line 1040: On second thoughts, do we really need to register the dead objects? Does it help at all, since the preceding and succeeding live objects will be registered anyway. Not wrong, but I do wonder if it helps at all or not. Probably not a big deal either way since dead objects are likely very few in regions promoted in place.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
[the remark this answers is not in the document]
kdnilsen: We don't clear object registrations when an old region is freed because often times, the free region may next be put into service as a young-region, where the object registrations are "don't care" values.  I'll make the adjustment to the comment.

earthling-amzn (approved):

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
kdnilsen: Existing code (and some existing assertions in ShenandoahVerifier) expect the filler objects to be registered.  You make a good point though.  We don't really need these to be registered.  If the first or only object in a card range is a filler, remembered set scanning can just ignore that.  The remembered set scanner wants to find the first non-filler object in the range.  We might want to address this in a separate ticket later, if we decide to keep the current remembered set representation longer term.

kdnilsen: Note that the cost of registering filler objects is minimal because there are never multiple contiguous filler objects residing within memory.  We always coalesce them together.  At most, the cost during remembered set scanning is that we look at one extra filler object per DIRTY card.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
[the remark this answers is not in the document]
ysramakrishna: In any run, the number of clearings is basically the number old region births if cleared at promotion or allocation of an old generation, or deaths if cleared at freeing of an old generation. In a non-terminating run,  the rate of births equals rate of deaths. So the difference is just _when_ we clear, not how much. The amount of clearing work in the long run is no different in either case. So the difference is cache pollution, mutator latency, and code maintenance. I'm guessing there's really no difference, since it all happens on concurrent GC threads. So code maintenance would probably trump everything, and thinking some more about it, I am guessing there's not much difference in that either in when the clearing is done.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
ysramakrishna: I agree, I don't think it helps one way or other. In fact, the current card scanning code looks for the last object on the _previous_ card from which to start walking objects to scan. In the case that the filler object is entirely on a card, this does not make any difference. However, in the case that the filler object is straddling a card, registering it shortens the walk by one object, the one preceding the filler. I think any performance difference is a wash, and I'd just scratch my thought above.

PR #311, opened 2023-08-25
8311843: GenShen: assertion failed "Old generation affiliated regions must be less than capacity"

earthling-amzn:
There are a few fixes here:

During a Full GC, the compaction process is not constrained by the maximum capacity for young and old generations. We must, therefore, reestablish these constraints after the Full GC completes by _forcing_ the transfer of regions to old, if necessary.

When a region is "flipped" from the Mutator's free set to the Collector's, we cannot _force_ the transfer of the region because the young generation may not have any free regions.

A logic error in the loop which rebuilds the free set could prevent transferring free regions from the mutator to the collector reserves.

Some unused methods and fields were removed.

---------
### Progress

### Issue
 * [JDK-8311843](https://bugs.openjdk.org/browse/JDK-8311843): GenShen: assertion failed "Old generation affiliated regions must be less than capacity" (**Bug** - P2)

### Reviewers
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [3af6309e](https://git.openjdk.org/shenandoah/pull/311/files/3af6309ed1dae15fcceac04712eca8c6fc7fe080)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/311/head:pull/311` \
`$ git checkout pull/311`

`$ git checkout pull/311` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/311/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 311`

`$ git pr show -t 311`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/311.diff">https://git.openjdk.org/shenandoah/pull/311.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/311#issuecomment-1693733974)


src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -1010,8 +1010,12 @@ void ShenandoahFreeSet::flip_to_old_gc(ShenandoahHeapRegion* r) {
   size_t region_capacity = alloc_capacity(r);
   _free_sets.move_to_set(idx, OldCollector, region_capacity);
   _free_sets.assert_bounds();
-  _heap->generation_sizer()->force_transfer_to_old(1);
+
```
kdnilsen, line 1013: IIUC, the new code will fail to transfer if the transfer makes YOUNG smaller than min_capacity or makes OLD larger than max_capacity.  It will never fail because unaffiliated_regions < 1.

I have two concerns with the code as written:
1. I don't think we should flip the region if we don't flip the capacity.  That will create future assertion failures.
2. If we fail to flip_to_old(), we need to return a status code to so indicate.  The caller needs to know that the requested flip cannot be achieved and thus the precipitating "allocation request" must fail.

Alternatively, we could adopt the convention that min/max generation sizes are "preferences" but are not absolutely enforced.  Maybe it is better to bend the rules to allow pause-free operation rather than forcing an old evacuation failure that would require a Full GC.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -1225,15 +1229,15 @@ void ShenandoahFreeSet::rebuild(size_t young_cset_regions, size_t old_cset_regio
       old_reserve = old_available;
     }
   }
-  if (old_reserve > _free_sets.capacity_of(OldCollector)) {
```
kdnilsen, line 1228: Good catch here.  Sorry for the code as originally written.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: If we adopt this convention, we should get rid of any asserts that endeavor to enforce max/min generation capacities  (or change these to enforce that capacity never shrinks below 0 and never grows above heap->capacity())

src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
```diff
@@ -1535,7 +1535,7 @@ void ShenandoahFullGC::phase5_epilog() {
         heap->generation_sizer()->transfer_to_young(excess_old_regions);
       } else if (old_capacity < old_usage) {
         size_t old_regions_deficit = (old_usage - old_capacity) / ShenandoahHeapRegion::region_size_bytes();
-        heap->generation_sizer()->transfer_to_old(old_regions_deficit);
+        heap->generation_sizer()->force_transfer_to_old(old_regions_deficit);
       }
```
kdnilsen, line 1539: So here, we may end up violating max/min sizes of young generation.  I'm inclined to allow this if that is required to make the just-completed full gc valid.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
```diff
@@ -979,15 +978,3 @@ void ShenandoahGeneration::record_success_degenerated() {
   heuristics()->record_success_degenerated();
   ShenandoahHeap::heap()->shenandoah_policy()->record_success_degenerated();
 }
-
```
kdnilsen, line 982: I gather that this code was redundant/not needed.  I'm not sure removal of this code was highlighted in the commit comment.  Maybe we should also mention some code cleanup regarding generation collection times.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
earthling-amzn: `transfer_to_old` does check if young has enough affiliated regions to transfer:
```
  if (young_gen->free_unaffiliated_regions() < regions) {
    return false;
```

I think the original error here was caused by the Mutator free set holding many more regions than were available to the young generation (the issue with `reserve_regions`). I don't expect that to happen again, but I did add a log message if the transfer fails.

We are already somewhat loose when it comes to enforcing generational capacities. Are you proposing we have `shFreeSet` be the sole arbiter of capacity? That sounds like a good simplification - also sounds like a different PR?

kdnilsen: I understand that it checks, but the check should never fail, because the two asserts that precede the call (and did not need to be removed) confirm that the region to be flipped is in the Mutator free set (i.e. is part of young capacity) and is entirely empty (so must be unaffiliated).

I need to study the reserve_regions problem more.  I saw that you refactored a bit.  I didn't catch that you were fixing a bug.

kdnilsen: Agree that changing shFreeSet to sole arbiter is probably a separate PR, that might be a good improvement.

src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
earthling-amzn: We may violate a configured constraint, but the usage accounting should be correct now.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
earthling-amzn: Added this cleanup to the PR description. Yes, this code was a vestige of the old mmu based resizer.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -1253,27 +1257,39 @@ void ShenandoahFreeSet::reserve_regions(size_t to_reserve, size_t to_reserve_old
...
+    if (move_to_young) {
+      // Note: In a previous implementation, regions were only placed into the survivor space (collector_is_free) if
+      // they were entirely empty.  I'm not sure I understand the rationale for that.  That alternative behavior would
+      // tend to mix survivor objects with ephemeral objects, making it more difficult to reclaim the memory for the
+      // ephemeral objects.  It also delays aging of regions, causing promotion in place to be delayed.
+      _free_sets.move_to_set(idx, Collector, ac);
+      log_debug(gc)("  Shifting region " SIZE_FORMAT " from mutator_free to collector_free", idx);
+    }
   }
 }
```
kdnilsen, line 1295: When we rebuild, we compute target amounts of memory to_reserve_old and to_reserve_young.  We may not satisfy our targets if the Mutator.is_free memory is "too fragmented".  This would be rare, I think, but maybe deserving of a log_gc(info) message whenever we get to the end of rebuild and we are short.  I think the rest of the implementation is supposed to be robust to this scenario.  (i.e. if we did not reserve as much as we wanted to, then we will target a smaller collection set (in old or young)).

src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
kdnilsen: good.  and i assume that the configured constraint violation no longer manifests as a failed assertion.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: Given that full-gc is allowed to violate the min/max constraints on generation sizes, I'm inclined to allow the same here, and revert this to force_transfer_to_old().  If that doesn't work, maybe a comment to explain why.  Because I'm not understanding this entirely.

kdnilsen (approved):

earthling-amzn:
I'll try letting `ShenandoahFreeSet::flip_to_old_gc` force transferring the region to old. I think it only caused problems because the free regions in the mutator set had not been properly distributed when the free sets were `rebuilt` (that is, it tried to force a young generation with no unaffiliated regions to give one up).

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
earthling-amzn: Yes, there was a logic error in `rebuild` that caused the loop to exit well before any regions were transferred to the collector's reserves. This left it in a state where `capacity_of(Mutator)` was well higher than it ought to have been (i.e., when `young->available() == 0`).

PR #313, opened 2023-08-28
8315247: GenShen: Condition calls to post-write barrier code generation by a flag

ysramakrishna:
Protect the card barrier code generation by a new global `ShenandoahCardBarrier`, which is disabled by default and enabled for GenShen (generational mode Shenandoah) only. If the user forces the barrier flag to a value inconsistent with mode, we exit with an appropriate error message.

The intention of the change is two-fold:
1. Make the card-barrier code in support of GenShen more idiomatic along the lines of existing barrier-gen code in Shenandoah
2. Reduce and simplify to the extent possible the impact of changes from GenShen on shared/legacy Shenandoah code, so as to make reviews/audits of shared/legacy code a bit easier

The changes have been made for x86, aarch, and ppc, and tested as follows:
1. I have tested on x86 on a cloud host running jtreg1-4 and SPECjbb.
3. A round of aarch64 and x86 pipeline tests have run (but ran into existing assertion that William is fixing).
4. Many thanks to Martin Doerr (SAP) for putting this through a "quick spin" on PPC
5. RISC-V & S390 don't currently support generational mode Shenandoah.

---------
### Progress

### Issue
 * [JDK-8315247](https://bugs.openjdk.org/browse/JDK-8315247): GenShen: Condition calls to post-write barrier code generation by a flag (**Task** - P3)

### Reviewers
 * [William Kemper](https://openjdk.org/census#wkemper) (@earthling-amzn - Committer) ⚠️ Review applies to [71bd4973](https://git.openjdk.org/shenandoah/pull/313/files/71bd49736f18dde3c13bd8a078ed3c01c8295b20)
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [71bd4973](https://git.openjdk.org/shenandoah/pull/313/files/71bd49736f18dde3c13bd8a078ed3c01c8295b20)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/313/head:pull/313` \
`$ git checkout pull/313`

`$ git checkout pull/313` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/313/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 313`

`$ git pr show -t 313`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/313.diff">https://git.openjdk.org/shenandoah/pull/313.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/313#issuecomment-1697960049)


earthling-amzn (changes requested):
Minor nit - can we have a consistent, more helpful message for the assertions?

```
  assert(ShenandoahHeap::heap()->mode()->is_generational(), "Not needed");
```
Maybe:
```
  assert(ShenandoahHeap::heap()->mode()->is_generational(), "Expected generational mode here");
```

src/hotspot/cpu/x86/gc/shenandoah/shenandoahBarrierSetAssembler_x86.cpp
```diff
@@ -208,6 +209,12 @@ void ShenandoahBarrierSetAssembler::arraycopy_prologue(MacroAssembler* masm, Dec
 
 void ShenandoahBarrierSetAssembler::arraycopy_epilogue(MacroAssembler* masm, DecoratorSet decorators, BasicType type,
                                                        Register src, Register dst, Register count) {
+
+  if (!ShenandoahCardBarrier) {
```
kdnilsen, line 213: Why wouldn't we handle this the same as other barriers.  assert(ShenandoahCardBarrier) here and check for ShenandoahCardBarrier before calling arraycopy_epilogue()?

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
```diff
@@ -518,6 +518,9 @@
   product(bool, ShenandoahIUBarrier, false, DIAGNOSTIC,                     \
           "Turn on/off I-U barriers barriers in Shenandoah")                \
                                                                             \
+  product(bool, ShenandoahCardBarrier, false, DIAGNOSTIC,                   \
+          "Turn on/off card-marking post-write barrier in Shenandoah")      \
+                                                                            \
```
kdnilsen, line 523: I was thinking this would not be visible on the command line.  If we make it visible, then we would need to assure consistency with mode == generational.

My thought was we would just set this global variable in, for example, ShenandoahArguments::initialize() if !strcmp(ShenandoahGCMode, "generational")

kdnilsen (commented):
This looks pretty good.  Have you sent it through the test pipeline?

If we have this ShenandoahCardBarrier flag, I'd be inclined to use it everywhere we currently have the query heap->mode()->is_generational().  It ought to be more efficient.  Before you make that global change, maybe we should confirm that William (and other interested reviewers) like that approach.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
earthling-amzn: This approach has parity with the other barrier flags. `shenandoahGenerationalMode.cpp` asserts that the flag is on. 

It shouldn't be strictly necessary with the default being `false`, but we may want to have `shenandoahPassiveMode.cpp` explicitly turn this flag off as it does the other barrier flags.

kdnilsen: If we make this visible on the command line, we will want to make sure it is consistent with mode->is_generational().  Right?

kdnilsen: (or you could say, if they turn off ShenandoahCardBarrier, that would mean we have no remembered set and always scan all of old-gen.  but that would be a lot more work for us, and of no value, I think...)

kdnilsen: maybe just fatal out in ShenandoahArguments if mode->is_generational() and !ShenandoahCardBarrier.  So the user has to specify ShenandoahCardBarrier if they specify ShenandoahGCMode == generational.

src/hotspot/cpu/aarch64/gc/shenandoah/shenandoahBarrierSetAssembler_aarch64.cpp
```diff
@@ -385,9 +385,9 @@ void ShenandoahBarrierSetAssembler::load_at(MacroAssembler* masm, DecoratorSet d
 }
 
 void ShenandoahBarrierSetAssembler::store_check(MacroAssembler* masm, Register obj) {
-  if (!ShenandoahHeap::heap()->mode()->is_generational()) {
-    return;
-  }
+  // This will need to be changed if we ever use a card-marking post-barrier w/a non-generational
+  // Shenandoah in the future.
+  assert(ShenandoahHeap::heap()->mode()->is_generational(), "Only use case for Shenandoah today");
```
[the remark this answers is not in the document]
ysramakrishna: Done.

src/hotspot/cpu/ppc/gc/shenandoah/shenandoahBarrierSetAssembler_ppc.cpp
```diff
@@ -597,9 +597,7 @@ void ShenandoahBarrierSetAssembler::load_at(
 }
 
 void ShenandoahBarrierSetAssembler::store_check(MacroAssembler* masm, Register base, RegisterOrConstant ind_or_offs, Register tmp) {
-  if (!ShenandoahHeap::heap()->mode()->is_generational()) {
-    return;
-  }
+  assert(ShenandoahCardBarrier && ShenandoahHeap::heap()->mode()->is_generational(), "Error");)
```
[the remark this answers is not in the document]
ysramakrishna: Done

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
ysramakrishna: I'm going to leave it visible on command-line like the other `Shenandoah*Barrier` flags, but force it off in all modes except generational where we force it on. Since these aren't manageable flags,, jcmd should not be able to enable them after start up either. 

If the concern is that we would be silently ignoring user request to explicitly enable it in non-generational mode, or disable it in generational mode, we can log a warning upon such an attempt and still do the right thing.

Let me test these various flag combos and ensure that they do the right thing. I'll update the PR once finished w/the changes and testing.

src/hotspot/cpu/x86/gc/shenandoah/shenandoahBarrierSetAssembler_x86.cpp
ysramakrishna: Today we call arraycopy_epilogue() unconditionally from common code (such as stub generators) irrespective of collector. We'll probably want a better barrier set abstraction across all collectors to do what you suggest in the future. For now, I'll just massage this to conform to the other platforms where we test and branch on the flag here (rather than test and return at the top, or do the job in the body that follows).

src/hotspot/cpu/aarch64/gc/shenandoah/shenandoahBarrierSetAssembler_aarch64.cpp
```diff
@@ -629,9 +631,9 @@ void ShenandoahBarrierSetAssembler::cmpxchg_oop(MacroAssembler* masm,
 
 void ShenandoahBarrierSetAssembler::gen_write_ref_array_post_barrier(MacroAssembler* masm, DecoratorSet decorators,
                                                                      Register start, Register count, Register scratch, RegSet saved_regs) {
-  if (!ShenandoahHeap::heap()->mode()->is_generational()) {
-    return;
-  }
+  // This will need to be changed if we ever use a card-marking post-barrier w/a non-generational
+  // Shenandoah in the future.
+  assert(ShenandoahHeap::heap()->mode()->is_generational(), "Only use case for Shenandoah today");
```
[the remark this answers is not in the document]
ysramakrishna: Done

src/hotspot/cpu/ppc/gc/shenandoah/shenandoahBarrierSetAssembler_ppc.cpp
```diff
@@ -795,9 +793,7 @@ void ShenandoahBarrierSetAssembler::cmpxchg_oop(MacroAssembler *masm, Register b
 
 void ShenandoahBarrierSetAssembler::gen_write_ref_array_post_barrier(MacroAssembler* masm, DecoratorSet decorators,
                                                                      Register addr, Register count, Register preserve) {
-  if (!ShenandoahHeap::heap()->mode()->is_generational()) {
-    return;
-  }
+  assert(ShenandoahCardBarrier && ShenandoahHeap::heap()->mode()->is_generational(), "Error");
```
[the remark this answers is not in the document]
ysramakrishna: Done.

ysramakrishna:
> Minor nit - can we have a consistent, more helpful message for the assertions?
> 
> ```
>   assert(ShenandoahHeap::heap()->mode()->is_generational(), "Not needed");
> ```
> 
> Maybe:
> 
> ```
>   assert(ShenandoahHeap::heap()->mode()->is_generational(), "Expected generational mode here");
> ```

Following Kelvin's suggestion, I removed any reference to generational mode in the barrier/generayion code itself, and instead just assert that the `ShenandoahCardBarrier` flag should be set. I am going to look for a suitable place during initialization of non-generational mode to assert that it isn't set. We do check and bail at generational mode initialization if the flag isn't set.

I used the message `Did you mean to enable ShenandoahCardBarrier?` if it isn't enabled, but I could replace it with `Expecting ShenandoahCardBarrier set here` in case that reads better in the error message, or `ShenandoahCardBarrier should be set`.

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
kdnilsen: Maybe the comment can say: "ShenandoahCardBarrier" is true by default iff ShenandoahGCMode is generational.

Otherwise, I think we would need to change all of the existing test configurations that have generational mode to add -XX:ShenandoahCardBarrier=true

ysramakrishna: I updated with an attempt to address the comments above. Take a look and see how it looks.

I'll also take this out of draft form to a regular PR.

ysramakrishna: So we'll now exit with an appropriate error message if the user changes `ShenandoahCardBarrier` to a value that conflicts with the `ShenandoahGCMode` they selected:

src/hotspot/cpu/x86/gc/shenandoah/shenandoahBarrierSetAssembler_x86.cpp
ysramakrishna: Updated.

ysramakrishna:
> This looks pretty good. Have you sent it through the test pipeline?
> 
> If we have this ShenandoahCardBarrier flag, I'd be inclined to use it everywhere we currently have the query heap->mode()->is_generational(). It ought to be more efficient. Before you make that global change, maybe we should confirm that William (and other interested reviewers) like that approach.

Did a spin through test pipeline and will run another following the latest changes. All looks good from what I can tell from tests, including jtreg's & SPECjbb as well.

Opened official PR from draft and will link JBS ticket momentarily.

kdnilsen (approved):

src/hotspot/share/gc/shenandoah/shenandoah_globals.hpp
ysramakrishna: Elaborated documentation so users aren't confused when they try and toggle settings to those that might conflict, and the JVM exits with a somewhat terse error message such as above.

PR #314, opened 2023-08-28
8315044: GenShen: Verifier detects clean card should be dirty

kdnilsen:
When a Reference object is newly discovered, it is placed onto the worker's thread-local discovered list.  This sometimes results in a reference from an old object to a young object, requiring that the remembered set card-table entry be marked as dirty.  This patch causes the marking to be performed.

---------
### Progress

### Issue
 * [JDK-8315044](https://bugs.openjdk.org/browse/JDK-8315044): GenShen: Verifier detects clean card should be dirty (**Bug** - P2)

### Reviewers
 * [Y. Srinivas Ramakrishna](https://openjdk.org/census#ysr) (@ysramakrishna - Committer) ⚠️ Review applies to [2618ff3c](https://git.openjdk.org/shenandoah/pull/314/files/2618ff3c9f2e2c5d67997e3f59c35cd1a5b6c0a2)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/314/head:pull/314` \
`$ git checkout pull/314`

`$ git checkout pull/314` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/314/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 314`

`$ git pr show -t 314`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/314.diff">https://git.openjdk.org/shenandoah/pull/314.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/314#issuecomment-1697862818)


src/hotspot/share/gc/shenandoah/shenandoahReferenceProcessor.cpp
```diff
@@ -372,6 +375,20 @@ bool ShenandoahReferenceProcessor::discover(oop reference, ReferenceType type, u
...
+    // We successfully set this reference object's next pointer to discovered_head.  This marks reference as discovered.
+    // If reference_cas_discovered fails, that means some other worker thread took credit for discovery of this reference,
+    // and that other thread will place reference on its discovered list, so I can ignore reference.
+
+    // In case we have created an interesting pointer, mark the remembered set card as dirty.
+    ShenandoahHeap* heap = ShenandoahHeap::heap();
+    if (ShenandoahCardBarrier) {
+      T* addr = reinterpret_cast<T*>(java_lang_ref_Reference::discovered_addr_raw(reference));
+      if (heap->is_in_old(addr) && heap->is_in_young(discovered_head)) {
+        heap->mark_card_as_dirty(addr);
```
ysramakrishna, line 387: Can you instead call the existing `card_mark_barrier` method?
At that time, could also look at the existing uses of that method and refactor them to use the `ShenandoahCardBarrier` test like you did above, and elide the generational check in the body of the method itself, replacing it with an assertion (if you want, that the heap is generational).

(PS: I was also thinking that in the fullness of time it might be worth doing an audit of the uses of `cas` and `set_oop` and making sure they would call the card-marking code instead of exposing the functionality at the call-sites like done here. Callers that want to avoid card-marking would need to call the respective `_raw` versions then. We should makr sure to conform to naming conventions for those methods that are consistent with those used in other parts of jvm/gc code. This can be done separately and might affect legacy Shenandoah code as well. Just leaving this comment here so as not to lose the thought, but I doubt we want to do that locally here in this PR.)

ysramakrishna (commented):
changes look good; minor suggestion on some potential refactor/consolidation with existing card marking code.

src/hotspot/share/gc/shenandoah/shenandoahReferenceProcessor.cpp
kdnilsen: Made this change and am testing result.

I agree that audit/refactoring in the future would be a good idea.

kdnilsen: I've made your suggested changes and am sending through the test pipeline again.

I agree that it would be good to eventually audit/refactor to have more consistent method names and conventions.

src/hotspot/share/gc/shenandoah/shenandoahReferenceProcessor.cpp
```diff
@@ -544,10 +565,29 @@ void ShenandoahReferenceProcessor::process_references(ShenandoahPhaseTimings::Ph
...
+    *reinterpret_cast<narrowOop*>(_pending_list_tail) = CompressedOops::encode(former_head_of_global_list);
   } else {
-    *reinterpret_cast<oop*>(_pending_list_tail) = Universe::swap_reference_pending_list(_pending_list);
+    *reinterpret_cast<oop*>(_pending_list_tail) = former_head_of_global_list;
+  }
+  ShenandoahHeap* heap = ShenandoahHeap::heap();
+  if (ShenandoahCardBarrier) {
+    if (heap->is_in_old(_pending_list_tail) && heap->is_in_young(former_head_of_global_list)) {
+      heap->mark_card_as_dirty(_pending_list_tail);
+    }
```
ysramakrishna, line 590: Can we just replace this section with `set_oop_field(_pending_list_tail, former_head_of_global_list)` ?

ysramakrishna (approved):
One more change, I think? See comment in-line.

May be there is an issue with it though that I am not seeing which perhaps you might have tried already, but doesn't work? 

Rest looks good.

src/hotspot/share/gc/shenandoah/shenandoahReferenceProcessor.cpp
kdnilsen: Have made this improvement, and am retesting.

earthling-amzn (approved):

PR #317, opened 2023-09-01
JDK-8315478: GenShen: Tolerate round-off errors in preselected promotion budget

kdnilsen:
An assert failure identified an error in previous implementation.  See JBS ticket for description of issue and reproducer.

---------
### Progress

### Issue
 * [JDK-8315478](https://bugs.openjdk.org/browse/JDK-8315478): GenShen: Tolerate round-off errors in preselected promotion budget (**Sub-task** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/317/head:pull/317` \
`$ git checkout pull/317`

`$ git checkout pull/317` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/317/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 317`

`$ git pr show -t 317`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/317.diff">https://git.openjdk.org/shenandoah/pull/317.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/317#issuecomment-1703028347)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
@@ -394,6 +394,17 @@ void ShenandoahGeneration::adjust_evacuation_budgets(ShenandoahHeap* heap, Shena
   // and promotion reserves.  Try shrinking OLD now in case that gives us a bit more runway for mutator allocations during
   // evac and update phases.
   size_t old_consumed = old_evacuated_committed + young_advance_promoted_reserve_used;
+
+  if (old_available < old_consumed) {
+    // This can happen due to round-off errors when adding the results of truncated integer arithmetic.
+    // We've already truncated old_evacuated_committed.  Truncate young_advance_promoted_reserve_used here.
+    assert(young_advance_promoted_reserve_used <= (33 * (old_available - old_evacuated_committed)) / 32,
+           "Round-off errors should be less than 3.125%%, committed: " SIZE_FORMAT ", reserved: " SIZE_FORMAT,
+           young_advance_promoted_reserve_used, old_available - old_evacuated_committed);
+    young_advance_promoted_reserve_used = old_available - old_evacuated_committed;
+    old_consumed = old_evacuated_committed + young_advance_promoted_reserve_used;
+  }
+
   assert(old_available >= old_consumed, "Cannot consume (" SIZE_FORMAT ") more than is available (" SIZE_FORMAT ")",
          old_consumed, old_available);
   size_t excess_old = old_available - old_consumed;
--- a/src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
@@ -394,7 +394,8 @@ void ShenandoahGeneration::adjust_evacuation_budgets(ShenandoahHeap* heap, Shena
   // and promotion reserves.  Try shrinking OLD now in case that gives us a bit more runway for mutator allocations during
   // evac and update phases.
   size_t old_consumed = old_evacuated_committed + young_advance_promoted_reserve_used;
-  assert(old_available >= old_consumed, "Cannot consume more than is available");
+  assert(old_available >= old_consumed, "Cannot consume (" SIZE_FORMAT ") more than is available (" SIZE_FORMAT ")",
+         old_consumed, old_available);
   size_t excess_old = old_available - old_consumed;
   size_t unaffiliated_old_regions = old_generation->free_unaffiliated_regions();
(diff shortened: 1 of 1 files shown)


earthling-amzn (approved):

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
```diff
@@ -394,7 +394,19 @@ void ShenandoahGeneration::adjust_evacuation_budgets(ShenandoahHeap* heap, Shena
...
+    // We've already truncated old_evacuated_committed.  Truncate young_advance_promoted_reserve_used here.
+    assert(young_advance_promoted_reserve_used <= (33 * (old_available - old_evacuated_committed)) / 32,
+           "Round-off errors should be less than 3.125%%, committed: " SIZE_FORMAT ", reserved: " SIZE_FORMAT,
+           young_advance_promoted_reserve_used, old_available - old_evacuated_committed);
+    young_advance_promoted_reserve_used = old_available - old_evacuated_committed;
+    old_consumed = old_evacuated_committed + young_advance_promoted_reserve_used;
+  }
+
+  assert(old_available >= old_consumed, "Cannot consume (" SIZE_FORMAT ") more than is available (" SIZE_FORMAT ")",
+         old_consumed, old_available);
```
ysramakrishna, line 409: So you are just adding debugging code to catch the issue more easily? (as well as the change further below)

I don't see any change in the control flow or the state here.

Is the intention to add this change and the one below to aid debugging of a very difficult to reproduce problem?

ysramakrishna: ah, never mind, I see that you are adjusting some values there, which I missed in a first reading of the code.

I don't understand the somewhat gnarly logic of the calculations by reading this code, but I'll assume this adjustment will then prevent whatever issue was seen in the parent ticket.

I don't understand this logic well enough to review this change, so please go ahead an land this with the review you have from William.

ysramakrishna:
```
   old_consumed = old_evacuated_committed + old_available - old_evacuated_committed
                             = old_available
```
So effectively this is saying if `old_available < old_consumed`, make `old_available = old_consumed` so the subsequent assertion is satisfied. You are asserting that this can only happen if the difference is because of round-off errors.

I do have a question about this, though. Can we avoid the round-off errors in the arithmetic entirely, by doing rounding in the "safe direction" (e.g. rounding down or up) at each step where rounding is needed, rather than relying on catching the condition and adjusting the values to re-establish the invariants?

kdnilsen: It would be possible to do all round-off arithmetic more conservatively, to avoid the need for this check.  The "problem" arises when the sum of individual round-off errors is different than the single round-off error for the sum of promotions.  My preference would be to fix this in a separate PR if we choose to do so.

kdnilsen:
Though all GHA checks have passed, there are still some failures on this branch in internal CI pipelines.  Specifically:

1. perf_specjbb (2 of 5 runs) encountered:
```
# A fatal error has been detected by the Java Runtime Environment
#  SIGSEGV (0xb) at pc=0x00007fdc7a8f3939, pid=15605, tid=15607
# JRE version: OpenJDK Runtime Environment (22.0) (build 22-root-tolerate-roundoff-with-preselected-promotion-gh-x86.5328a6)
# Java VM: OpenJDK 64-Bit Server VM (22-root-tolerate-roundoff-with-preselected-promotion-gh-x86.5328a6, mixed mode, sharing, compressed oops, compressed class ptrs, shenandoah gc, linux-amd64)
# Problematic frame:
  void ShenandoahScanRemembered<ShenandoahDirectCardMarkRememberedSet>::process_clusters<ShenandoahMarkRefsClosure<(ShenandoahGenerationType)2> >(unsigned long, unsigned long, HeapWordImpl**, ShenandoahMarkRefsClosure<(ShenandoahGenerationType)2>*, bool, unsigned int)+0x289
```
2. stress_dacapo (5 of 5 runs) encountered:
```
# A fatal error has been detected by the Java Runtime Environment:
#  Internal Error (/codebuild/output/src4035/src/s3/00/src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp:1276), pid=11604, tid=11612
#  Error: Verify init-mark remembered set violation; clean card should be dirty
```
We believe the first of these is covered by issue: https://bugs.openjdk.org/browse/JDK-8315560

and the second is covered by issue: https://bugs.openjdk.org/browse/JDK-8315044

kdnilsen:
I will /integrate now, and address the known failures in separate efforts.

PR #318, opened 2023-09-01
JDK-8315479: GenShen: Expand old-gen while selecting collection set during GLOBAL GC

kdnilsen:
With GenShen, the size of old generation is as small as possible in order to maximize memory available to young gen.  This generally allows less frequent (and therefore more efficient) young-gen collections.

Under normal operation (concurrent marking of old followed by a series of mixed evacuations), old-gen is expanded at the end of a previous GC cycle in order to accommodate evacuation of mixed-evacuation candidates.

However, when we do a GLOBAL collection, we need to expand OLD gen during the selection of the collection set, depending on how many old-gen regions are selected to be evacuated during the GC evacuation phase.  This PR makes the change.

### DIFF
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahGlobalHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahGlobalHeuristics.cpp
@@ -80,15 +80,39 @@ void ShenandoahGlobalHeuristics::choose_global_collection_set(ShenandoahCollecti
                                                               size_t size, size_t actual_free,
                                                               size_t cur_young_garbage) const {
   ShenandoahHeap* heap = ShenandoahHeap::heap();
+  size_t region_size_bytes = ShenandoahHeapRegion::region_size_bytes();
   size_t capacity = heap->young_generation()->max_capacity();
-  size_t garbage_threshold = ShenandoahHeapRegion::region_size_bytes() * ShenandoahGarbageThreshold / 100;
-  size_t ignore_threshold = ShenandoahHeapRegion::region_size_bytes() * ShenandoahIgnoreGarbageThreshold / 100;
+  size_t garbage_threshold = region_size_bytes * ShenandoahGarbageThreshold / 100;
+  size_t ignore_threshold = region_size_bytes * ShenandoahIgnoreGarbageThreshold / 100;
   const uint tenuring_threshold = heap->age_census()->tenuring_threshold();
 
   size_t max_young_cset = (size_t) (heap->get_young_evac_reserve() / ShenandoahEvacWaste);
   size_t young_cur_cset = 0;
   size_t max_old_cset = (size_t) (heap->get_old_evac_reserve() / ShenandoahOldEvacWaste);
   size_t old_cur_cset = 0;
+
+  // Figure out how many unaffiliated young regions are dedicated to mutator and to evacuator.  Allow the young
+  // collector's unaffiliated regions to be transferred to old-gen if old-gen has more easily reclaimed garbage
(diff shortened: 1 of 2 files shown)


src/hotspot/share/gc/shenandoah/heuristics/shenandoahGlobalHeuristics.cpp
```diff
@@ -101,31 +125,50 @@ void ShenandoahGlobalHeuristics::choose_global_collection_set(ShenandoahCollecti
...
+          max_old_cset += region_size_bytes / ShenandoahOldEvacWaste;
+        }
+      }
       if ((new_cset <= max_old_cset) && (r->garbage() > garbage_threshold)) {
         add_region = true;
         old_cur_cset = new_cset;
       }
-    } else if (r->age() < tenuring_threshold) {
+    } else {
+      // r->is_young() && (r->age() < tenuring_threshold)
```
earthling-amzn, line 146: `assert`?

earthling-amzn (commented):
A couple of minor nits.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahGlobalHeuristics.cpp
```diff
@@ -101,31 +125,50 @@ void ShenandoahGlobalHeuristics::choose_global_collection_set(ShenandoahCollecti
   for (size_t idx = 0; idx < size; idx++) {
     ShenandoahHeapRegion* r = data[idx]._region;
     if (cset->is_preselected(r->index())) {
+      assert(false, "There should be no preselected regions during GLOBAL GC");
```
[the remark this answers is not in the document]
kdnilsen: Thanks.  I'll make this change.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahGlobalHeuristics.cpp
kdnilsen: I'll make this change also.

kdnilsen:
Though GHA checks have all passed, our internal CI pipelines have encountered a few problems:
1. stress_dacapo (5 of 5 runs) encounters:
```
#  Error: Verify init-mark remembered set violation; clean card should be dirty
```
This issue is being addressed under separate ticket: https://bugs.openjdk.org/browse/JDK-8315044

src/hotspot/share/gc/shenandoah/heuristics/shenandoahGlobalHeuristics.cpp
```diff
@@ -101,31 +125,50 @@ void ShenandoahGlobalHeuristics::choose_global_collection_set(ShenandoahCollecti
...
+          max_old_cset += region_size_bytes / ShenandoahOldEvacWaste;
+        }
+      }
       if ((new_cset <= max_old_cset) && (r->garbage() > garbage_threshold)) {
         add_region = true;
         old_cur_cset = new_cset;
       }
-    } else if (r->age() < tenuring_threshold) {
+    } else {
+      assert(r->is_young() && (r->age() < tenuring_threshold), "DeMorgan\'s law");
```
earthling-amzn, line 146: Not quite DeMorgan's law because `r` could be unaffiliated (it isn't _supposed_ to be, of course).

earthling-amzn (approved):

src/hotspot/share/gc/shenandoah/heuristics/shenandoahGlobalHeuristics.cpp
kdnilsen: I'm fixing the message here.

PR #325, opened 2023-09-14
8316299: GenShen: Reduce frequency of expedited GC

kdnilsen:
We have found that expedited GCs too frequently yield very little benefit because they occur so soon after the previous GC that there has been no accumulation of garbage.

The primary motivation for expediting GC is to avoid a situation under which there is so much "promotion" work to be performed that the urgent need to reclaim garbage from young-generation is obstructed by this promotion work.

This PR ends the practice of expediting to support promotion in place.  The effort required to promote a region in place is minimal and unlikely to contend in a major way with young collection efforts.

This PR also reduces the likelihood that we will expedite for promotion by evacuation, because it requires the amount of accumulated promo potential to exceed a particular threshold.

---------
### Progress

### Issue
 * [JDK-8316299](https://bugs.openjdk.org/browse/JDK-8316299): GenShen: Reduce frequency of expedited GC (**Bug** - P3)

### Reviewers
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - Committer) ⚠️ Review applies to [44b203a0](https://git.openjdk.org/shenandoah/pull/325/files/44b203a0d287b9b69fc4c50cbf812b2ccaea72f9)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/325/head:pull/325` \
`$ git checkout pull/325`

`$ git checkout pull/325` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/325/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 325`

`$ git pr show -t 325`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/325.diff">https://git.openjdk.org/shenandoah/pull/325.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/325#issuecomment-1719678818)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahYoungHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahYoungHeuristics.cpp
@@ -135,25 +135,16 @@ bool ShenandoahYoungHeuristics::should_start_gc() {
   // be done, we start up the young-gen GC threads so they can do some of this old-gen work.  As implemented, promotion
   // gets priority over old-gen marking.
   ShenandoahHeap* heap = ShenandoahHeap::heap();
+
+  size_t promo_expedite_threshold = (heap->young_generation()->max_capacity() * ShenandoahEvacReserve) / 512;
   size_t promo_potential = heap->get_promotion_potential();
-  if (promo_potential > 0) {
+  if (promo_potential > promo_expedite_threshold) {
     // Detect unsigned arithmetic underflow
     assert(promo_potential < heap->capacity(), "Sanity");
     log_info(gc)("Trigger (%s): expedite promotion of " SIZE_FORMAT "%s",
                  _space_info->name(),
                  byte_size_in_proper_unit(promo_potential),
                  proper_unit_for_byte_size(promo_potential));
-    return true;
-  }
-
-  size_t promo_in_place_potential = heap->get_promotion_in_place_potential();
-  if (promo_in_place_potential > 0) {
-    // Detect unsigned arithmetic underflow
-    assert(promo_in_place_potential < heap->capacity(), "Sanity");
-    log_info(gc)("Trigger (%s): expedite promotion in place of " SIZE_FORMAT "%s",
-                 _space_info->name(),
-                 byte_size_in_proper_unit(promo_in_place_potential),
-                 proper_unit_for_byte_size(promo_in_place_potential));
     return true;
   }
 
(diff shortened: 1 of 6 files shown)


src/hotspot/share/gc/shenandoah/heuristics/shenandoahYoungHeuristics.cpp
```diff
@@ -135,8 +135,10 @@ bool ShenandoahYoungHeuristics::should_start_gc() {
   // be done, we start up the young-gen GC threads so they can do some of this old-gen work.  As implemented, promotion
   // gets priority over old-gen marking.
   ShenandoahHeap* heap = ShenandoahHeap::heap();
+
+  size_t promo_expedite_threshold = (heap->young_generation()->max_capacity() * ShenandoahEvacReserve) / 512;
```
shipilev, line 139: What's the reasoning behind `512` divisor here? Since `ShenandoahEvacReserve` is percents of young/total heap size, this makes `promo_expedite_threshold` about 20% of heap size. Is that what we want?

shipilev (approved):
Looks okay, but I have a question:

src/hotspot/share/gc/shenandoah/heuristics/shenandoahYoungHeuristics.cpp
shipilev: Err, miscomputed. That would be only `1%` of heap size.

kdnilsen: Yes.  20% of 5%.  The idea: In the rare scenario that we anticipate that promotion will consume over 20% of the evacuation reserve, we'll expedite the cycle so as to improve the likelihood that GC will replenish young allocation pool before it becomes exhausted.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahYoungHeuristics.cpp
```diff
@@ -146,17 +148,6 @@ bool ShenandoahYoungHeuristics::should_start_gc() {
     return true;
   }
 
-  size_t promo_in_place_potential = heap->get_promotion_in_place_potential();
```
earthling-amzn, line 149: Can we also now delete `ShenandoahHeap::get_promotion_in_place_potential` and all the attendant code that maintains the value? This heuristic was the only place reading it.

earthling-amzn (changes requested):

src/hotspot/share/gc/shenandoah/heuristics/shenandoahYoungHeuristics.cpp
kdnilsen: Good point.  I'll do that.

earthling-amzn (approved):
Looks good - thanks!

PR #326, opened 2023-09-18
8315560: GenShen: assert failed: Object klass pointer must go to metaspace

earthling-amzn:
The SATB barrier which is on during concurrent marking of old may cause pointers into the collection set to be held by the SATB buffer. When a concurrent cycle degenerates during or before the root scan, these buffers are purged. However, they must also be purged at the end of a degenerated cycle. The buffers cannot be purged unconditionally at the beginning of a degenerated cycle because it could lose pointers necessary to complete the concurrent mark of the young generation.

---------
### Progress

### Issue
 * [JDK-8315560](https://bugs.openjdk.org/browse/JDK-8315560): GenShen: assert failed: Object klass pointer must go to metaspace (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/326/head:pull/326` \
`$ git checkout pull/326`

`$ git checkout pull/326` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/326/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 326`

`$ git pr show -t 326`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/326.diff">https://git.openjdk.org/shenandoah/pull/326.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/326#issuecomment-1723892895)


kdnilsen (approved):
As discussed off-channel, there may be compounding issues that trigger this same or similar error message.  My rr recording of this symptom might suggest that there's something else at play here, but I agree that this fix is necessary.

PR #327, opened 2023-09-18
8312116  GenShen: make instantaneous allocation rate triggers more timely

kdnilsen:
When heuristics fail to trigger because an instantaneous "allocation spike" is not so large as to consume all available memory before GC completes, this assessment is based on an assumption that the allocation rate remains constant, and it ignores the time that will be lost due to the _interval between consecutive allocation spike measurements.

This PR watches for "acceleration" of allocation rates.  When acceleration of allocation is detected in 3 consecutive allocation spike measurements, it calculates a best-fit curve (assuming constant acceleration) and predicts the memory to be consumed during the time that spans both the next sample interval and the GC effort that follows it.  If the memory to be allocated according to anticipated acceleration of allocations during this time span exceeds what is available, we trigger immediately.


kdnilsen:
I'm leaving this in draft mode until I complete functional and performance testing and remove the temporary instrumentation code.

kdnilsen:
Experiments with various workloads show these changes to decrease the number of degenerated GCs.  In some cases, the greater sensitivity of the acceleration trigger causes more frequent GCs than without it.  I am experimenting with various configuration options on multiple workloads to determine an "optimal" default configuration.

kdnilsen:
This now passes all regression tests.  However, performance is not acceptable.  Going to do a bit of refinement.

PR #337, opened 2023-10-11
8318019: GenShen: Fix assertion to allow empty evacuation cycles

kdnilsen:
Assertion needs to allow for gc phase EVACUATION even if no forwarding objects because we are only promoting in place.

---------
### Progress

### Issue
 * [JDK-8318019](https://bugs.openjdk.org/browse/JDK-8318019): GenShen: Fix assertion to allow empty evacuation cycles (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/337/head:pull/337` \
`$ git checkout pull/337`

`$ git checkout pull/337` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/337/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 337`

`$ git pr show -t 337`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/337.diff">https://git.openjdk.org/shenandoah/pull/337.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/337#issuecomment-1759739230)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -2423,9 +2423,9 @@ void ShenandoahHeap::set_concurrent_young_mark_in_progress(bool in_progress) {
 void ShenandoahHeap::set_concurrent_old_mark_in_progress(bool in_progress) {
 #ifdef ASSERT
   // has_forwarded_objects() iff UPDATEREFS or EVACUATION
-  bool has_forwarded = has_forwarded_objects()? 1: 0;
-  bool updating_or_evacuating = _gc_state.is_set(UPDATEREFS | EVACUATION)? 1: 0;
-  bool evacuating = gc_state.is_set(EVACUATION)? 1: 0;
+  bool has_forwarded = has_forwarded_objects();
+  bool updating_or_evacuating = _gc_state.is_set(UPDATEREFS | EVACUATION);
+  bool evacuating = _gc_state.is_set(EVACUATION);
   assert ((has_forwarded == updating_or_evacuating) || (evacuating && !has_forwarded && collection_set()->is_empty()),
           "Updating or evacuating iff has forwarded object, or evacuation phase is promoting in place without forwarding");
 #endif
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -2425,7 +2425,9 @@ void ShenandoahHeap::set_concurrent_old_mark_in_progress(bool in_progress) {
   // has_forwarded_objects() iff UPDATEREFS or EVACUATION
   bool has_forwarded = has_forwarded_objects()? 1: 0;
   bool updating_or_evacuating = _gc_state.is_set(UPDATEREFS | EVACUATION)? 1: 0;
-  assert (has_forwarded == updating_or_evacuating, "Has forwarded objects iff updating or evacuating");
+  bool evacuating = gc_state.is_set(EVACUATION)? 1: 0;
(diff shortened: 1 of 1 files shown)


kdnilsen:
After testing, I'll match this to a JBS issue.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -2425,7 +2425,9 @@ void ShenandoahHeap::set_concurrent_old_mark_in_progress(bool in_progress) {
   // has_forwarded_objects() iff UPDATEREFS or EVACUATION
   bool has_forwarded = has_forwarded_objects()? 1: 0;
   bool updating_or_evacuating = _gc_state.is_set(UPDATEREFS | EVACUATION)? 1: 0;
-  assert (has_forwarded == updating_or_evacuating, "Has forwarded objects iff updating or evacuating");
+  bool evacuating = gc_state.is_set(EVACUATION)? 1: 0;
```
zhengyu123, line 2428: suggestion: 
`
bool evacuating = gc_state.is_set(EVACUATION);
`
pre-exist:
`bool has_forwarded = has_forwarded_objects()? 1: 0;` to `bool has_forwarded = has_forwarded_objects();`

zhengyu123 (changes requested):

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
kdnilsen: Thanks.  I'll fix those, and my misspelling of gc_state.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -2423,9 +2423,11 @@ void ShenandoahHeap::set_concurrent_young_mark_in_progress(bool in_progress) {
 void ShenandoahHeap::set_concurrent_old_mark_in_progress(bool in_progress) {
 #ifdef ASSERT
   // has_forwarded_objects() iff UPDATEREFS or EVACUATION
-  bool has_forwarded = has_forwarded_objects()? 1: 0;
-  bool updating_or_evacuating = _gc_state.is_set(UPDATEREFS | EVACUATION)? 1: 0;
-  assert (has_forwarded == updating_or_evacuating, "Has forwarded objects iff updating or evacuating");
+  bool has_forwarded = has_forwarded_objects();
+  bool updating_or_evacuating = _gc_state.is_set(UPDATEREFS | EVACUATION);
+  bool evacuating = _gc_state.is_set(EVACUATION);
+  assert ((has_forwarded == updating_or_evacuating) || (evacuating && !has_forwarded && collection_set()->is_empty()),
```
zhengyu123, line 2429: Sorry, I did not realize this is generational Shenandoah. I am a bit puzzled why setting `EVACUATION` flag if collection set is empty?

kdnilsen: It's a trick we use for a form of "abbreviated" cycle.  Sometimes, we have nothing to evacuate, but we have a bunch of regions that reside in young, which are old enough to be promoted.  If they are populated with a sufficient amount of live data, we promote them "in place" by simply relabeling the affiliation of the region rather than evacuating it.  But there's a non-trivial amount of processing required on the content of the region when it becomes old, so that the region can now maintain a remembered set of pointers from old to young.  We normally do this processing as part of the evacuation cycle.  In very rare scenarios, we'll end up with an evacuation cycle that only deals with promote-in-place regions, and nothing else...  That's the special scenario we're talking about here.

zhengyu123: Thanks for the explanation!

earthling-amzn (approved):

PR #343, opened 2023-10-18
8318462: [GenShen] Prevent unsafe access to displaced mark-word

rkennke:
We must be very careful when we access (load or update) the object age concurrently because the age is stored in the object's header and that header can be 'displaced'. This means that the header is overloaded with a pointer (stack or ObjectMonitor*) and the original header is stored at that location. The header can also be in INFLATING state, which means that a stack-lock is currently in the process to be inflated to an ObjectMonitor. Let's consider the cases separately:

Loading object age:
- INFLATING: we can not access the header, and thus not the age.
- Stack-locked: the thread which locks the object can unlock at any time, concurrently. It is not safe to access the header, and thus the age.
- Monitor-locked: monitors don't go away without coordinating with Java threads and possibly GC threads. This coordination is done by handshaking the thread - a Java thread would be brought to its next safepoint, and GC threads, which are typically participating in handshakes by joining the SuspendibleThreadSet, can opt to handshake at a safe point by calling SuspendibleThreadSet::yield(). If 'our' thread participates in handshakes in one way or the other, it is safe to access an ObjectMonitor once we got a valid ObjectMonitor* from an object's header, but only until that thread handshakes. Long story short: it is typically safe to access ObjectMonitor :-)

Updating object age:
Updating object age only happens during evacuation, and only on a new copy of an object, before that copy has been published. Access to the object header is therefore fully thread-local and not problematic. What *is* problematic is when we have to deal with displaced headers, because displaced headers are *not* thread local, and must be considered a shared resource. A competing thread may succeed to evacuate the object and publish its copy before 'our' thread can do so. If 'our' thread would update the displaced header, it may over-write whatever the other thread has done.
- INFLATING: We cannot access the header at all. However, that should not happen: a thread can only inflate once it has successfully evacuated an object, and 
- Stack-locked: the other thread may succeed evacuation and continue to unlock the object, at which point the stack-pointer would be 'dangling' and we may over-write random information on the foreign thread stack.
- Monitor-locked: even though monitor deflation is coordinated (see above), updating the displaced header in a monitor might over-write whatever a competing thread has done there.

For these reasons, we must not update displaced headers, and must be careful when loading displaced headers - in particular, only ever load from ObjectMonitor.

---------
### Progress

### Issue
 * [JDK-8318462](https://bugs.openjdk.org/browse/JDK-8318462): [GenShen] Prevent unsafe access to displaced mark-word (**Bug** - P2)

### Reviewers
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [113a3d2b](https://git.openjdk.org/shenandoah/pull/343/files/113a3d2bbec5679a9c25550fc825feb75a7b41e8)
 * [William Kemper](https://openjdk.org/census#wkemper) (@earthling-amzn - Committer) ⚠️ Review applies to [113a3d2b](https://git.openjdk.org/shenandoah/pull/343/files/113a3d2bbec5679a9c25550fc825feb75a7b41e8)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/343/head:pull/343` \
`$ git checkout pull/343`

`$ git checkout pull/343` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/343/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 343`

`$ git pr show -t 343`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/343.diff">https://git.openjdk.org/shenandoah/pull/343.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/343#issuecomment-1770660068)


src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
```diff
@@ -537,32 +538,36 @@ inline oop ShenandoahHeap::try_evacuate_object(oop p, Thread* thread, Shenandoah
...
 // mutator
-uint ShenandoahHeap::get_object_age_concurrent(oop obj) {
+uint ShenandoahHeap::get_object_age(oop obj) {
   // This is impossible to do unless we "freeze" ABA-type oscillations
   // With Lilliput, we can do this more easily.
   markWord w = obj->mark();
-  // We can do better for objects with inflated monitor
-  if (w.is_being_inflated() || w.has_displaced_mark_helper()) {
+  if (w.has_monitor()) {
+    w = w.monitor()->header();
```
ysramakrishna, line 569: Isn't this prone to potential error because the monitor might be recycled between line 567 and line 569?

rkennke: Monitor deflation is coordinated with Java threads by safepointing (or more precisely handshaking) the Java threads. The same works with GC threads if they participate in safepointing/handshaking. This is usually achieved by joining SuspendibleThreadSet. The effect of is that all such threads can safely access monitors without fearing that they go away. When we got a valid monitor pointer, then that monitor will not disappear before the next safepoint, and safepoints cannot happen at random points in execution. For example, a thread that has joined STS would only be able to safepoint when it calls STS::yield(). We might want to verify that all GC threads which can evacuate objects are properly joining (and yielding) STS.

zhengyu123:
Just FYI: Here is another place to update object age, probably you want to consolidate.

https://github.com/openjdk/shenandoah/blob/master/src/hotspot/share/gc/shenandoah/shenandoahStringDedup.inline.hpp#L55

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
```diff
@@ -537,32 +538,46 @@ inline oop ShenandoahHeap::try_evacuate_object(oop p, Thread* thread, Shenandoah
...
 void ShenandoahHeap::increase_object_age(oop obj, uint additional_age) {
-  markWord w = obj->has_displaced_mark() ? obj->displaced_mark() : obj->mark();
-  w = w.set_age(MIN2(markWord::max_age, w.age() + additional_age));
-  if (obj->has_displaced_mark()) {
-    obj->set_displaced_mark(w);
-  } else {
+  // This operates on new copy of an object. This means that the object's mark-word
+  // is thread-local and therefore safe to access. However, when the mark is
+  // displaced (i.e. stack-locked or monitor-locked), then it must be considered
+  //a shared memory location. It can/ be accessed by other threads.
```
kdnilsen, line 544: Minor formatting issues in this line of comment.

kdnilsen: and below.

src/hotspot/share/gc/shenandoah/shenandoahMark.inline.hpp
```diff
@@ -119,7 +119,7 @@ inline void ShenandoahMark::count_liveness(ShenandoahLiveData* live_data, oop ob
     assert(heap->mode()->is_generational(), "Only if generational");
     if (ShenandoahGenerationalAdaptiveTenuring && !ShenandoahGenerationalCensusAtEvac) {
       assert(region->is_young(), "Only for young objects");
-      uint age = ShenandoahHeap::get_object_age_concurrent(obj);
+      uint age = ShenandoahHeap::get_object_age(obj);
       CENSUS_NOISE(heap->age_census()->add(age, region->age(), region->youth(), size, worker_id);)
```
kdnilsen, line 123: I'm thinking if get_object_age() returns max_age+1, the CENSUS_NOISE macros might end up with invalid data.  Don't we need a check here?

kdnilsen: I think the safe thing to do is treat max_age+1 as if it were 0, insofar as promotion semantics is concerned.  We may delay promotion that way.

rkennke:
> Just FYI: Here is another place to update object age, probably you want to consolidate.
> 
> https://github.com/openjdk/shenandoah/blob/master/src/hotspot/share/gc/shenandoah/shenandoahStringDedup.inline.hpp#L55

Right. The problem there is that this block of code does a sort of concurrent/atomic update of the object age that would require re-working and/or adding age helper methods. I'd rather not do that now.

src/hotspot/share/gc/shenandoah/shenandoahMark.inline.hpp
kdnilsen: Thanks for change.  Code might read easier with parens around (age == markWord::max_age + 1), but ok as is.

kdnilsen (approved):

src/hotspot/share/gc/shenandoah/shenandoahMark.inline.hpp
ysramakrishna: The add method of ShenandoahAgeCensus explicitly recognizes and uses the sentinel value, and explicitly skips sentinels that are used to signal that the age could not be read:

```
CENSUS_NOISE(void ShenandoahAgeCensus::add(uint obj_age, uint region_age, uint region_youth, size_t size, uint worker_id) {)
NO_CENSUS_NOISE(void ShenandoahAgeCensus::add(uint obj_age, uint region_age, size_t size, uint worker_id) {)
  if (obj_age <= markWord::max_age) {
    assert(obj_age < MAX_COHORTS && region_age < MAX_COHORTS, "Should have been tenured");
#ifdef SHENANDOAH_CENSUS_NOISE
    // Region ageing is stochastic and non-monotonic; this vitiates mortality
    // demographics in ways that might defeat our algorithms. Marking may be a
    // time when we might be able to correct this, but we currently do not do
    // this. Like skipped statistics further below, we want to track the
    // impact of this noise to see if this may be worthwhile. JDK-<TBD>.
    uint age = obj_age;
    if (region_age > 0) {
      add_aged(size, worker_id);   // this tracking is coarse for now
      age += region_age;
      if (age >= MAX_COHORTS) {
        age = (uint)(MAX_COHORTS - 1);  // clamp
        add_clamped(size, worker_id);
      }
    }
    if (region_youth > 0) {   // track object volume with retrograde age
      add_young(size, worker_id);
    }
#else   // SHENANDOAH_CENSUS_NOISE
    uint age = MIN2(obj_age + region_age, (uint)(MAX_COHORTS - 1));  // clamp
#endif  // SHENANDOAH_CENSUS_NOISE
    get_local_age_table(worker_id)->add(age, size);
  } else {
    // update skipped statistics
    CENSUS_NOISE(add_skipped(size, worker_id);)
  }
}
```

I think we should leave in place the original code, where all of the special handling of the sentinel values is relegated to the ShenandoahAgeCensus's add method, with no special handling needed at the caller.

kdnilsen: Thanks.  That looks good.  I concur.

ysramakrishna: I realize though that your comment kind of applies to the evacuation tracker code that needs to be cognizant of the change in the semantics of `get_object_age` following Roman's change.

This is the relevant call at:
https://github.com/openjdk/shenandoah/blob/113a3d2bbec5679a9c25550fc825feb75a7b41e8/src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp#L491
which lands us at:
https://github.com/openjdk/shenandoah/blob/113a3d2bbec5679a9c25550fc825feb75a7b41e8/src/hotspot/share/gc/shenandoah/shenandoahEvacTracker.cpp#L60
and thence at:
https://github.com/openjdk/shenandoah/blob/113a3d2bbec5679a9c25550fc825feb75a7b41e8/src/hotspot/share/gc/shared/ageTable.hpp#L63
which might either assert or silently overflow past the end of the table.

It would be useful to filter those records out in ShenandoahEvacuationStats::record_age() or at the caller.

earthling-amzn (approved):
Internal testing looks good.

zhengyu123:
> > Just FYI: Here is another place to update object age, probably you want to consolidate.
> > https://github.com/openjdk/shenandoah/blob/master/src/hotspot/share/gc/shenandoah/shenandoahStringDedup.inline.hpp#L55
> 
> Right. The problem there is that this block of code does a sort of concurrent/atomic update of the object age that would require re-working and/or adding age helper methods. I'd rather not do that now.

I think for GenShen, StringDeduplication should not update age, so code can just be simplified as following:

```
bool ShenandoahStringDedup::is_candidate(oop obj) {
  if (!is_string_candidate(obj)) {
    return false;
  }

  const markWord mark = obj->mark();

  // Having/had displaced header, too risky to deal with them, skip
  if ((LockingMode != LM_LIGHTWEIGHT && mark == markWord::INFLATING()) || mark.has_displaced_mark_helper()) {
    return false;
  }

  return StringDedup::is_threshold_age(new_mark.age()) &&
             !dedup_requested(obj);
}
```

ysramakrishna (approved):
Nice; thanks also for the documentation!

🚢

rkennke:
Thanks!

/integrate

PR #352, opened 2023-11-03
8319342: GenShen: Reset the count of degenerated cycles in a row following Full GC

kdnilsen:
After a Full GC completes, reset the count on consecutive degenerated cycles to zero.  This prevents automatic triggering of back-to-back Full GC.

---------
### Progress

### Issue
 * [JDK-8319342](https://bugs.openjdk.org/browse/JDK-8319342): GenShen: Reset the count of degenerated cycles in a row following Full GC (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/352/head:pull/352` \
`$ git checkout pull/352`

`$ git checkout pull/352` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/352/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 352`

`$ git pr show -t 352`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/352.diff">https://git.openjdk.org/shenandoah/pull/352.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/352#issuecomment-1793123187)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahPassiveHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahPassiveHeuristics.cpp
@@ -44,6 +44,11 @@ bool ShenandoahPassiveHeuristics::should_unload_classes() {
   return can_unload_classes();
 }
 
+bool ShenandoahPassiveHeuristics::should_degenerate_cycle() {
+  // Always fal to Degenerated GC, if enabled
+  return ShenandoahDegeneratedGC;
+}
+
 void ShenandoahPassiveHeuristics::choose_collection_set_from_regiondata(ShenandoahCollectionSet* cset,
                                                                         RegionData* data, size_t size,
                                                                         size_t actual_free) {
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahPassiveHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahPassiveHeuristics.cpp
@@ -44,11 +44,6 @@ bool ShenandoahPassiveHeuristics::should_unload_classes() {
   return can_unload_classes();
 }
 
-bool ShenandoahPassiveHeuristics::should_degenerate_cycle() {
-  // Always fail to Degenerated GC, if enabled
-  return ShenandoahDegeneratedGC;
-}
-
 void ShenandoahPassiveHeuristics::choose_collection_set_from_regiondata(ShenandoahCollectionSet* cset,
                                                                         RegionData* data, size_t size,
                                                                         size_t actual_free) {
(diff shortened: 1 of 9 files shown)


src/hotspot/share/gc/shenandoah/shenandoahCollectorPolicy.hpp
```diff
@@ -61,6 +61,8 @@ class ShenandoahCollectorPolicy : public CHeapObj<mtGC> {
   ShenandoahSharedFlag _in_shutdown;
   ShenandoahTracer* _tracer;
 
+  uint _degenerated_cycles_in_a_row;
```
earthling-amzn, line 64: Nit, but we could call this `_consecutive_degenerated_gcs` for consistency with other fields in this class? (Also, possibly rename `_abbreviated_cycles` to `_abbreviated_gcs` for the same reason.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahPassiveHeuristics.cpp
```diff
@@ -44,11 +44,6 @@ bool ShenandoahPassiveHeuristics::should_unload_classes() {
   return can_unload_classes();
 }
 
-bool ShenandoahPassiveHeuristics::should_degenerate_cycle() {
-  // Always fail to Degenerated GC, if enabled
-  return ShenandoahDegeneratedGC;
```
earthling-amzn, line 49: Are we changing how the passive heuristic behaves now? With this change `ShenandoahCollectorPolicy::should_degenerate_cycle` would return `false` sometimes when the passive heuristic would have "degenerated".

earthling-amzn (changes requested):

src/hotspot/share/gc/shenandoah/heuristics/shenandoahPassiveHeuristics.cpp
ysramakrishna: I agee with William... You can leave all state (counters) and its manipulation entirely in the heuristics object, and let the policy object query the heuristics like you do today, and inform it when it does do a full gc, etc. so relevant state may be updated.

Of course, the policy should not independently reach in and change the state maintained in the heuristics, just query that state, and inform it of events that might entail changes to that state.

We can clean things up by asking what the role of heuristics is vs what the role of policy is. It seems as if policy is broad and heuristics is more specific within that broad policy. It is also possible to keep the state in the policy, be mutated only by policy in response to events, and have the heuristics be completely state-free, but call-back to policy to use any state it would use for its decisions. Clarifying this structure and their relationship, and sticking to a consistent (documented) protocol here would avoid confusion and errors.

src/hotspot/share/gc/shenandoah/shenandoahCollectorPolicy.hpp
kdnilsen: Thanks.  Making this change.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahPassiveHeuristics.cpp
kdnilsen: Good catch.  That was an unintended consequence.  I will return the methods to the Heuristics hierarchy, but will query the consecutive_degenerated_gcs value from collector policy.

earthling-amzn (approved):

ysramakrishna (approved):

PR #353, opened 2023-11-08
8319671 GenShen: Old marking may encounter invalid pointers in SATB buffers

earthling-amzn:
Weak handles which are null'd out after final marking of young, but during marking of old, will have their original referent placed into a SATB buffer. These entries need to be purged before old marking resumes. In the case of abbreviated young cycles, these entries would not be purged.

---------
### Progress

### Issue
 * [JDK-8319671](https://bugs.openjdk.org/browse/JDK-8319671): GenShen: Old marking may encounter invalid pointers in SATB buffers (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/353/head:pull/353` \
`$ git checkout pull/353`

`$ git checkout pull/353` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/353/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 353`

`$ git pr show -t 353`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/353.diff">https://git.openjdk.org/shenandoah/pull/353.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/353#issuecomment-1800842768)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
@@ -1266,8 +1266,14 @@ void ShenandoahConcurrentGC::op_final_roots() {
   heap->set_evacuation_in_progress(false);
 
   if (heap->mode()->is_generational()) {
-    ShenandoahMarkingContext *ctx = heap->complete_marking_context();
+    // If the cycle was shortened for having enough immediate garbage, this could be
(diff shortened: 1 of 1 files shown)


ysramakrishna (approved):
This seems safe, but we should separately explore as a follow-up a more discriminatory SATB barrier for old vs young marking so a to avoid these wrinkles and the extra work.

PR #355, opened 2023-11-09
8319867: GenShen: Make old regions parseable at end of concurrent cycles

earthling-amzn:
### Background
When Shenandoah is performing a concurrent mark of the old generation, it is not safe for the young generation's remembered set scan to use the old mark bitmap. For this reason, any old regions that were not included in a collection set (either mixed or a global collection) must be made _parseable_ for the remembered set scan (we call this "coalescing and filling" in the code).

### Description
Before this change, Shenandoah has deferred making these regions parseable until it begins the young generation bootstrap cycle. This is fine, except that we have also recently made old generation collections subordinate to young collections. In other words, an old collection cannot start unless Shenandoah wants to first start a young collection.

Unfortunately, the additional time to concurrently coalesce and fill old regions before resetting the old mark bitmaps is not accounted for in the heuristics and this increases the likelihood that the heuristic has started the cycles too late and that the cycle will degenerate into a stop the world pause.

After this change, Shenandoah will make old regions parseable immediately following the last mixed collection (or following a global collection). At this point, Shenandoah has just finished reclaiming memory and we expect less urgency to coalesce and fill old regions. It also means that young bootstrap cycles will not take (much) longer than conventional young generation collections.


kdnilsen (approved):
Thank you.  I believe this will improve latency for several of our workloads.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
```diff
@@ -383,67 +374,68 @@ void ShenandoahOldGeneration::transition_to(State new_state) {
...
-      assert(heap->young_generation()->old_gen_task_queues() == nullptr, "Cannot become idle when setup for bootstrapping.");
-      break;
     case FILLING:
-      assert(_state == IDLE || _state == WAITING_FOR_FILL, "Cannot begin filling without first completing evacuations, state is '%s'", state_name(_state));
-      assert(heap->is_prepare_for_old_mark_in_progress(), "Should be preparing for old mark now.");
+      assert(heap->is_old_bitmap_stable(), "Cannot begin filling without first completing marking, state is '%s'", state_name(_state));
+      assert(_old_heuristics->has_coalesce_and_fill_candidates(), "Cannot begin filling without something to fill.");
+      break;
+    case WAITING_FOR_BOOTSTRAP:
+      // GC cancellation can send us back to IDLE from any state.
```
ysramakrishna, line 432: Since there isn't now a state named "IDLE", it might be less confusing to just say "FILLING" or "WAITING_FOR_BOOTSTRAP"? Should we assert here that there are no `coalesce_and_fill_candidates` or is that implied by perhaps stronger checks below? (Wouldn't hurt even if so.)

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
```diff
@@ -383,67 +374,68 @@ void ShenandoahOldGeneration::transition_to(State new_state) {
...
-      // GC cancellation can send us back to IDLE from any state.
-      assert(!heap->is_concurrent_old_mark_in_progress(), "Cannot become idle during old mark.");
-      assert(_old_heuristics->unprocessed_old_collection_candidates() == 0, "Cannot become idle with collection candidates");
-      assert(!heap->is_prepare_for_old_mark_in_progress(), "Cannot become idle while making old generation parseable.");
-      assert(heap->young_generation()->old_gen_task_queues() == nullptr, "Cannot become idle when setup for bootstrapping.");
-      break;
     case FILLING:
-      assert(_state == IDLE || _state == WAITING_FOR_FILL, "Cannot begin filling without first completing evacuations, state is '%s'", state_name(_state));
-      assert(heap->is_prepare_for_old_mark_in_progress(), "Should be preparing for old mark now.");
+      assert(heap->is_old_bitmap_stable(), "Cannot begin filling without first completing marking, state is '%s'", state_name(_state));
```
ysramakrishna, line 428: Do you also want to say:

```
assert(_state > FILLING && _state != WAITING_FOR_BOOTSTRAP, "Illegal state transition"):
```

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
ysramakrishna:
```
assert(_state != BOOTSTRAP && _state != WAITING_FOR_BOOTSTRAP, "Illegal transition");
```

ysramakrishna: I am thinking that the most parsimonious way of covering all bases is by associating a pre-condition with entering each state and a post-condition with exiting each state, and verifying those conditions in the transition validation methods. But may be that can be done as a separate PR if it makes sense.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
```diff
@@ -452,14 +444,10 @@ void ShenandoahOldGeneration::validate_transition(State new_state) {
       assert(heap->young_generation()->old_gen_task_queues() != nullptr, "Young generation needs old mark queues.");
       assert(heap->is_concurrent_old_mark_in_progress(), "Should be marking old now.");
       break;
-    case WAITING_FOR_EVAC:
-      assert(_state == IDLE || _state == MARKING, "Cannot have old collection candidates without first marking, state is '%s'", state_name(_state));
+    case EVACUATING:
+      assert(_state == WAITING_FOR_BOOTSTRAP || _state == MARKING, "Cannot have old collection candidates without first marking, state is '%s'", state_name(_state));
```
ysramakrishna, line 448: FWIW, the ASCII art state-transition-diagram doesn't show the transition from WAITING_FOR_BOOTSTRAP directly into EVACUATING, without passing through an intermediate MARKING.

ysramakrishna (commented):
Some remarks, not all of them need to be immediately actioned (if the remarks are correct).

Was there a test case that revealed this? If so, can we add that to the JBS ticket and mentioned as part of the testing for the PR?

ysramakrishna (approved):

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
```diff
@@ -265,9 +256,10 @@ bool ShenandoahOldGeneration::entry_coalesce_and_fill() {
   return coalesce_and_fill();
 }
 
+// Make the old generation regions parseable, so they can be safely
```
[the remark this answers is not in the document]
earthling-amzn:
```
[0] % grep -irI parsable src/ | wc -l 
352

[0] % grep -irI parseable src/ | wc -l
84
```
Indeed. I'll fix up all the misspellings in `gc/shenandoah` in a separate PR. Thanks for the tip!

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
earthling-amzn: Hmm - I thought that's what this code is doing. Or rather, it is asserting what conditions must hold to _be_ in this state (checking pre/post conditions simultaneously).

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
earthling-amzn: No - at the end of a _GLOBAL_ collection, it may put the old generation in this state without progressing through the other states.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
earthling-amzn: Yes, that can only happen at the end of a _GLOBAL_ collection. This is explained in the comment above the diagram. I will add it to the diagram.

earthling-amzn:
We don't have a test that shows this, just thought experiments.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
ysramakrishna: Got it. One way to manage this and still get stronger checking during all, but in particular non-global, collections, might be to organize the state transition checks per collection type, and make assertions specific to the kind of collection in which the transition is happening/being checked. But perhaps that is overkill, and in any case not necessary in this PR.

ysramakrishna (approved):
LGTM; reviewed!

🚢

PR #361, opened 2023-11-17
JDK-8320336: GenShen: Reduce proactive humongous defragmentation efforts

kdnilsen:
Only trigger defragmentation of old-gen memory if JVM is configured to reserve a certain amount of the heap to support humongous objects.  Even when the JVM is so configured, be slightly less aggressive in triggering defragmentation of old-gen memory.

---------
### Progress

### Issue
 * [JDK-8320336](https://bugs.openjdk.org/browse/JDK-8320336): GenShen: Reduce proactive humongous defragmentation efforts (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/361/head:pull/361` \
`$ git checkout pull/361`

`$ git checkout pull/361` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/361/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 361`

`$ git pr show -t 361`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/361.diff">https://git.openjdk.org/shenandoah/pull/361.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/361#issuecomment-1816807892)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -3126,16 +3126,19 @@ void ShenandoahHeap::rebuild_free_set(bool concurrent) {
   // Rebuild free set based on adjusted generation sizes.
   _free_set->rebuild(young_cset_regions, old_cset_regions);
 
-  if (mode()->is_generational()) {
+  if (mode()->is_generational() && (ShenandoahGenerationalHumongousReserve > 0)) {
     size_t old_region_span = (first_old_region <= last_old_region)? (last_old_region + 1 - first_old_region): 0;
     size_t allowed_old_gen_span = num_regions() - (ShenandoahGenerationalHumongousReserve * num_regions() / 100);
 
     // Tolerate lower density if total span is small.  Here's the implementation:
-    //   if old_gen spans more than 100% and density < 87.5%, trigger old-defrag
-    //   else if old_gen spans more than 87.5% and density < 75%, trigger old-defrag
-    //   else if old_gen spans more than 75% and density < 62.5%, trigger old-defrag
-    //   else if old_gen spans more than 62.5% and density < 50%, trigger old-defrag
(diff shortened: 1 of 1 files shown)


earthling-amzn (approved):
Looks good to me.

ysramakrishna (approved):
Sorry for the delay. LGTM.

One question relating to the performance changes stemming from this fix:
- Effectively, it seems that the density/span based triggering is now disabled by default (because of the change at line 3129 and the default setting of humongous reserve at 0)
- Does this effectively win back the difference caused by the original change?
- However, were you to enable it, say by an appropriate positive value for the humongous reserve, does the less aggressive triggering result in a performance improvement?
- Are there any (rough) performance numbers, by chance, to share for the two changes separately?

kdnilsen:
The trigger is disabled by default.  Humongous fragmentation is one of the issues that results in very high variability in performance on some workloads.  Humongous memory gets fragmented when humongous objects happen to be allocated in "the middle" of the heap instead of getting allocated at the start of the heap.  It is difficult to predict or even contrive programs that will cause this to happen.  But the typical scenario is:

1. We are operating near the "cliff's edge" so young gen consumes all of the available heap.
2. Some number of young-gen regions residing near the bottom of heap which are not necessarily contiguous are promoted in place rather than getting compacted.
3. Free regions are now scattered between these promoted-in-place heap regions, and we have no sufficiently long run of consecutive free regions to satisfy an allocation request.

With this change: We will still endeavor to defragment the old-gen memory whenever we perform a concurrent marking effort.  Concurrent marking efforts are now triggered by guaranteed interval or by growth since previous old mark.

If a service desires to keep memory defragmented because they expect frequent allocation of humongous objects, they would be wise to set -XX:ShenandoahGenerationalHumongousReserve=5 (or any value grater than 0).

PR #371, opened 2023-12-14
8321816: GenShen: Provide a minimum amount of time for an old collection to run

earthling-amzn:
Young collections are allowed to interrupt old collections. Under heavy load, this may result in starvation of old collections - which may in turn lead to even more memory pressure on the young generation. Adding a minimum guaranteed amount of time for the old gc to make progress will help avoid this situation.

---------
### Progress

### Issue
 * [JDK-8321816](https://bugs.openjdk.org/browse/JDK-8321816): GenShen: Provide a minimum amount of time for an old collection to run (**Enhancement** - P4)

### Reviewers
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [25de1177](https://git.openjdk.org/shenandoah/pull/371/files/25de11778e7b20fb53636535e58fcee4c88fb30b)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/371/head:pull/371` \
`$ git checkout pull/371`

`$ git checkout pull/371` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/371/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 371`

`$ git pr show -t 371`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/371.diff">https://git.openjdk.org/shenandoah/pull/371.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/371#issuecomment-1856443368)


kdnilsen (approved):

src/hotspot/share/gc/shenandoah/shenandoahRegulatorThread.cpp
```diff
@@ -76,7 +76,7 @@ void ShenandoahRegulatorThread::regulate_concurrent_cycles() {
           log_info(gc)("Heuristics request for global (unload classes) accepted.");
         }
       } else {
-        if (_young_heuristics->should_start_gc()) {
+        if (should_start_young()) {
```
ysramakrishna, line 79: 1. Could you explain (and document) why we don't do the same check `should_start_young` also at line 88 further below, _then_ `start_young_cycle`, where we find the control thread in the mode `servicing_old`?
2. The split of the roles of checks between the control points of this loop and the heuristics, and the actual requests to start a cycle look like they could do with a bit of clean up for consistency. At the current time (not as a result of your current change), the checks and the actions are scattered over a few routines so that reasoning about their correctness (or their behaviour) can be a bit challenging. This latter clean-up can, however, wait for the future.

src/hotspot/share/gc/shenandoah/shenandoahRegulatorThread.hpp
```diff
@@ -87,6 +87,8 @@ class ShenandoahRegulatorThread: public ConcurrentGCThread {
   void regulator_sleep();
 
   bool request_concurrent_gc(ShenandoahGenerationType generation);
+
+  bool should_start_young();
```
ysramakrishna, line 91: Please add a brief line of documentation.

ysramakrishna (commented):
Looks generally good, but a couple of clarification/documentation requests/questions.

src/hotspot/share/gc/shenandoah/shenandoahRegulatorThread.cpp
earthling-amzn: Good catch. I published this PR prematurely.

earthling-amzn:
The difference between `ShenandoahMinimumOldMarkTimeMs=10` and `ShenandoahMinimumOldMarkTimeMs=100` is not so great, but the effect is notable with `ShenandoahMinimumOldMarkTimeMs=`1000`.

### -XX:ShenandoahMinimumOldMarkTimeMs=1000

```
1000/genshen/specjbb2015/1/specjbb2015.jvm.log     6 Completed Old GCs (2.39%)
1000/genshen/specjbb2015/1/specjbb2015.jvm.log       8 mixed
1000/genshen/specjbb2015/1/specjbb2015.jvm.log       2 interruptions
--
1000/genshen/specjbb2015/2/specjbb2015.jvm.log     7 Completed Old GCs (2.49%)
1000/genshen/specjbb2015/2/specjbb2015.jvm.log      10 mixed
1000/genshen/specjbb2015/2/specjbb2015.jvm.log       3 interruptions
```

### -XX:ShenandoahMinimumOldMarkTimeMs=100
```
100/genshen/specjbb2015/1/specjbb2015.jvm.log     7 Completed Old GCs (1.91%)
100/genshen/specjbb2015/1/specjbb2015.jvm.log       8 mixed
100/genshen/specjbb2015/1/specjbb2015.jvm.log      24 interruptions
--
100/genshen/specjbb2015/2/specjbb2015.jvm.log     6 Completed Old GCs (2.39%)
100/genshen/specjbb2015/2/specjbb2015.jvm.log       6 mixed
100/genshen/specjbb2015/2/specjbb2015.jvm.log       8 interruptions
```

### -XX:ShenandoahMinimumOldMarkTimeMs=10
```
10/genshen/specjbb2015/1/specjbb2015.jvm.log     6 Completed Old GCs (2.91%)
10/genshen/specjbb2015/1/specjbb2015.jvm.log       6 mixed
10/genshen/specjbb2015/1/specjbb2015.jvm.log       7 interruptions
--
10/genshen/specjbb2015/2/specjbb2015.jvm.log     7 Completed Old GCs (2.50%)
10/genshen/specjbb2015/2/specjbb2015.jvm.log       7 mixed
10/genshen/specjbb2015/2/specjbb2015.jvm.log      14 interruptions
```

kdnilsen (approved):

src/hotspot/share/gc/shenandoah/shenandoahRegulatorThread.hpp
```diff
@@ -65,16 +65,26 @@ class ShenandoahRegulatorThread: public ConcurrentGCThread {
...
+  void regulate_young_and_global_cycles();
+  // Default behavior for other modes (single generation).
+  void regulate_global_cycles();
+
+  // These return true if a cycle was started.
   bool start_old_cycle();
   bool start_young_cycle();
   bool start_global_cycle();
 
   bool should_unload_classes();
```
ysramakrishna, line 80: Could you also please write a 1-line documentation comment for this as well?

src/hotspot/share/gc/shenandoah/shenandoahRegulatorThread.cpp
```diff
@@ -53,18 +53,18 @@ ShenandoahRegulatorThread::ShenandoahRegulatorThread(ShenandoahControlThread* co
...
   } else {
-    regulate_heap();
+    regulate_global_cycles();
   }
 
   log_info(gc)("%s: Done.", name());
 }
 
-void ShenandoahRegulatorThread::regulate_concurrent_cycles() {
+void ShenandoahRegulatorThread::regulate_young_and_old_cycles() {
```
ysramakrishna, line 67: I am wondering if it might not be simpler and clearer to write this method as:

```
...
while (!should_terminate()) {
  switch (mode) {
     case: SCT::none : {
        // no collection is in progress
        // 1. check and start a global collection if needed
        if (should_unload_classes() && start_global_cycle()) {
          log_info...
        } else if (start_old_cycle()) {
           log_info...
        } else if (start_young_cycle()) {
           log_info...
        }
        break;
     }
     case SCT::servicing_old: {
         if (start_young_cycle()) {
           log_info...
         }
         break;
      }
      default: // fall through
   }
   // sleep a tad before next check; sleep duration inversely proportional to allocation rate
   regulator_sleep();   // My worry is that heuristics says wait for old increment and we sleep longer than that? May be that's not an issue?
}
...
```

The rest of the checks etc. seem already to be done in the `start_...` methods from what I can tell.

I may have missed some subtlety though.

src/hotspot/share/gc/shenandoah/shenandoahRegulatorThread.cpp
```diff
@@ -168,21 +175,14 @@ bool ShenandoahRegulatorThread::request_concurrent_gc(ShenandoahGenerationType g
...
-
-bool ShenandoahRegulatorThread::start_global_cycle() {
-  return _global_heuristics->should_start_gc() && request_concurrent_gc(ShenandoahControlThread::select_global_generation());
-}
-
 void ShenandoahRegulatorThread::stop_service() {
   log_info(gc)("%s: Stop requested.", name());
 }
 
 bool ShenandoahRegulatorThread::should_unload_classes() {
```
ysramakrishna, line 182: I'd rename this to:

`should_start_metaspace_gc()`

to avoid confusion with similarly named method in other classes (heuristics) with a different semantics.

Relatedly, do we still need the second check in this method?

```
bool ShenandoahHeuristics::should_unload_classes() {
  if (!can_unload_classes()) return false;
  if (has_metaspace_oom()) return true;
  return ClassUnloadingWithConcurrentMark;
}
```

If not, it can be simplified to `return can_unload_classes() && ClassUnloadingWithConcurrentMark;`

I do believe, we should write a line documenting the intended semantics of `should_unload_classes()` in all these places because there ay otherwise be some confusion about what the truth modality "should" really means.

ysramakrishna (commented):
A few comments for your consideration.

src/hotspot/share/gc/shenandoah/shenandoahRegulatorThread.hpp
earthling-amzn: Done.

src/hotspot/share/gc/shenandoah/shenandoahRegulatorThread.cpp
earthling-amzn: When regulating young and old cycles (i.e., generational mode), the regulator overrides the global heuristics decision to start a global cycle. In other modes, whether or not to unload classes is decided _after_ a cycle is started. However, the generational mode can only unload classes in a _global_ cycle, so we use the oom in metaspace to _trigger_ a global cycle (even if the global heuristic would not otherwise trigger a cycle).

The other subtlety here is that old cycles are only started when a young cycle wants to run. Each old cycle is preceded by a young "bootstrap" cycle, so the young heuristic trigger is still satisfied. This was meant to avoid the situation where old cycles are triggered shortly after a young cycle - which runs an unproductive "bootstrap" cycle. This is why we check the young heuristic and then the old heuristic before starting an old cycle.

Your comment about starting a global cycle (to unload classes) even during an old cycle is interesting though. There's no specific reason why we couldn't interrupt and abandon the old cycle in this case, but I think that's a change for another PR.

src/hotspot/share/gc/shenandoah/shenandoahRegulatorThread.cpp
earthling-amzn: I'll add comments to `should_unload_classes`, but I don't want to change the implementation. This method is used by other Shenandoah modes. As I think on it, I'm going to make the regulator honor any class unloading restrictions before starting a fruitless cycle that won't unload anything.

src/hotspot/share/gc/shenandoah/shenandoahRegulatorThread.cpp
earthling-amzn: I'm not too worried about the regulator's sleep method. The new flag introduces a delay if an old cycle is running, but the delay to start a young cycle because the regulator thread was sleeping is unchanged. Note that the feature can also be disabled by setting `ShenandoahMinimumOldMarkTimeMs` to zero.

ysramakrishna: Thanks for the clarifications here and for the extra documentation comments. Yes, I'd missed the subtleties here, which are important to document. Thanks for that!

ysramakrishna (approved):
Thanks for the clarifying documentation comments & the changes.

Looks good to me! 🚢

PR #372, opened 2023-12-15
[GenShen] TestPeriodicGC fails with "Generational mode: Should have periodic young GC in logs"

ysramakrishna:
/summary

The test uncovered a flaw in `should_unload_classes()` such that it would answer `yes` even when there wasn't a need for a collection. This was being checked by the regulator thread as the condition to start a global collection in the case of generational shenandoah.

**Testing:**

- [ ] GHA
- [ ] Code pipeline testing
- [ ] TestPeriodicGC
- [ ] SpecJBB locally

/issue JDK-8322150


ysramakrishna:
Converting to a draft until all failures are resolved satisfactorily.

kdnilsen (approved):

ysramakrishna:
/open

preparatory to integrating soon.

shipilev:
> The test uncovered a flaw in should_unload_classes() such that it would answer yes even when there wasn't a need for a collection.

But that is the intended semantics of the method: it asks whether a _current cycle_ should unload classes, not whether we should _start_ a cycle to unload classes. In other words, `should_unload_classes()` answer should only be used to pick the flavor of the GC cycle that we have already decided to do, rather than deciding whether to trigger a GC. See the uses:

```
      // Ask policy if this cycle wants to process references or unload classes
      heap->set_unload_classes(heuristics->should_unload_classes());
```

If regulator uses this to trigger the cycle, then the bug is in the regulator.

ysramakrishna:
> > The test uncovered a flaw in should_unload_classes() such that it would answer yes even when there wasn't a need for a collection.
> 
> But that is the intended semantics of the method: it asks whether a _current cycle_ should unload classes, not whether we should _start_ a cycle to unload classes. In other words, `should_unload_classes()` answer should only be used to pick the flavor of the GC cycle that we have already decided to do, rather than deciding whether to trigger a GC. See the uses:
> 
> ```
>       // Ask policy if this cycle wants to process references or unload classes
>       heap->set_unload_classes(heuristics->should_unload_classes());
> ```
> 
> If regulator uses this to trigger the cycle, then the bug is in the regulator.

Thanks for clarifying, Aleksey! I'll document this, and change the PR to implement this correctly in GenShen.

Converting to a draft so this can be correctly treated in GenShen, per the above semantics.

earthling-amzn:
The regulator only calls `should_unload_classes` when running in `generational` mode.

PR #380, opened 2024-01-05
8323023: GenShen: Region logging test fails intermittently

earthling-amzn:
This test used to exercise a specific Shenandoah feature, but we've long since rewritten that feature to be based on unified logging. The test doesn't need to retest unified logging features.

---------
### Progress

### Issue
 * [JDK-8323023](https://bugs.openjdk.org/browse/JDK-8323023): GenShen: Region logging test fails intermittently (**Bug** - P4)

### Reviewers
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - Committer)
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [e28aca2b](https://git.openjdk.org/shenandoah/pull/380/files/e28aca2be45662f1a8583e00127943e48d1bf675)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/380/head:pull/380` \
`$ git checkout pull/380`

`$ git checkout pull/380` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/380/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 380`

`$ git pr show -t 380`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/380.diff">https://git.openjdk.org/shenandoah/pull/380.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/380#issuecomment-1877959354)


test/hotspot/jtreg/gc/shenandoah/TestShenandoahRegionLogging.java
```diff
@@ -0,0 +1,57 @@
...
+
+public class TestShenandoahRegionLogging {
+    private static final long TARGET_MB = Long.getLong("target", 1);
+    private static Object sink;
+
+    public static void main(String[] args) throws Exception {
+        long count = TARGET_MB * 1024 * 1024 / 16;
+        for (long c = 0; c < count; c++) {
+            sink = new Object();
+            Thread.sleep(1);
```
shipilev, line 47: Pre-existing, but: The problem with doing this sleep is that we are doing about 64K of 1ms sleeps in this tests. Even if we manage to do 1ms sleep, that is already 65 seconds spent in this test. But we might as well spend single-digit ms per sleep. For 10ms sleeps, that is 10 minutes for a test, which would be playing it close to timeout.

So I wonder if this test should be time-bound.

shipilev (approved):
Looks okay, but there is a pre-existing issue, I think.

kdnilsen (approved):

test/hotspot/jtreg/gc/shenandoah/TestShenandoahRegionLogging.java
kdnilsen: Would it be better to reduce the frequency of the sleep, to one in 10, for example?

earthling-amzn: The test is really just waiting for region sampling to write to the logs. I think one call to `System.gc` would suffice. I'll see if I can reduce the runtime for this test.

shipilev (approved):
Looks good, as long the test runs well :) Might want to try with JTREG=REPEAT_COUNT=100 or something.

ysramakrishna (approved):
Looks good. Should the test additionally check the contents of one of the files to find a line describing a region snapshot?

PR #382, opened 2024-01-12
8323630: GenShen: Control thread may (still) ignore requests to start concurrent GC

earthling-amzn:
A race condition exists in which the control thread may clear the `_requested_gc_cause` immediately after a mutator requests an explicit gc. When this happens, the control thread will no longer accept requests from the regulator to start concurrent GC cycles. The mutator thread will never wakeup, eventually the application will run out of memory or no progress will be made.

The change here is intended to simplify the thread communication protocol by reducing the number of variables in play.

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
@@ -62,7 +62,7 @@ ShenandoahControlThread::ShenandoahControlThread() :
   _control_lock(Mutex::nosafepoint - 2, "ShenandoahControlGC_lock", true),
   _regulator_lock(Mutex::nosafepoint - 2, "ShenandoahRegulatorGC_lock", true),
   _periodic_task(this),
-  _requested_gc_cause(GCCause::_no_cause_specified),
+  _requested_gc_cause(GCCause::_no_gc),
   _requested_generation(select_global_generation()),
   _degen_point(ShenandoahGC::_degenerated_outside_cycle),
   _degen_generation(nullptr),
@@ -117,8 +117,7 @@ void ShenandoahControlThread::run_service() {
     // Figure out if we have pending requests.
     bool alloc_failure_pending = _alloc_failure_gc.is_set();
     bool humongous_alloc_failure_pending = _humongous_alloc_failure_gc.is_set();
-    GCCause::Cause cause = _requested_gc_cause;
-    _requested_gc_cause = GCCause::_no_gc;
+    GCCause::Cause cause = Atomic::xchg(&_requested_gc_cause, GCCause::_no_gc);
 
     bool explicit_gc_requested = is_explicit_gc(cause);
     bool implicit_gc_requested = is_implicit_gc(cause);
@@ -866,7 +865,12 @@ bool ShenandoahControlThread::request_concurrent_gc(ShenandoahGenerationType gen
   }
 
   if (gc_mode() == none) {
-    _requested_gc_cause = GCCause::_shenandoah_concurrent_gc;
+    GCCause::Cause existing = Atomic::cmpxchg(&_requested_gc_cause, GCCause::_no_gc, GCCause::_shenandoah_concurrent_gc);
+    if (existing != GCCause::_no_gc) {
+      log_debug(gc, thread)("Reject request for concurrent gc because another gc is pending: %s", GCCause::to_string(existing));
+      return false;
+    }
+
     _requested_generation = generation;
     notify_control_thread();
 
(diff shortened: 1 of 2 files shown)


src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
```diff
@@ -118,10 +117,10 @@ void ShenandoahControlThread::run_service() {
...
     bool alloc_failure_pending = _alloc_failure_gc.is_set();
     bool humongous_alloc_failure_pending = _humongous_alloc_failure_gc.is_set();
-    bool is_gc_requested = _gc_requested.is_set();
-    GCCause::Cause requested_gc_cause = _requested_gc_cause;
-    bool explicit_gc_requested = is_gc_requested && is_explicit_gc(requested_gc_cause);
-    bool implicit_gc_requested = is_gc_requested && is_implicit_gc(requested_gc_cause);
+    GCCause::Cause cause = Atomic::xchg(&_requested_gc_cause, GCCause::_no_gc);
+
+    bool explicit_gc_requested = is_explicit_gc(cause);
+    bool implicit_gc_requested = is_implicit_gc(cause);
```
ysramakrishna, line 123: const them all.

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
```diff
@@ -254,16 +244,16 @@ void ShenandoahControlThread::run_service() {
       }
     }
 
-    // Blow all soft references on this cycle, if handling allocation failure,
-    // either implicit or explicit GC request, or we are requested to do so unconditionally.
-    if (generation == select_global_generation() && (alloc_failure_pending || implicit_gc_requested || explicit_gc_requested || ShenandoahAlwaysClearSoftRefs)) {
-      heap->soft_ref_policy()->set_should_clear_all_soft_refs(true);
-    }
-
     bool gc_requested = (gc_mode() != none);
```
ysramakrishna, line 247: const the variable.

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
```diff
@@ -118,10 +117,10 @@ void ShenandoahControlThread::run_service() {
     // Figure out if we have pending requests.
     bool alloc_failure_pending = _alloc_failure_gc.is_set();
     bool humongous_alloc_failure_pending = _humongous_alloc_failure_gc.is_set();
```
ysramakrishna, line 119: const them.

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
```diff
@@ -254,16 +244,16 @@ void ShenandoahControlThread::run_service() {
...
-    }
-
     bool gc_requested = (gc_mode() != none);
-    assert (!gc_requested || cause != GCCause::_last_gc_cause, "GC cause should be set");
+    assert (!gc_requested || cause != GCCause::_no_gc, "GC cause should be set");
 
     if (gc_requested) {
+      // Blow away all soft references on this cycle, if handling allocation failure,
+      // either implicit or explicit GC request, or we are requested to do so unconditionally.
+      if (generation == select_global_generation() && (alloc_failure_pending || implicit_gc_requested || explicit_gc_requested || ShenandoahAlwaysClearSoftRefs)) {
```
ysramakrishna, line 253: Not a change you made, but why do we clear all soft refs if it's an implicit gc? Is memory pressure supposed to be very high here?

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
```diff
@@ -395,8 +385,8 @@ void ShenandoahControlThread::run_service() {
       last_shrink_time = current;
     }
 
-    // Don't wait around if there was an allocation failure - start the next cycle immediately.
-    if (!is_alloc_failure_gc()) {
+    // Don't wait if there was an allocation failure or another request was made mid-cycle.
```
ysramakrishna, line 388: I'd rewrite the comment as "Wait for ShenandoahControlIntervalMax, unless there was an allocation failure or ..."

ysramakrishna (approved):
Looks good; some minor comments & suggestions. I wonder if the xchg and cmpxchg couldn't just be replaced with a condition variable protected by a lock, and that might simplify the code some more. But this looks fine.

kdnilsen (approved):

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
ysramakrishna: Thanks!

ysramakrishna (approved):
Re-reviewed!
🚢

PR #398, opened 2024-02-16
8325808: GenShen: Move generational mode code out of shFullGC.cpp

earthling-amzn:
This change reduces the differences from the upstream branch by moving large chunks of generational mode code into separate files.

---------
### Progress

### Issue
 * [JDK-8325808](https://bugs.openjdk.org/browse/JDK-8325808): GenShen: Move generational mode code out of shFullGC.cpp (**Task** - P4)

### Reviewers
 * [Y. Srinivas Ramakrishna](https://openjdk.org/census#ysr) (@ysramakrishna - Committer) ⚠️ Review applies to [cfed36ea](https://git.openjdk.org/shenandoah/pull/398/files/cfed36ea88a8f0dc476752f5b64644937ab2d35a)
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [cfed36ea](https://git.openjdk.org/shenandoah/pull/398/files/cfed36ea88a8f0dc476752f5b64644937ab2d35a)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/398/head:pull/398` \
`$ git checkout pull/398`

`$ git checkout pull/398` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/398/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 398`

`$ git pr show -t 398`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/398.diff">https://git.openjdk.org/shenandoah/pull/398.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/398#issuecomment-1949224150)


src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
```diff
@@ -1118,10 +770,9 @@ void ShenandoahFullGC::phase2_calculate_target_addresses(ShenandoahHeapRegionSet
 
     distribute_slices(worker_slices);
 
-    size_t num_workers = heap->max_workers();
-
+    // TODO: This is ResourceMark is missing upstream.
```
ysramakrishna, line 773: May be roll it into the refactoring PR you have open and under review at https://github.com/openjdk/jdk/pull/17894/files#diff-d58d0fe2415e3fa64f687435baf73227fb4e3e9eee8ba52ba1f3bb8301437106 ?

src/hotspot/share/gc/shenandoah/shenandoahGenerationalFullGC.hpp
```diff
@@ -0,0 +1,83 @@
...
+public:
+  static void prepare(ShenandoahHeap* heap);
+  static void handle_completion(ShenandoahHeap* heap);
+  static void rebuild_remembered_set(ShenandoahHeap* heap);
+  static void balance_old_generation(ShenandoahHeap* heap);
+  static void balance_generations(ShenandoahHeap* heap);
+  static void log_live_in_old(ShenandoahHeap* heap);
+  static void account_for_region(ShenandoahHeapRegion* r, size_t &region_count, size_t &region_usage, size_t &humongous_waste);
+  static void restore_top_before_promote(ShenandoahHeap* heap);
+  static void maybe_coalesce_and_fill_region(ShenandoahHeapRegion* r);
```
ysramakrishna, line 46: Please provide a 1-line documentation for each of these public API methods.

src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
```diff
@@ -775,56 +447,39 @@ void ShenandoahPrepareForCompactionTask::work(uint worker_id) {
...
+  }
+}
 
-      // Compacted the region to somewhere else? From-region is empty then.
-      if (!cl.is_compact_same_region()) {
-        empty_regions.append(from_region);
-      }
-      from_region = it.next();
+template<typename ClosureType>
+void ShenandoahPrepareForCompactionTask::prepare_for_compaction(ClosureType& cl,
```
ysramakrishna, line 460: Looking at this method, it looks like it actually belongs as a public method in the closure, with that closure's methods invoked here becoming private to the closure. Makes for a narrower public API all around and keeps the loop in the right place.

src/hotspot/share/gc/shenandoah/shenandoahGenerationalFullGC.hpp
```diff
@@ -0,0 +1,83 @@
...
+};
+
+class ShenandoahPrepareForGenerationalCompactionObjectClosure : public ObjectClosure {
```
ysramakrishna, line 49: See comment in `ShenandoahPrepareForCompactionTask::prepare_for_compaction()`. I think this closure should include `prepare_for_compaction` in its public API, and full gc should just invoke that method of the closure directly.

If this simplification works out, then it would also affect the sister upstream PR https://github.com/openjdk/jdk/pull/17894 similarly.

ysramakrishna (commented):
Looks good, but left a few suggestions, one of which (if the suggested slight simplification were to work out), would affect the sister PR as well.

Otherwise looks great; the refactoring and pulling the work out of line into its separate class like you've done here should make upstreaming easier. Thanks!

src/hotspot/share/gc/shenandoah/shenandoahGenerationalFullGC.hpp
ysramakrishna: Left a link there as well: https://github.com/openjdk/jdk/pull/17894#discussion_r1494924914

src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
```diff
@@ -276,6 +180,7 @@ void ShenandoahFullGC::do_it(GCCause::Cause gc_cause) {
 
     // b. Cancel all concurrent marks, if in progress
     if (heap->is_concurrent_mark_in_progress()) {
+      // TODO: Send cancel_concurrent_mark upstream? Does it really not have it already?
```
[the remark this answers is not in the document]
earthling-amzn: We changed the behavior of cancellation somewhat for the generational mode. We essentially overloaded 'cancellation' to mean 'suspension' for old gen marking. I agree it needs to be reconciled, but I think it's outside the scope of this PR.

ysramakrishna (approved):
Changes look great; thanks for the clearer renaming as well as the great documentation!

src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
ysramakrishna: Yes, I think you are right. When I was writing that comment, I was thinking that the loop could just be duplicated into each of the two closures. I guess may be that doesn't particularly make it any simpler and moreover leads to some avoidable duplication. I'll leave it up to you as to which makes more sense to you.

Reviewed and approved!

kdnilsen (approved):
Thanks for this work and thanks Ramki for thorough review.

src/hotspot/share/gc/shenandoah/shenandoahFullGC.cpp
earthling-amzn: I'll leave it for now. This form also reduces some of the delta with upstream (that is, with fewer changes upstream).

PR #407, opened 2024-03-15
8328235: GenShen: Robustify ShenandoahGCSession and fix missing use

ysramakrishna:
/issue JDK-8328235

ShenandoahGCSession is intended to create a scope where the ShenandoahHeap's _gc_cause and _gc_generation field reflect the current gc cycle. We now check that we do not overwrite existing non-default settings (respectively _no_gc and nullptr). The destructor of the scope/stack object also resets these fields to their default settings, ensuring intended uses. This uncovered a situation where the scope was not entered when it should have been, which we have now fixed.

A case of flickering of active_generation() was identified when used concurrently by mutators while it was being modified by the controller thread. To deal with this, we have carefully gone through the setting and use of the field, and found that an expedient fix for the race is to split the field into two:
- _gc_generation is set & cleared by the controller thread whenever it enters and exits a GC scope, and services concurrent gc cycles for young or old generations.
- _active_generation is set to the value in _gc_generation at the start of each Shenandoah GC safepoint operation so that mutator threads and load barriers always see a consistent value between safepoints.

Asserts check the protocol for setting and clearing these fields.

The protocol for use of the fields is that mutator threads may never use the _gc_generation field since it's subject to asynchronously changing based on actions of the coordinator thread. Mutator threads may only use the _active_generation field which changes synchronously at safepoints. Worker threads will generally use the former, but they may also use the latter as part of the load barrier.

An alternative approach would be to not use a global variable for the _gc_generation indirected through the heap, but rather to pass it into the gc closures that do the work for specific phases of the GC that need to know which generation is currently subject to collector actions. This would work as well, but the changes would potentially touch more code. We would still have to have set the variable that is consulted by the load barriers, viz. _active_generation, in a mutator-safe fashion at a safepoint, like we do today. This or other alternative approaches may be investigated in the future to potentially make this protocol more self-contained and robust rather than leaking as it does today into many places in the code.

*Testing*:
- [x] code pipeline
- [x] specjbb testing
- [x] specjbb performance
- [x] jtreg:hotspot_gc and jtreg:hotspot:tier1 w/fastdebug
- [x] GHA

*Performance Notes:*
Code pipeline performance tests do not flag any specific changes in performance. This includes both the standard and the non-standard "weak ref patch" variant of SPECjbb configured there (8 cores, 8 gb heap). The changes would generally be expected to be largely performance neutral as indicated by the code pipeline performance numbers.

Somewhat surprisingly (in a pleasant way), however, we have noted a doubling (or better) in the SPECjbb scores (both critical and max) with our change on a larger machine (40 cores, 10 gb heap with 8 gb young gen) with GenShen. It is suspected that the correcting of reference processing behavior because of synchronous updating of the LRB's active generation parameter has reduced the incidence of degenerated collections early during ramp-up that would otherwise throttle the max and critical jops load because of the longer pauses that the fix avoids. This is indicated in the almost 10x larger reference processing times noted with the baseline build. While we do not have a full explanation of this improvement, we believe the change should go in as it fixes a correctness/robustness issue, does not regress performance with any benchmarks, and indeed improves specjbb scores on this setup.

```

BASELINE:
results_reference_default_1/composite.out:RUN RESULT: hbIR (max attempted) = 34282, hbIR (settled) = 28585, max-jOPS = 5828, critical-jOPS = 5828
results_reference_default_2/composite.out:RUN RESULT: hbIR (max attempted) = 34282, hbIR (settled) = 28585, max-jOPS = 8228, critical-jOPS = 8228
results_reference_default_3/composite.out:RUN RESULT: hbIR (max attempted) = 34282, hbIR (settled) = 28585, max-jOPS = 7885, critical-jOPS = 7656
results_reference_default_4/composite.out:RUN RESULT: hbIR (max attempted) = 34282, hbIR (settled) = 28585, max-jOPS = 7199, critical-jOPS = 7199
results_reference_default_5/composite.out:RUN RESULT: hbIR (max attempted) = 34282, hbIR (settled) = 28585, max-jOPS = 11999, critical-jOPS = 11672

SPECIMEN:
results_specimen_default_1/composite.out:RUN RESULT: hbIR (max attempted) = 34282, hbIR (settled) = 28585, max-jOPS = 29140, critical-jOPS = 27411
results_specimen_default_2/composite.out:RUN RESULT: hbIR (max attempted) = 34282, hbIR (settled) = 29477, max-jOPS = 28111, critical-jOPS = 26868
results_specimen_default_3/composite.out:RUN RESULT: hbIR (max attempted) = 34282, hbIR (settled) = 28585, max-jOPS = 29825, critical-jOPS = 28166
results_specimen_default_4/composite.out:RUN RESULT: hbIR (max attempted) = 34282, hbIR (settled) = 28585, max-jOPS = 27083, critical-jOPS = 26006
results_specimen_default_5/composite.out:RUN RESULT: hbIR (max attempted) = 34282, hbIR (settled) = 29204, max-jOPS = 28797, critical-jOPS = 27213

```

---------
### Progress

### Issue
 * [JDK-8328235](https://bugs.openjdk.org/browse/JDK-8328235): GenShen: Robustify ShenandoahGCSession and fix missing use (**Sub-task** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/407/head:pull/407` \
`$ git checkout pull/407`

`$ git checkout pull/407` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/407/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 407`

`$ git pr show -t 407`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/407.diff">https://git.openjdk.org/shenandoah/pull/407.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/407#issuecomment-1998777459)


ysramakrishna:
Code pipeline testing revealed a further scenario when `active_generation()` was used outside of what should have been a gc session (temporal) scope. I'll revert this to a draft until I fix that issue as well.

ysramakrishna:
A case of flickering of active_generation() was identified, and found to be benign. An assert now checks for this situation. The code has been made robust wrt the flickering.

ysramakrishna:
> A case of flickering of active_generation() was identified, and found to be benign. An assert now checks for this situation. The code has been made robust wrt the flickering.

Testing revealed an issue with this PR, in particular with the clearing of active_generation() via Scopes -- this needs closer examination, so I'll once again take this back to draft. Sorry for the, errr, flickering in this PR :-)

src/hotspot/share/gc/shenandoah/shenandoahSTWMark.cpp
```diff
@@ -104,7 +104,8 @@ void ShenandoahSTWMark::mark() {
   ShenandoahCodeRoots::arm_nmethods_for_mark();
 
   // Weak reference processing
-  ShenandoahReferenceProcessor* rp = heap->active_generation()->ref_processor();
+  ShenandoahReferenceProcessor* rp = heap->gc_generation()->ref_processor();
+  assert(heap->gc_generation() == heap->active_generation(), "Should be identical at stw phases");
```
ysramakrishna, line 108: `assert_generations_reconciled();` instead.

src/hotspot/share/gc/shenandoah/shenandoahSTWMark.cpp
```diff
@@ -172,7 +173,9 @@ void ShenandoahSTWMark::mark_roots(uint worker_id) {
 void ShenandoahSTWMark::finish_mark(uint worker_id) {
   ShenandoahPhaseTimings::Phase phase = _full_gc ? ShenandoahPhaseTimings::full_gc_mark : ShenandoahPhaseTimings::degen_gc_stw_mark;
   ShenandoahWorkerTimingsTracker timer(phase, ShenandoahPhaseTimings::ParallelMark, worker_id);
-  ShenandoahReferenceProcessor* rp = ShenandoahHeap::heap()->active_generation()->ref_processor();
+  ShenandoahReferenceProcessor* rp = ShenandoahHeap::heap()->gc_generation()->ref_processor();
+  assert(ShenandoahHeap::heap()->gc_generation() == ShenandoahHeap::heap()->active_generation(),
```
ysramakrishna, line 177: `assert_generations_reconciled();` instead.

test/hotspot/jtreg/gc/shenandoah/oom/TestThreadFailure.java
```diff
@@ -78,6 +78,7 @@ public static void main(String[] args) throws Exception {
         {
             ProcessBuilder pb = ProcessTools.createLimitedTestJavaProcessBuilder(
                     "-Xmx32m",
+                    // "-XX:+UnlockExperimentalVMOptions", "-XX:ShenandoahNoProgressThreshold=12", "-XX:+ShowMessageBoxOnError",
```
ysramakrishna, line 81: Ignore this. Will be removed; debugging detritus left in place until we believe we are ready.

earthling-amzn (changes requested):
Left a few minor coding suggestings. On a higher level, it's not clear to me when and which threads should use `gc_generation` or `active_generation`. I gather that active_generation is set on the safepoint and `gc_generation` is set by control thread, does that mean any code before `init-mark` needs to use `gc_generation`?

ysramakrishna:
> Left a few minor coding suggestings. On a higher level, it's not clear to me when and which threads should use `gc_generation` or `active_generation`. I gather that active_generation is set on the safepoint and `gc_generation` is set by control thread, does that mean any code before `init-mark` needs to use `gc_generation`?

Thanks for the review. I'll make the suggested changes and respond individually to the feedback.

As regards, `gc_generation` vs `active_generation`, the intention is that only the load-reference-barrier and the mutator threads use `active_generation` and the GC worker threads use `gc_generation` (unless executing the load-reference-barrier). The idea is that the LRB's notion of an active generation cannot be updated safely except at a safepoint. I suspect this splitting of roles still presents a problem when evacuations occur during mixed collections. So there is likely more work to do to clean up the two roles.

More generally, `gc_generation` is tightly controlled by scopes by the coordinator thread, and indicate the generation that GC worker threads are dealing with at any time, except during evacuation/update-refs when the load-reference-barrier uses the active_generation.

I suspect there is still some badness or mix-ups in these LRB paths that is leading to issues that need to be resolved (there are some failures since a recent merge that I am yet to resolve).

Once I am done with those, I hope to have a cleaner solution.

Meanwhile, I am going to go over your suggestions and make the changes stemming from them.

src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
```diff
@@ -928,6 +928,7 @@ void ShenandoahEvacUpdateCleanupOopStorageRootsClosure::do_oop(oop* p) {
   const oop obj = RawAccess<>::oop_load(p);
   if (!CompressedOops::is_null(obj)) {
     if (!_mark_context->is_marked(obj)) {
+      _heap->assert_generations_reconciled();
```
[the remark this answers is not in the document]
ysramakrishna: Done.

src/hotspot/share/gc/shenandoah/shenandoahGenerationalHeap.cpp
```diff
@@ -790,7 +793,9 @@ class ShenandoahGenerationalUpdateHeapRefsTask : public WorkerTask {
 
     ShenandoahHeapRegion* r = _regions->next();
     // We update references for global, old, and young collections.
-    assert(_heap->active_generation()->is_mark_complete(), "Expected complete marking");
+    ShenandoahGeneration* const gc_generation = _heap->gc_generation();
```
earthling-amzn, line 796: It would certainly involve more changes, but we could pass the correct generation from `shConcurrentGC` in this case. I think we could do this in most cases (save for the barriers, but you've established a safe pattern for them to access the  active generation already).

test/hotspot/jtreg/gc/shenandoah/oom/TestThreadFailure.java
ysramakrishna: Removed in latest. Resolving.

src/hotspot/share/gc/shenandoah/shenandoahGenerationalEvacuationTask.cpp
```diff
@@ -131,7 +131,8 @@ void ShenandoahGenerationalEvacuationTask::promote_in_place(ShenandoahHeapRegion
 
   {
     const size_t old_garbage_threshold = (ShenandoahHeapRegion::region_size_bytes() * ShenandoahOldGarbageThreshold) / 100;
-    assert(_heap->active_generation()->is_mark_complete(), "sanity");
+    assert(_heap->gc_generation()->is_mark_complete(), "sanity");
```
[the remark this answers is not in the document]
ysramakrishna: Done.

src/hotspot/share/gc/shenandoah/shenandoahGenerationalHeap.cpp
ysramakrishna: Let'd defer that cleanup for later.

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
```diff
@@ -360,31 +360,39 @@ bool ShenandoahHeap::is_in(const void* p) const {
...
+    // no collection is happening: only expect this to be called
     // when concurrent processing is active, but that could change
     return false;
   }
 
   assert(is_in(obj), "only check if is in active generation for objects (" PTR_FORMAT ") in heap", p2i(obj));
-  assert((active_generation() == (ShenandoahGeneration*) old_generation()) ||
-         (active_generation() == (ShenandoahGeneration*) young_generation()) ||
-         (active_generation() == global_generation()), "Active generation must be old, young, or global");
+  assert(gen == (ShenandoahGeneration*)old_generation() ||
```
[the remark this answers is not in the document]
ysramakrishna: Done... Seems more patently tautological now in the new form, but I'll let it stay in this form for now. Let me know if it can or should be removed entirely.

earthling-amzn (approved):

ysramakrishna:
Thanks for your review, William!

/integrate

PR #406, opened 2024-03-14
8328220: GenShen: Move generational mode operational parameters out of ShenandoahHeap

earthling-amzn:
* Many fields previously defined in `ShenandoahHeap` have been moved to `ShenandoahOldGeneration`.
* Generational specific serviceability has been moved to `ShenandoahGenerationalHeap`.
* Methods for sizing the old generation have been moved to `ShenandoahGenerationalHeap`.
* `ShenandoahGenerationalHeap::heap()` now asserts that the generational mode is active.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (1 review required, with at least 1 [Committer](https://openjdk.org/bylaws#committer))

### Issue
 * [JDK-8328220](https://bugs.openjdk.org/browse/JDK-8328220): GenShen: Move generational mode operational parameters out of ShenandoahHeap (**Task** - P4)

### Reviewers
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [00f8778b](https://git.openjdk.org/shenandoah/pull/406/files/00f8778b6e0c576710f850b17f287e54e8c53c1c)
 * [Y. Srinivas Ramakrishna](https://openjdk.org/census#ysr) (@ysramakrishna - Committer)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/406/head:pull/406` \
`$ git checkout pull/406`

Update a local copy of the PR: \
`$ git checkout pull/406` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/406/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 406`

View PR using the GUI difftool: \
`$ git pr show -t 406`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/shenandoah/pull/406.diff">https://git.openjdk.org/shenandoah/pull/406.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/406#issuecomment-1998539820)

PR #410, opened 2024-03-20
8328626: GenShen: Combine old generation surplus/deficit fields into a single balance field

earthling-amzn:
This is a follow up to https://github.com/openjdk/shenandoah/pull/406.


src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -1163,76 +1163,78 @@ void ShenandoahFreeSet::prepare_to_rebuild(size_t &young_cset_regions, size_t &o
...
   }
 
-  // Evac reserve: reserve trailing space for evacuations, with regions reserved for old evacuations placed to the right
-  // of regions reserved of young evacuations.
-  if (!_heap->mode()->is_generational()) {
-    young_reserve = (_heap->max_capacity() / 100) * ShenandoahEvacReserve;
-    old_reserve = 0;
+  // All allocations taken from the old collector set are performed by GC, generally using PLABs for both
+  // promotions and evacuations.  The partition between which old memory is reserved for evacuation and
+  // which is reserved for promotion is enforced using thread-local variables that prescribe intentons for
```
kdnilsen, line 1220: typo: intentions

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -1163,76 +1163,78 @@ void ShenandoahFreeSet::prepare_to_rebuild(size_t &young_cset_regions, size_t &o
...
+
+  reserve_regions(young_reserve, old_reserve);
+  _free_sets.establish_alloc_bias(OldCollector);
+  _free_sets.assert_bounds();
+  log_status();
+}
+
+void ShenandoahFreeSet::compute_young_and_old_reserves(size_t young_cset_regions, size_t old_cset_regions,
+                                                       bool have_evacuation_reserves, size_t* young_reserve_result,
+                                                       size_t* old_reserve_result) const {
```
kdnilsen, line 1185: Might read more "naturally" if args declared as:
```
size_t& young_reserve_result, size_t& old_reserve_result

File: src/hotspot/share/gc/shenandoah/shenandoahGenerationalHeap.cpp [hunk 3]
```diff
@@ -121,19 +121,19 @@ ShenandoahGenerationalHeap::TransferResult ShenandoahGenerationalHeap::balance_g
...
-  const size_t old_region_deficit = old_gen->get_region_deficit();
-  old_gen->set_region_surplus(0);
-  old_gen->set_region_deficit(0);
+  const ssize_t old_region_balance = old_gen->get_region_balance();
+  old_gen->set_region_balance(0);
 
-  if (old_region_surplus) {
-    bool success = generation_sizer()->transfer_to_young(old_region_surplus);
+  if (old_region_balance > 0) {
+    const auto old_region_surplus = checked_cast<size_t>(old_region_balance);
```
kdnilsen, line 128: Just curious: what happens with a checked_cast<size_t> if old_region_balance < 0.  (I see that's impossible here, but what is the point of declaring checked_cast?)

kdnilsen (approved):

src/hotspot/share/gc/shenandoah/shenandoahGenerationalHeap.cpp
```diff
@@ -218,34 +218,29 @@ void ShenandoahGenerationalHeap::compute_old_generation_balance(size_t old_xfer_
...
     // The old region deficit (rounded up) will come from young
-    old_region_deficit = (old_need + region_size_bytes - 1) / region_size_bytes;
+    size_t old_region_deficit = (old_need + region_size_bytes - 1) / region_size_bytes;
 
     // Round down the regions we can transfer from young to old. If we're running short
     // on young-gen memory, we restrict the xfer. Old-gen collection activities will be
     // curtailed if the budget is restricted.
-    const size_t max_old_region_xfer = old_xfer_limit / region_size_bytes;
     old_region_deficit = MIN2(old_region_deficit, max_old_region_xfer);
+    old_generation()->set_region_balance(0 - checked_cast<ssize_t>(old_region_deficit));
```
ysramakrishna, line 242: The code in the if and the else branches is highly isomorphic. I have the feeling you may be able to fold them together, but I am not sure how clean it will be. Also if you start with `ssize_t` for some of the operands that feed into the surplus calculation (or balance if you will, but we can call it surplus wit the understanding that a negative surplus is a deficit), then you may be able to avoid some of the later casts and the signing of unsigned quantities, including the `(cast)x`, `-(cast)x` or `0-(cast)x`constructs above.

src/hotspot/share/gc/shenandoah/shenandoahGenerationalHeap.cpp
```diff
@@ -121,19 +121,19 @@ ShenandoahGenerationalHeap::TransferResult ShenandoahGenerationalHeap::balance_g
...
+    const auto old_region_surplus = checked_cast<size_t>(old_region_balance);
+    const bool success = generation_sizer()->transfer_to_young(old_region_surplus);
     return TransferResult {
       success, old_region_surplus, "young"
     };
   }
 
-  if (old_region_deficit) {
+  if (old_region_balance < 0) {
+    const auto old_region_deficit = checked_cast<size_t>(-old_region_balance);
```
ysramakrishna, line 136: Like I remarked elsewhere, the code in the if > 0 and if < 0 branches (or if you will, the if and else branches) are largely isomorphic, and can probably be folded together by calling a generic `transfer_from_old(ssize_t num_regions)` on ShenandoahGenerationSizer. The latter new method would be a folding together of the old methods `transfer_to_young()` and `transfer_to_old()`, and would return the `TransferResult` that this method `balance_generations` returns to its caller.

ysramakrishna (approved):
A few suggestions that might lead to further simplifications, it would seem. Let me know if that is not the case for some reason.

The rest of the changes look good to me, so I am approving the changes, with the suggested changes as something to look into.

If those suggested changes should work out, I'll be happy to do a rereview.

Thanks for the cleanup & refactoring of the code; it feels more compact, and easier to understand and to maintain.

src/hotspot/share/gc/shenandoah/shenandoahGenerationalHeap.cpp
earthling-amzn: `checked_cast` asserts that the cast did not change the value, defined here: https://github.com/openjdk/shenandoah/blob/master/src/hotspot/share/utilities/checkedCast.hpp#L39. If `old_region_balance` were less than zero, it would assert out. It will also assert out if `old_region_balance` were greater than `SSIZE_MAX`. Which is unlikely, but there is no harm in checking.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
earthling-amzn: Hmm, I usually see 'result' parameters like this passed as pointers in the JDK. It makes it clearer at the call site. Or did you mean to format them on the same line?

src/hotspot/share/gc/shenandoah/shenandoahGenerationalHeap.cpp
earthling-amzn: I agree the code in the branches is similar, but the APIs we have for managing capacity for a generation will ultimately make us chose old or young for the source or destination of the transfer based on whether `old_region_balance` is positive or negative.

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.hpp
```diff
@@ -125,7 +125,11 @@ class ShenandoahGenerationSizer {
   // given the number of heap regions depending on the kind of sizing algorithm.
   void recalculate_min_max_young_length(size_t heap_region_count);
 
-public:
+  bool transfer_regions(ShenandoahGeneration* src, ShenandoahGeneration* dst, size_t regions) const;
```
ysramakrishna, line 128: Transfer regions has a bit of nuance, in that it fails without affecting the transfer if the transfer won't succeed without violating source or destination size bounds. I'd explain when it returns true and when false in a documentation line here.

src/hotspot/share/gc/shenandoah/shenandoahGenerationalHeap.cpp
ysramakrishna: Yes, I was thinking that the generation sizer could do that based on the sign of the balance.
But I realize that might not necessarily simplify the code, as you seem to have discovered. OK, thanks for checking.

ysramakrishna (approved):
LGTM; thanks!

PR #414, opened 2024-04-02
Fix coalesce and fill

kdnilsen:
It appears that there was an error in our c&f code that mishandled the case when a filll object bumps up against top, and top aligns with the start of a remembered set card.

---------
### Progress
- [ ] Commit message must refer to an issue

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/414/head:pull/414` \
`$ git checkout pull/414`

`$ git checkout pull/414` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/414/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 414`

`$ git pr show -t 414`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/414.diff">https://git.openjdk.org/shenandoah/pull/414.diff</a>

</details>


kdnilsen:
FWIW, I have confirmed that the "problem" workload we have been debugging does emit the message:
  "KELVIN CONFIRMS PROBLEM"

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
```diff
@@ -293,6 +376,8 @@ ShenandoahCardCluster<RememberedSet>::block_start(const size_t card_index) const
     assert(oopDesc::is_oop(cast_to_oop(left)), "Should be an object");
     return left;
   }
+  // If memory has not yet been allocated within card card_index, the above test will fail, but it still may be the
```
ysramakrishna, line 379: One assumption here is that we ask this only of a dirty card. (May be we can assert that as a pre-condition?) Is it ever the case that we have a dirty card but in which no object has been allocated yet? When does that happen?

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
```diff
@@ -312,8 +397,16 @@ ShenandoahCardCluster<RememberedSet>::block_start(const size_t card_index) const
   size_t offset = get_last_start(cur_index);
   // can avoid call via card size arithmetic below instead
   p = _rs->addr_for_card_index(cur_index) + offset;
-  // Recall that we already dealt with the co-initial object case above
+  // In the case that no object has been allocated in this card, we may not have dealt with co-initial object case above.
```
ysramakrishna, line 400: See comment above. I suppose we should examine the allocation paths to see when objects are allocated in cards and when cards dirtied.

ysramakrishna (commented):
A question/comment...

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
kdnilsen: That's a good question.  I'm not sure we're to the bottom of this yet.  It may be that this was a problem of my own creation, resulting from queries introduced by my "enhanced debugging instrumentation".  But I don't think that's the entire problem, because we have been seeing assert failures at the bottom of block_start() in our standard regression testing.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
```diff
@@ -367,6 +395,32 @@ void ShenandoahOldGeneration::prepare_regions_and_collection_set(bool concurrent
     _old_heuristics->prepare_for_old_collections();
   }
 
+  log_info(gc)("After choosing global collection set, mixed candidates: " UINT32_FORMAT
```
kdnilsen, line 398: Place this change in "Improve logging"

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
```diff
@@ -225,15 +238,36 @@ ShenandoahCardCluster<RememberedSet>::coalesce_objects(HeapWord* address, size_t
     if (get_last_start(card_at_start) > coalesced_offset) {
       // Existing last start is being coalesced, create new last start
       set_last_start(card_at_start, coalesced_offset);
+    } else {
```
kdnilsen, line 241: Place these changes in "Refine remset scan assertions"

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
```diff
@@ -312,8 +397,16 @@ ShenandoahCardCluster<RememberedSet>::block_start(const size_t card_index) const
...
+  // In the case that no object has been allocated in this card, we may not have dealt with co-initial object case above.
   assert(p < left, "obj should start before left");
+  obj = cast_to_oop(p);
+  if (p + obj->size() == left) {
+#ifdef KELVIN_DEBUG_DEEPER
+    log_info(gc)("KELVIN CONFIRMS PROBLEM");
+#endif
+    return left;
+  }
+
```
kdnilsen, line 409: Make this change in "fix block_start", or add an assert to prevent this use case in invocations of block_start.

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
```diff
@@ -328,14 +421,15 @@ ShenandoahCardCluster<RememberedSet>::block_start(const size_t card_index) const
   //    evacuation phase) of young collections. This is never called
   //    during old or global collections.
   // 4. Every allocation under TAMS updates the object start array.
-  NOT_PRODUCT(obj = cast_to_oop(p);)
   assert(oopDesc::is_oop(obj), "Should be an object");
 #define WALK_FORWARD_IN_BLOCK_START false
   while (WALK_FORWARD_IN_BLOCK_START && p + obj->size() < left) {
     p += obj->size();
+    obj = cast_to_oop(p);
```
kdnilsen, line 428: Make this change in "fix block_start"

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
```diff
@@ -1333,12 +1397,40 @@ void ShenandoahVerifier::help_verify_region_rem_set(ShenandoahHeapRegion* r, She
       } else {
         // This object is not live so we don't verify dirty cards contained therein
         HeapWord* tams = ctx->top_at_mark_start(r);
+        HeapWord* orig_obj_addr = obj_addr;
```
kdnilsen, line 1400: Make this change in better-verification-of-remset

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
```diff
@@ -1333,12 +1397,40 @@ void ShenandoahVerifier::help_verify_region_rem_set(ShenandoahHeapRegion* r, She
...
         obj_addr = ctx->get_next_marked_addr(obj_addr, tams);
+        confirm_filled(orig_obj_addr, obj_addr, message);
       }
     }
   }
 }
 
+void ShenandoahVerifier::confirm_filled(HeapWord* start, HeapWord* end, const char *message) {
```
kdnilsen, line 1408: Also part of better-verification-of-remset

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
```diff
@@ -1409,7 +1524,9 @@ void ShenandoahVerifier::verify_rem_set_before_mark() {
           } else {
             // This object is not live so we don't verify dirty cards contained therein
             assert(tams != nullptr, "If object is not live, ctx and tams should be non-null");
+            HeapWord* orig_obj_addr = obj_addr;
```
kdnilsen, line 1527: Part of improve-verification-of-remset

src/hotspot/share/gc/shenandoah/shenandoahScanRemembered.inline.hpp
kdnilsen: Answers to your questions:
1. There can be dirty cards even if no object is allocated within a card's memory span.  This happens if a large reference array is allocated before the card's start address and ends after the card's start address, and an array entry that is within this card's span is overwritten with a pointer to young.
2. I added the assert you suggested, to confirm that the card being probed by block_start() is dirty.  This assert failed many times during tier0 jtreg testing.  I'll tabulate the failures and stack backtraces as soon as the test run finishes.

kdnilsen: There were 15 failures of this assert, all with a similar stack backtrace:

`
#  assert(_rs->is_card_dirty(card_index)) failed: Only ask block_start() if card is dirty

V  [libjvm.so+0x174ac1d]  ShenandoahCardCluster<ShenandoahDirectCardMarkRememberedSet>::block_start(unsigned long) const+0x57d  (shenandoahScanRemembered.inline.hpp:296)
V  [libjvm.so+0x174b230]  void ShenandoahScanRemembered<ShenandoahDirectCardMarkRememberedSet>::process_clusters<ShenandoahConcUpdateRefsClosure>(unsigned long, unsigned long, HeapWor\
dImpl**, ShenandoahConcUpdateRefsClosure*, bool, unsigned int)+0x570  (shenandoahScanRemembered.inline.hpp:690)
V  [libjvm.so+0x174d335]  void ShenandoahUpdateHeapRefsTask<true>::do_work<ShenandoahConcUpdateRefsClosure>(unsigned int)+0xf25  (shenandoahScanRemembered.inline.hpp:889)
V  [libjvm.so+0x174d663]  ShenandoahUpdateHeapRefsTask<true>::work(unsigned int)+0x73  (shenandoahHeap.cpp:2514)
V  [libjvm.so+0x1af77f0]  WorkerThread::run()+0x80  (workerThread.cpp:70)
V  [libjvm.so+0x19b5cea]  Thread::call_run()+0xba  (thread.cpp:221)
V  [libjvm.so+0x150306a]  thread_native_entry(Thread*)+0x12a  (os_linux.cpp:853)
`

kdnilsen: The assert failures were apparently due to confusion over is_card_dirty() vs is_write_card_dirty().  Resolving this issue, no more assertion failures in tier0 testing.  Going to test this version of code more broadly.

PR #420, opened 2024-04-05
8329790: GenShen: Old regions that are pinned during final mark are not made parseable

earthling-amzn:
Pinned regions that are not humongous may be added to the set of _candidates_ for mixed collection. They still may not be added to the collection set for a mixed collection. Regions that are not in this collection of _candidates_ will not be coalesced and filled.

---------
### Progress

### Issue
 * [JDK-8329790](https://bugs.openjdk.org/browse/JDK-8329790): GenShen: Old regions that are pinned during final mark are not made parseable (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/420/head:pull/420` \
`$ git checkout pull/420`

`$ git checkout pull/420` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/420/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 420`

`$ git pr show -t 420`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/420.diff">https://git.openjdk.org/shenandoah/pull/420.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/420#issuecomment-2040694766)


src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -436,7 +440,8 @@ void ShenandoahOldHeuristics::prepare_for_old_collections() {
     while ((defrag_count < bound_on_additional_regions) &&
            (total_uncollected_old_regions < 15 * span_of_uncollected_regions / 16)) {
       ShenandoahHeapRegion* r = candidates[_last_old_collection_candidate]._region;
-      assert (r->is_regular(), "Only regular regions are in the candidate set");
+      assert((r->is_regular() || r->is_pinned()) && !r->is_humongous(), "Region " SIZE_FORMAT " has wrong state for collection: %s",
```
kdnilsen, line 443: I think assert should be r->is_regular(). (testing for r->is_pinned() and !r->is_humongous() seems redundant and unnecessary).  Or explain the rationale for assert as written with comment.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -310,37 +309,29 @@ void ShenandoahOldHeuristics::prepare_for_old_collections() {
...
-        region->make_trash_immediate();
-        immediate_regions++;
-        immediate_garbage += garbage;
-      } else {
-        region->begin_preemptible_coalesce_and_fill();
-        candidates[cand_idx]._region = region;
-        candidates[cand_idx]._u._live_data = live_bytes;
-        cand_idx++;
-      }
+    if (region->is_trash()) {
```
kdnilsen, line 327: In revised code structure, we handle trash regular regions with same code that handles trash humongous regions.  I like this improvement.  This is a note to self.  Please let me know if I am misunderstaning.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -310,37 +309,29 @@ void ShenandoahOldHeuristics::prepare_for_old_collections() {
...
+      // Count humongous objects made into trash here.
+      immediate_regions++;
+      immediate_garbage += garbage;
     } else if (region->is_humongous_start()) {
+      // This will handle humongous start regions whether they are also pinned, or not.
```
kdnilsen, line 332: Since we've already handled region->is_trash() above, do we need this code?  If we want to preserve the assert(!is_pinned), we can insert that into the is_trash() code above.  Right?

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
kdnilsen: Need to double-check myself here.  is_trash() may not be exactly the same as !region->has_live().  If these are distinct, maybe we want to check for either condition right here.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -352,10 +343,22 @@ void ShenandoahOldHeuristics::prepare_for_old_collections() {
         size_t region_count = heap->trash_humongous_region_at(region);
         log_debug(gc)("Trashed " SIZE_FORMAT " regions for humongous object.", region_count);
       }
-    } else if (region->is_trash()) {
-      // Count humongous objects made into trash here.
-      immediate_regions++;
-      immediate_garbage += garbage;
+    } else if (region->is_regular() || region->is_pinned()) {
+      if (!region->has_live()) {
```
kdnilsen, line 347: Here, I'm not sure why we ask "|| region->is_pinned()".  We don't want humongous regions that are pinned.  And we do want regular regions whether or not they are pinned.  Also, I think it better to handle the case of !region->has_live() at line 327 above.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -457,13 +462,12 @@ void ShenandoahOldHeuristics::prepare_for_old_collections() {
...
-               SIZE_FORMAT " regions), Old-Gen Immediate Garbage: " SIZE_FORMAT "%s over " SIZE_FORMAT " regions.",
-               byte_size_in_proper_unit(collectable_garbage), proper_unit_for_byte_size(collectable_garbage),
-               byte_size_in_proper_unit(unfragmented),        proper_unit_for_byte_size(unfragmented),
-               old_candidates, defrag_count,
-               byte_size_in_proper_unit(immediate_garbage),   proper_unit_for_byte_size(immediate_garbage), immediate_regions);
+  log_info(gc)("Old-Gen Collectable Garbage: " PROPERFMT " consolidated with free: " PROPERFMT ", over " SIZE_FORMAT " regions",
+               PROPERFMTARGS(collectable_garbage), PROPERFMTARGS(unfragmented), old_candidates);
+  log_info(gc)("Old-Gen Immediate Garbage: " PROPERFMT " over " SIZE_FORMAT " regions",
+              PROPERFMTARGS(immediate_garbage), immediate_regions);
+  log_info(gc)("Old regions selected for defragmentation: " SIZE_FORMAT, defrag_count);
```
kdnilsen, line 469: Might be good to add a comment here that defrag_count is a subset of old_candidates.  defrag_count represents regions that were placed into the old collection set in order to defragment the memory that we try to "reserve" for humongous allocations.  old_candidates includes defrag_count, plus it also includes old-gen regions that are to be collected because they contain an abundance of garbage.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -310,37 +309,29 @@ void ShenandoahOldHeuristics::prepare_for_old_collections() {
...
       continue;
     }
 
     size_t garbage = region->garbage();
     size_t live_bytes = region->get_live_data_bytes();
-    total_garbage += garbage;
     live_data += live_bytes;
 
-    // Only place regular regions into the candidate set
-    if (region->is_regular()) {
```
kdnilsen, line 331: I understand that the rationale for the PR is to make sure we do not overlook pinned regions.  What I'm not seeing is how the original code is excluding the pinned regions from being inserted into the "old collection set".  To help my review, can you point me to that part of the original code?

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
kdnilsen: I see that we need to explicitly make_trash() after we discover that !region->has_live().  So these are different conditions.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
kdnilsen: No need for this.  I now understand that is_regular() does not include is_pinned_regular().

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
earthling-amzn: I'm adding a `is_regular_pinned` method to `shHeapRegion`.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
earthling-amzn: I agree, it's confusing to have `is_pinned` mask the underlying state of the region.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
earthling-amzn: Added a comment.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
earthling-amzn: I didn't change the implementation of `is_humongous_start`, so it still includes pinned humongous objects.

kdnilsen: My question was misguided.  Ignore it.  (I was confused when I was first reading this PR.)

kdnilsen (approved):

PR #415, opened 2024-04-04
8329699: GenShen: Move promotion logic out of shHeap and shHeapRegion

earthling-amzn:
The code to handle promotions during evacuation has been moved into a new class in its own files: `shGenerationalEvacuationTask`

The methods for managing `plabs` have been moved into `shGenerationalHeap`. `plabs` are now also only created when running the generational mode.

```
-  inline HeapWord* allocate_from_plab(Thread* thread, size_t size, bool is_promotion);
-  HeapWord* allocate_from_plab_slow(Thread* thread, size_t size, bool is_promotion);
-  HeapWord* allocate_new_plab(size_t min_size, size_t word_size, size_t* actual_size);
-  void retire_plab(PLAB* plab);
-  void retire_plab(PLAB* plab, Thread* thread);
```

The code to handle promotion in place has been moved from `shHeapRegion` into `shGenerationalEvacuationTask`:
```
-  void promote_humongous();
-  void promote_in_place();
```

Unused methods related to global coalesce and fill have been deleted from `shHeapRegion`:
```
-  void global_oop_iterate_and_fill_dead(OopIterateClosure* cl);
-  void global_oop_iterate_objects_and_fill_dead(OopIterateClosure* cl);
-  void oop_iterate_humongous(OopIterateClosure* cl);
-  void oop_iterate_humongous(OopIterateClosure* cl, HeapWord* start, size_t words);
```

Additionally, these methods in `shHeap` which just delegated to old or young generation instances have more or less been inlined at the usage site (many `includes` have been fixed):
```
-  ShenandoahOldHeuristics* old_heuristics();
-  ShenandoahYoungHeuristics* young_heuristics();
-
-  bool doing_mixed_evacuations();
-  bool is_old_bitmap_stable() const;
-  bool is_gc_generation_young() const;
```

The boolean `is_promotion` that was passed through all of the allocation calls has been moved into `shAllocRequest` to reduce the upstream delta for `shHeap`.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (1 review required, with at least 1 [Committer](https://openjdk.org/bylaws#committer))

### Issue
 * [JDK-8329699](https://bugs.openjdk.org/browse/JDK-8329699): GenShen: Move promotion logic out of shHeap and shHeapRegion (**Task** - P4)

### Reviewers
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer)

[...]

PR #423, opened 2024-04-12
Better verification of remset

kdnilsen:
We do not need a remembered set at the start of a Global GC because Global GC scans all of old and all of young.  Because it is not necessary, the implementation of System.gc() may jump directly into marking without finishing the evacuation or coalescing and filling of previously identified mixed-evacuation candidates.  This means the remembered set may be in an incoherent state at the start of Global GC.  If this occurs, the verifier should not attempt to verify the remembered set at the start of Global GC.

---------
### Progress
- [ ] Commit message must refer to an issue

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/423/head:pull/423` \
`$ git checkout pull/423`

`$ git checkout pull/423` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/423/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 423`

`$ git pr show -t 423`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/423.diff">https://git.openjdk.org/shenandoah/pull/423.diff</a>

</details>


kdnilsen:
I'll leave this in draft mode until I have finished more extensive testing.

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
```diff
@@ -1354,7 +1382,18 @@ void ShenandoahVerifier::verify_rem_set_before_mark() {
 
   log_debug(gc)("Verifying remembered set at %s mark", _heap->doing_mixed_evacuations()? "mixed": "young");
```
earthling-amzn, line 1383: Might change this log message to something like:
```C++
log_debug(gc)("Verifying remembered set at %s%s mark", _heap->active_generation()->name(), _heap->doing_mixed_evacuations()? " (mixed)": "");
```

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
```diff
@@ -1333,12 +1333,40 @@ void ShenandoahVerifier::help_verify_region_rem_set(ShenandoahHeapRegion* r, She
...
+        HeapWord* orig_obj_addr = obj_addr;
         obj_addr = ctx->get_next_marked_addr(obj_addr, tams);
+        confirm_filled(orig_obj_addr, obj_addr, message);
       }
     }
   }
 }
 
+void ShenandoahVerifier::confirm_filled(HeapWord* start, HeapWord* end, const char *message) {
+  ShenandoahGenerationalHeap* gen_heap = ShenandoahGenerationalHeap::heap();
```
earthling-amzn, line 1345: `ShenandoahGenerationalHeap::heap` asserts that the mode is generational.

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
```diff
@@ -1333,12 +1333,40 @@ void ShenandoahVerifier::help_verify_region_rem_set(ShenandoahHeapRegion* r, She
...
+  assert(gen_heap->mode()->is_generational(), "Filled is for generational");
+  ShenandoahOldGeneration* old_gen = gen_heap->old_generation();
+  ShenandoahOldHeuristics* old_heuristics = (ShenandoahOldHeuristics*) old_gen->heuristics();
```
earthling-amzn, line 1348: Cast shouldn't be necessary here.

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
```diff
@@ -1333,12 +1333,40 @@ void ShenandoahVerifier::help_verify_region_rem_set(ShenandoahHeapRegion* r, She
...
+    oop obj = cast_to_oop(start);
+
+    // This object is not marked.  It should be a fill object, or an object that is about to be filled.  If it is a fill
+    // object, it should be array of int.
+
+    // if there are coalesce-and-fill candidates, skip over this object.  Otherwise, this regions has already been
+    // coalesced and filled, so this should be a fill object.
+    if (check_fill_objects && !obj->is_array()) {
+      log_info(gc)("confirm_filled thinks thinks is_old_bitmap_stable(): %s, old_gen->is_mark_complete(): %s",
+                   gen_heap->is_old_bitmap_stable()? "yes": "no", old_gen->is_mark_complete()? "yes": "no");
```
earthling-amzn, line 1361: Shoot, I just integrated https://github.com/openjdk/shenandoah/pull/415, it replaces `shHeap::is_old_bitmap_stable` with `shOldGen::is_mark_complete`.

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
```diff
@@ -1333,12 +1333,40 @@ void ShenandoahVerifier::help_verify_region_rem_set(ShenandoahHeapRegion* r, She
...
+      ShenandoahAsserts::print_failure(ShenandoahAsserts::_safe_all, obj, start, nullptr, message,
```
earthling-amzn, line 1362: Why not `assert` here?

earthling-amzn (changes requested):
Maybe something for a different PR, but can we now also stop swapping the remembered set for the verifier for global collections? We do this here: https://github.com/openjdk/shenandoah/blob/master/src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp#L622

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
kdnilsen: merged in and resolved conflict.

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
kdnilsen: Thanks.  Removed.

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
kdnilsen: Thanks.  Removed the extra assert.

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
kdnilsen: Thanks.  Changing.

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
kdnilsen: Asserts:print_failure() makes it easier to provide more diagnostic information than would be provided by an assert failures.

PR #427, opened 2024-05-02
8331609: GenShen: Refactor generational mode allocations

earthling-amzn:
* The code for resizing and retrying young allocation requests according to what's available in the young generation has been removed. This code became redundant after the free set was rewritten and began providing the same functionality.
* An unused field and its attendant methods were removed from `ShenandoahThreadLocalData`.
* The code for managing `plab` expenditures and configuring other plab behavior has been moved from `ShenandoahHeap` into `ShenandoahOldGeneration`.

---------
### Progress

### Issue
 * [JDK-8331609](https://bugs.openjdk.org/browse/JDK-8331609): GenShen: Refactor generational mode allocations (**Task** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/427/head:pull/427` \
`$ git checkout pull/427`

`$ git checkout pull/427` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/427/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 427`

`$ git pr show -t 427`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/427.diff">https://git.openjdk.org/shenandoah/pull/427.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/427#issuecomment-2091902382)


src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -1108,189 +1108,47 @@ HeapWord* ShenandoahHeap::allocate_memory(ShenandoahAllocRequest& req) {
...
-  if (result != nullptr) {
-    req.set_actual_size(smaller_req.actual_size());
+  // If TLAB request size is greater than available, allocate() will attempt to downsize request to fit within available
+  // memory.
+  HeapWord* result = _free_set->allocate(req, in_new_region);
+
+  // Record the plab configuration for this result and register the object.
+  if (result != nullptr && req.is_old()) {
+    old_generation()->configure_plab_for_current_thread(req);
+    // TODO: Is this only necessary for 'shared' allocations?
```
kdnilsen, line 1129: I believe this is only necessary for shared allocs.  I'm supportive of making this change as part of PR, though it might be safer to do in separate PR.

(If this is a PLAB allocation, the objects allocated within the PLAB will be independently registered when they are allocated.)

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
```diff
@@ -218,10 +218,65 @@ size_t ShenandoahOldGeneration::unexpend_promoted(size_t decrement) {
...
+  const size_t requested_bytes = req.size() * HeapWordSize;
+  if (can_promote(requested_bytes)) {
+    // The promotion reserve should be able to accommodate this request. The request
+    // might still fail if alignment with the card table increases the size. The request
+    // may also fail if the heap is badly fragmented and the free set cannot find room for it.
+    return true;
+  }
+
+  if (req.type() == ShenandoahAllocRequest::_alloc_plab) {
+    // The promotion reserve is exhausted. Check if we still have room for evacuations.
```
kdnilsen, line 237: We don't KNOW promotion reserve is entirely exhausted, just that what remains in promotion reserve is less than requested bytes.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
```diff
@@ -218,10 +218,65 @@ size_t ShenandoahOldGeneration::unexpend_promoted(size_t decrement) {
...
+    // Note that we cannot really know how much of the plab will be used for evacuations,
+    // so here we only check that some evacuation reserve still exists.
+    return get_evacuation_reserve() > 0;
+  }
+
+  // This is a shared allocation promotion request. However, we do not have room for any
```
kdnilsen, line 243: How do we know this is NOT a shared allocation evacuation request?  Is there a precondition on the method call?

Implementation below looks we're not sure this is a promotion.  We still need to check.

I think what we're saying here is:
  if req.is_promotion();   return false because we already checked above, and found that can_promote(requested_bytes) to be false.
  
  So what if req is not promotion?  wouldn't we want to confirm that get_evacuation_reserve() > requested_bytes?

kdnilsen (changes requested):

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
```diff
@@ -218,10 +218,65 @@ size_t ShenandoahOldGeneration::unexpend_promoted(size_t decrement) {
...
-size_t ShenandoahOldGeneration::get_promoted_expended() {
+size_t ShenandoahOldGeneration::get_promoted_expended() const {
   return Atomic::load(&_promoted_expended);
 }
 
+bool ShenandoahOldGeneration::can_allocate(const ShenandoahAllocRequest &req) const {
+  assert(req.type() != ShenandoahAllocRequest::_alloc_gclab, "GCLAB pertains only to young-gen memory");
+
+  const size_t requested_bytes = req.size() * HeapWordSize;
+  if (can_promote(requested_bytes)) {
```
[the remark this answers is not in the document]
earthling-amzn: Done

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
earthling-amzn: I'm going to test making this change as part of this PR.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
earthling-amzn: Fixed comment.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
earthling-amzn: > How do we know this is NOT a shared allocation evacuation request?
We check if the request type is  `ShenandoahAllocRequest::_alloc_plab` on line 239.

> So what if req is not promotion? wouldn't we want to confirm that get_evacuation_reserve() > requested_bytes?
[The previous version of this code](https://github.com/openjdk/shenandoah/pull/427/files#diff-5ca0a05384b7b2604dd3c9b55d91a7010a42cd0ba246600c844e19821bd6b60bL1178) had a comment for this case (i.e., shared evacuation) that read:
```C
// This is a shared allocation for evacuation. Memory has already been reserved for this purpose.
```
https://github.com/openjdk/shenandoah/pull/427/files#diff-5ca0a05384b7b2604dd3c9b55d91a7010a42cd0ba246600c844e19821bd6b60bL1178

I believe this change preserves the original behavior, but it makes sense to me to also enforce evacuation reserve constraints for `shared` evacuations.

earthling-amzn: I've fixed these comments.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
earthling-amzn: Pipelines look okay with only registering shared objects here.

kdnilsen (approved):
Looks good.  Thanks.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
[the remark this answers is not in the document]
ysramakrishna: Upon reading further and understanding how this is working, scratch the above suggestions.

We should leave things as they are here in terms of using `promotion` in the naming of the methods and the fields.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -1108,189 +1108,47 @@ HeapWord* ShenandoahHeap::allocate_memory(ShenandoahAllocRequest& req) {
...
+      //
+      // Before any objects are allocated, this card's memory range holds no objects.  Note that allocation of object a
+      // wants to set the starts-object, first-start, and last-start attributes of the preceding card region.
+      // Allocation of object b wants to set the starts-object, first-start, and last-start attributes of this card region.
+      // Allocation of object c also wants to set the starts-object, first-start, and last-start attributes of this
+      // card region.
+      //
+      // The thread allocating b and the thread allocating c can "race" in various ways, resulting in confusion, such as
+      // last-start representing object b while first-start represents object c.  This is why we need to require all
+      // register_object() invocations to be "mutually exclusive" with respect to each card's memory range.
```
ysramakrishna, line 1147: One can simply and more succinctly state:

```
  // Register with object start array while holding heap lock to protect against
  // update races between threads allocating to the same card.
```

ysramakrishna (commented):
A few minor comments; overall looks good, but is still a bit difficult to reason about because of having to carry several case analyses simultaneously in these allocation paths.

Was there any performance difference measured in any of the code pipeline performance tests on account of avoiding some of the redundant tests that you removed?

PR #440, opened 2024-05-25
8327000: GenShen: Integrate updated Shenandoah implementation of FreeSet into GenShen

kdnilsen:
The mainline implementation of ShenandoahFreeSet was recently updated.  This PR integrates the upstream changes
into Generational Shenandoah.

---------
### Progress

### Issue
 * [JDK-8327000](https://bugs.openjdk.org/browse/JDK-8327000): GenShen: Integrate updated Shenandoah implementation of FreeSet into GenShen (**Bug** - P4)

### Reviewers
 * [William Kemper](https://openjdk.org/census#wkemper) (@earthling-amzn - Committer) ⚠️ Review applies to [91c3ab07](https://git.openjdk.org/shenandoah/pull/440/files/91c3ab07d71a5936f4b886eae83f1f09c077fb3a)
 * [Y. Srinivas Ramakrishna](https://openjdk.org/census#ysr) (@ysramakrishna - Committer) ⚠️ Review applies to [91c3ab07](https://git.openjdk.org/shenandoah/pull/440/files/91c3ab07d71a5936f4b886eae83f1f09c077fb3a)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/440/head:pull/440` \
`$ git checkout pull/440`

`$ git checkout pull/440` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/440/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 440`

`$ git pr show -t 440`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/440.diff">https://git.openjdk.org/shenandoah/pull/440.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/440#issuecomment-2159147314)


test/hotspot/jtreg/gc/shenandoah/TestAllocIntArrays.java
```diff
@@ -207,9 +207,14 @@ public class TestAllocIntArrays {
     public static void main(String[] args) throws Exception {
         final int min = 0;
         final int max = 384 * 1024;
+        // Each allocated int array is assumed to consume 16 bytes for alignment and header, plus
+        //  an everage of 4 * the average number of elements in the array.
```
earthling-amzn, line 211: s/everage/average

test/hotspot/jtreg/gc/shenandoah/oom/TestThreadFailure.java
```diff
@@ -46,17 +46,25 @@ static class NastyThread extends Thread {
         public void run() {
             List<Object> root = new ArrayList<Object>();
             while (true) {
-                root.add(new Object[SIZE]);
+              root.add(new Object[SIZE]);
```
earthling-amzn, line 49: Accidental? Rest of file uses 4 spaces for indent.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -127,13 +127,18 @@ void ShenandoahHeapRegion::make_young_maybe() {
...
    case _humongous_cont:
+
+
+
      if (affiliation() != YOUNG_GENERATION) {
        ShenandoahHeap* heap = ShenandoahHeap::heap();
        if (heap->mode()->is_generational()) {
          if (is_old()) {
            heap->old_generation()->decrement_affiliated_region_count();
+           heap->old_generation()->decrease_capacity(region_size_bytes());
```
earthling-amzn, line 138: Should capacity be updated every time affiliated region count changes? If yes, then we could have increase/decrease affiliated region count increase/decrease capacity directly. If no, is this something peculiar to full GC?

earthling-amzn (changes requested):
Minor suggestions. I think it will be easier to compare the freeset changes here with upstream (i.e., openjdk:master).

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -34,459 +34,682 @@
...
-  _region_counts[orig_set]--;
-  _region_counts[new_set]++;
+  //  During rebuild:         Mutator => Collector
+  //                          Mutator empty => Collector
+  //                          Mutator empty => OldCollector
+  //  During flip_to_gc:      Mutator empty => Collector
+  //                          Mutator empty => OldCollector
+  // At start of update refs: Collector => Mutator
+  //                          OldCollector Empty => Mutator
+  assert (((available <= _region_size_bytes) &&
```
earthling-amzn, line 344: This is quite hard for my feeble brain to parse. Could we break this into multiple asserts? or break some of the expressions out into functions?

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -714,11 +962,17 @@ HeapWord* ShenandoahFreeSet::try_allocate_in(ShenandoahHeapRegion* r, Shenandoah
...
 
-  try_recycle_trashed(r);
-  if (!r->is_affiliated()) {
-    ShenandoahMarkingContext* const ctx = _heap->complete_marking_context();
+  if (in_new_region) {
+    log_debug(gc)("Using new region (" SIZE_FORMAT ") for %s (" PTR_FORMAT ").",
+                       r->index(), ShenandoahAllocRequest::alloc_type_to_string(req.type()), p2i(&req));
+    assert(!r->is_affiliated(), "New region " SIZE_FORMAT " should be unaffiliated", r->index());
     r->set_affiliation(req.affiliation());
+    ShenandoahMarkingContext* const ctx = _heap->complete_marking_context();
```
earthling-amzn, line 975: Can we move this `ctx` closer to the `asserts` that use it, or perhaps inline the variable in the asserts?

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -896,10 +1177,11 @@ HeapWord* ShenandoahFreeSet::allocate_contiguous(ShenandoahAllocRequest& req) {
   size_t remainder = words_size & ShenandoahHeapRegion::region_size_words_mask();
   ShenandoahMarkingContext* const ctx = _heap->complete_marking_context();
```
earthling-amzn, line 1178: `ctx` looks unused.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -921,53 +1203,27 @@ HeapWord* ShenandoahFreeSet::allocate_contiguous(ShenandoahAllocRequest& req) {
...
-    return r->free();
-  }
-}
-
-bool ShenandoahFreeSet::has_alloc_capacity(ShenandoahHeapRegion *r) const {
-  return alloc_capacity(r) > 0;
-}
-
-void ShenandoahFreeSet::try_recycle_trashed(ShenandoahHeapRegion *r) {
+void ShenandoahFreeSet::try_recycle_trashed(ShenandoahHeapRegion *r, bool is_generational) {
```
earthling-amzn, line 1226: `is_generational` looks unused.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -1634,29 +2054,27 @@ double ShenandoahFreeSet::internal_fragmentation() {
...
-          empty_contig++;
-        } else {
-          empty_contig = 1;
-        }
+  idx_t rightmost = _partitions.rightmost(ShenandoahFreeSetPartitionId::Mutator);
+  for (idx_t index = _partitions.leftmost(ShenandoahFreeSetPartitionId::Mutator); index <= rightmost; ) {
+    assert(_partitions.in_free_set(ShenandoahFreeSetPartitionId::Mutator, index),
+           "Boundaries or find_first_set_bit failed: " SSIZE_FORMAT, index);
+    ShenandoahHeapRegion* r = _heap->get_region(index);
+    if (r->is_empty()) {
```
earthling-amzn, line 2066: This algorithm has changed from upstream. Is that intentional?

earthling-amzn (changes requested):
It looks like the comments for `ShenandoahFreeSet::internal_fragmentation` and `ShenandoahFreeSet::external_fragmentation` have been moved from the header into the implementation file when comparing against upstream (`jdk:master`). Can we keep them in-sync with upstream for the sake of minimizing differences here?

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
kdnilsen: This is a good observation and suggestion.  This doesn't seem to be a consistent behavior.  promote_in_place() adjusts affiliated regions without adjusting capacity, but maybe it should.  I'll explore further.

kdnilsen: Looks like this is something special about full gc.  The call to make_young_maybe() only occurs when !_is_generational().  kind of a peculiar name for this function, given that we're non-generational here.  Still exploring.

kdnilsen: That code should not even be reached.  I'm testing a change here.

kdnilsen: Have replace this code with an assert.  Thanks for catching the inconsistency.

kdnilsen:
> It looks like the comments for `ShenandoahFreeSet::internal_fragmentation` and `ShenandoahFreeSet::external_fragmentation` have been moved from the header into the implementation file when comparing against upstream (`jdk:master`). Can we keep them in-sync with upstream for the sake of minimizing differences here?

Agree.  I'm making this change.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -34,459 +34,682 @@
...
+                                                                   idx_t old_collector_leftmost_empty,
+                                                                   idx_t old_collector_rightmost_empty,
+                                                                   size_t old_collector_region_count, size_t old_collector_used) {
+  _region_counts[int(ShenandoahFreeSetPartitionId::OldCollector)] = old_collector_region_count;
+  _leftmosts[int(ShenandoahFreeSetPartitionId::OldCollector)] = old_collector_leftmost;
+  _rightmosts[int(ShenandoahFreeSetPartitionId::OldCollector)] = old_collector_rightmost;
+  _leftmosts_empty[int(ShenandoahFreeSetPartitionId::OldCollector)] = old_collector_leftmost_empty;
+  _rightmosts_empty[int(ShenandoahFreeSetPartitionId::OldCollector)] = old_collector_rightmost_empty;
+
+  _region_counts[int(ShenandoahFreeSetPartitionId::OldCollector)] = old_collector_region_count;
```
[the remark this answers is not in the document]
kdnilsen: Thanks here also.  Removing.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: Thanks.  Done.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: Removed.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: Thanks.  Removing this unneeded argument.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: That change came in when I merged from master.  I had assumed the change had also been applied upstream.  Is your diff compared to a "recent" upstream?
```
dc184f1099e0 (Kelvin Nilsen    2024-05-16 16:47:09 +0000 2037)     ShenandoahHeapRegion* r = _heap->get_region(index);
dc184f1099e0 (Kelvin Nilsen    2024-05-16 16:47:09 +0000 2038)     if (r->is_empty()) {
dc184f1099e0 (Kelvin Nilsen    2024-05-16 16:47:09 +0000 2039)       free += ShenandoahHeapRegion::region_size_bytes();
dc184f1099e0 (Kelvin Nilsen    2024-05-16 16:47:09 +0000 2040)       if (last_idx + 1 == index) {
dc184f1099e0 (Kelvin Nilsen    2024-05-16 16:47:09 +0000 2041)         empty_contig++;
f37149b1c510 (Aleksey Shipilev 2020-03-23 19:14:01 +0100 2042)       } else {
f37149b1c510 (Aleksey Shipilev 2020-03-23 19:14:01 +0100 2043)         empty_contig = 0;
f37149b1c510 (Aleksey Shipilev 2020-03-23 19:14:01 +0100 2044)       }
f37149b1c510 (Aleksey Shipilev 2020-03-23 19:14:01 +0100 2045)
f37149b1c510 (Aleksey Shipilev 2020-03-23 19:14:01 +0100 2046)       max_contig = MAX2(max_contig, empty_contig);
f37149b1c510 (Aleksey Shipilev 2020-03-23 19:14:01 +0100 2047)       last_idx = index;
f37149b1c510 (Aleksey Shipilev 2020-03-23 19:14:01 +0100 2048)     }
5f22805a50d4 (Kelvin Nilsen    2024-06-13 14:04:30 +0000 2049)     index = _partitions.find_index_of_next_available_region(Shenando\
ahFreeSetPartitionId::Mutator, index + 1);
f37149b1c510 (Aleksey Shipilev 2020-03-23 19:14:01 +0100 2050)   }
```

kdnilsen: I think we may have drifted from openjdk/jdk after we forked our generational shenandoah work?

kdnilsen: I must have messed something up during manual merge conflict resolution.
I'm fixing this code to match upstream.

ysramakrishna (commented):
Here's an initial tranche of comments. More as I complete the two remaining files or so. Apologies for the delay.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -205,18 +364,30 @@ class ShenandoahFreeSet : public CHeapObj<mtGC> {
   // ShenandoahOldCompactionReserve.  In a future planned enhancement, the reserve for old_collector set when the
   // evacuation reserves are unknown, is based in part on anticipated promotion as determined by analysis of live data
   // found during the previous GC pass which is one less than the current tenure age.
-  void rebuild(size_t young_cset_regions, size_t old_cset_regions, bool have_evacuation_reserves = false);
-
-  void move_collector_sets_to_mutator(size_t cset_regions);
-
-  void add_old_collector_free_region(ShenandoahHeapRegion* region);
+  void finish_rebuild(size_t young_cset_regions, size_t old_cset_regions, size_t num_old_regions,
```
ysramakrishna, line 372: For the documentation comments, for the individual public methods themselves, viz. `prepare_to_rebuild`, `finish_rebuild`, `find_regions_with_alloc_capacity` etc., it's easier to have a terser comment that describes the parameters to each call and their side-effect/result.

Then, in the callers of these methods (the implementation of the rebuild) you can place the commentary about how the rebuild is happening by stitching these methods together. Otherwise the header file can be confusing in the form of an imperative narrative.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -205,18 +364,30 @@ class ShenandoahFreeSet : public CHeapObj<mtGC> {
...
+
+  // Move up to cset_regions number of regions from being available to the collector to being available to the mutator.
+  //
+  // Typical usage: At the end of evacuation, when the collector no longer needs the regions that had been reserved
+  // for evacuation, invoke this to make regions available for mutator allocations.
+  //
+  // Note that we plan to replenish the Collector reserve at the end of update refs, at which time all
+  // of the regions recycled from the collection set will be available.  In the very unlikely event that there
+  // are fewer regions in the collection set than remain in the collector's free set, we limit the transfer in order
+  // to assure that the replenished Collector reserves can be sufficiently large.
```
ysramakrishna, line 387: Apropos my previous comment, this part would seem to belong in the caller(s?) of the method.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
```diff
@@ -1063,77 +1342,150 @@ void ShenandoahFreeSet::find_regions_with_alloc_capacity(size_t &young_cset_regi
...
+                _partitions.leftmost(ShenandoahFreeSetPartitionId::Mutator),
+                _partitions.rightmost(ShenandoahFreeSetPartitionId::Mutator),
+                _partitions.leftmost(ShenandoahFreeSetPartitionId::OldCollector),
+                _partitions.rightmost(ShenandoahFreeSetPartitionId::OldCollector));
 }
 
-// Move no more than cset_regions from the existing Collector and OldCollector free sets to the Mutator free set.
-// This is called from outside the heap lock.
-void ShenandoahFreeSet::move_collector_sets_to_mutator(size_t max_xfer_regions) {
+void ShenandoahFreeSet::move_regions_from_collector_to_mutator(size_t max_xfer_regions) {
```
ysramakrishna, line 1425: This method has 3 isomorphic loops. I wonder if it might read better to extract a separate worker method for the loop and invoke it thrice from this method with appropriate parms. Also I'd advise using a local variable to countdown the regions to be transferred rather than modifying the method's input parm.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -230,8 +401,25 @@ class ShenandoahFreeSet : public CHeapObj<mtGC> {
 
   void print_on(outputStream* out) const;
 
+  // This function places all regions that have allocation capacity into the mutator partition, or if the region
+  // is already affiliated with old, into the old collector partition, identifying regions that have no allocation
+  // capacity as NotFree.  Subsequently, we will move some of the mutator regions into the collector and old collector
+  // partitions with the intent of packing old collector memory into the highest (far rightmost) addresses of the heap,
+  // young collector memory into higher address, and mutator memory consuming the lowest addresses of the heap.
```
ysramakrishna, line 448: Is the part of the comment that starts at "Subsequently, we will ..." done as part of this method? If not, then it should go into the caller of this method (after this method is called), rather than stating what the caller will do. I'd keep documentation of these methods very simple: what its parms are and what it does.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -230,8 +401,25 @@ class ShenandoahFreeSet : public CHeapObj<mtGC> {
...
   void find_regions_with_alloc_capacity(size_t &young_cset_regions, size_t &old_cset_regions,
                                         size_t &first_old_region, size_t &last_old_region, size_t &old_region_count);
+
+
+  // Having placed all regions that have allocation capacity into the mutator partition, move some of these regions from
+  // the mutator partition into the collector partition in order to assure that the memory available for allocations within
+  // the collector partition is at least to_reserve, and move others of these regions from the mutator partition into the
+  // old collector partition to assure that old collector partition has at least old_reserve.  Update old_region_count
+  // to represent the total number of regions in the old generation by adding the number of regions moved from the
+  // mutator partition to the old collector partition.
```
ysramakrishna, line 458: How about:

```
Ensure that the Collector has at least `to_reserve`, the OldCollector has at least `old_reserve`. Upon return `old_region_count` holds the number of regions in OldCollector.
```

This makes me wonder about the asymmetry of the return value. Also could it be a plain return value from the method rather than var parameter to the method?

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -230,8 +401,25 @@ class ShenandoahFreeSet : public CHeapObj<mtGC> {
...
+  void reserve_regions(size_t to_reserve, size_t old_reserve, size_t &old_region_count);
+
+
+
   void reserve_regions(size_t young_reserve, size_t old_reserve);
```
ysramakrishna, line 463: Is the old variant without the var parameter still used somewhere?

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -146,48 +254,99 @@ class ShenandoahSetsOfFree {
...
 };
 
+// Publicly, ShenandoahFreeSet represents memory that is available to mutator threads.  The public capacity(), used(),
+// and available() methods represent this public notion of memory that is under control of the mutator.  Separately,
+// ShenandoahFreeSet also represents memory available to garbage collection activities for compaction purposes.
+//
+// The Shenandoah garbage collector evacuates live objects out of specific regions that are identified as members of the
+// collection set (cset).
+//
+// The ShenandoahFreeSet endeavors to congregrate survivor objects (objects that have been evacuated at least once) at the
```
ysramakrishna, line 269: May be "tries to colocate" is simpler than "endeavors to congregate"?

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -146,48 +254,99 @@ class ShenandoahSetsOfFree {
...
+// Mutator and garbage collection activities tend to scramble the content of regions.  Twice, during each GC pass, we rebuild
+// the free set in an effort to restore the efficient segregation of Collector and Mutator regions:
+//
+//  1. At the start of evacuation, we know exactly how much memory is going to be evacuated, and this guides our
+//     sizing of the Collector free set.
+//
+//  2. At the end of GC, we have reclaimed all of the memory that was spanned by the cset.  We rebuild here to make
+//     sure there is enough memory reserved at the high end of memory to hold the objects that might need to be evacuated
+//     during the next GC pass.
+
```
ysramakrishna, line 285: This is a very useful block comment, thanks!

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
```diff
@@ -146,48 +254,99 @@ class ShenandoahSetsOfFree {
...
+  ShenandoahRegionPartitions _partitions;
+  size_t _retired_old_regions;
 
   HeapWord* allocate_aligned_plab(size_t size, ShenandoahAllocRequest& req, ShenandoahHeapRegion* r);
 
-  // Satisfy young-generation or single-generation collector allocation request req by finding memory that matches
-  // affiliation, which either equals req.affiliation or FREE.  We know req.is_young().
-  HeapWord* allocate_with_affiliation(ShenandoahAffiliation affiliation, ShenandoahAllocRequest& req, bool& in_new_region);
+  // Satisfy collector allocation request req by finding memory that matches affiliation from within the
+  // free partition associated with which_partition.
```
ysramakrishna, line 295: ```Returns the address allocated and `in_new_region` indicates if the allocation was satisfied from a new, previously empty region.```

test/hotspot/jtreg/gc/shenandoah/TestAllocIntArrays.java
kdnilsen: Thanks.  Corrected.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -987,8 +987,12 @@ HeapWord* ShenandoahHeap::allocate_memory(ShenandoahAllocRequest& req) {
     // is testing that the GC overhead limit has not been exceeded.
     // This will notify the collector to start a cycle, but will raise
     // an OOME to the mutator if the last Full GCs have not made progress.
-    if (result == nullptr && !req.is_lab_alloc() && get_gc_no_progress_count() > ShenandoahNoProgressThreshold) {
+    // gc_no_progress_count is incremented following each degen or full GC that fails to achieve is_good_progress().
+    // Note that Generational Shenandoah may increment no_progress_count faster than traditional Shenandoah because young
+    // GCs, which may degenerate, typically occur more frequently than single-generation Global GCs.
```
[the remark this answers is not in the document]
kdnilsen: Thanks for asking these questions and digging into the details.  I see in your other comment that it is the intention that we only increment gc_no_progress_count for consecutive full GCs.  However, our implementation seems to have drifted from that intention.  We also increment following a degen that does not make progress.  And then we increment again in the Full GC after we upgrade to full.  Maybe there are better ways to fix this.  I'm investigating.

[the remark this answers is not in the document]
kdnilsen: I'm of the "opinion" that consecutive Full GC failures means no intervening successful concurrent or degenerated collections.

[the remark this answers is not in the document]
kdnilsen: I'm proposing two changes:
1. Do not increment gc_no_progress in failed degens
2. Reset gc_no_progress to zero after successful concurrent gc

I will remove this comment, and I will test this change.  I believe this may allow me to backoff the extreme testing overrides of ShenandoahNoProgressThreshold.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -999,13 +1003,33 @@ HeapWord* ShenandoahHeap::allocate_memory(ShenandoahAllocRequest& req) {
...
+    //           So the second thread experiences OOMError even through another GC would have reclaimed the memory it wanted
+    //           to allocate.
+    //        2. A GLOBAL GC won't necessarily reclaim all garbage.  Following a concurrent Generational GLOBAL GC, we may
+    //           need to perform multiple concurrent mixed evacuations in order to reclaim all of the dead memory identified
+    //           by the GLOBAL GC mark.  However, the first evacuation performed by the GLOBAL GC will normally reclaim
+    //           a significant amount of garbage (as guided by garbage first heuristic).  If this is not enough memory
+    //           to satisfy the pending allocation request, we are in "dire straits", and a fail-fast OOMError is probably
+    //           the better remediation than repeated attempts to allocate following repeated GC cycles.
+
+    while ((result == nullptr) && (original_count == shenandoah_policy()->full_gc_count())) {
```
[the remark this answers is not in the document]
kdnilsen: These are subtle issues.  I'm not sure how best to address.  Will go with consensus.  One other observation:
```
Exception in thread thread_name: java.lang.OutOfMemoryError: GC Overhead limit exceeded
    Cause: The detail message "GC overhead limit exceeded" indicates that the garbage collector is running all the time and Java program is making very slow progress. After a garbage collection, if the Java process is spending more than approximately 98% of its time doing garbage collection and if it is recovering less than 2% of the heap and has been doing so far the last 5 (compile time constant) consecutive garbage collections, then a java.lang.OutOfMemoryError is thrown. This exception is typically thrown because the amount of live data barely fits into the Java heap having little free space for new allocations.
```
from https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/memleaks002.html

This description aligns with our default behavior of ShenandoahNoProgressThreshold default of 5, which would trigger in this PR's current implementation if we do 5 back-to-back unproductive full GCs (regardless of how much CPU time is consumed by GC vs mutator threads).  Unproductive is defined as NOT is_good_progress(), where is_good_progress() (see shenandoahMetrics.cpp) is:

1. free memory is at least 1% of the heap capacity, and
    a. The amount of used memory is decreased by at least the size of one region, or
    b. We have decreased internal fragmentation by at least 1%, or
    c. We have decreased external fragmentation by at least 1%

One problem that I have observed with certain test programs is that this definition of "is_good_progress" is pretty generous.  We can have many back-to-back full gcs that "make progress" according to this generous definition, while a mutator thread repeatedly retries and fails to satisfy its allocation request.

My judgement is that the mutator is much better served by getting a quick OOM exception rather than chugging away in this retry loop.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -127,13 +127,18 @@ void ShenandoahHeapRegion::make_young_maybe() {
    case _cset:
    case _humongous_start:
    case _humongous_cont:
+
+
+
```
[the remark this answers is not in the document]
kdnilsen: Thanks.  Removed.

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
```diff
@@ -426,8 +426,10 @@ class ShenandoahGenerationStatsClosure : public ShenandoahHeapRegionClosure {
 
     size_t generation_capacity = generation->max_capacity();
     guarantee(stats.span() <= generation_capacity,
-              "%s: generation (%s) size spanned by regions (" SIZE_FORMAT ") must not exceed current capacity (" PROPERFMT ")",
-              label, generation->name(), stats.regions(), PROPERFMTARGS(generation_capacity));
+              "%s: generation (%s) size spanned by regions (" SIZE_FORMAT ") * region size: " PROPERFMT
```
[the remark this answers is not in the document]
kdnilsen: Thanks.  Fixed.

test/hotspot/jtreg/gc/shenandoah/oom/TestThreadFailure.java
```diff
@@ -76,9 +84,13 @@ public static void main(String[] args) throws Exception {
...
 
         {
+            // Generational Shenandoah needs a higher value of ShenandaohNoProgressThreshold than traditional
+            // Shenandoah.  This is because Generational Shenandoah may experience many failed young-generation
+            // GC's which degenerate and increment the count of no-progress collections in the same time that a
+            // global collection occurs in traditional Shenandoah.
             ProcessBuilder pb = ProcessTools.createLimitedTestJavaProcessBuilder(
                     "-Xmx32m",
-                    "-XX:+UnlockExperimentalVMOptions", "-XX:ShenandoahNoProgressThreshold=12",
+                    "-XX:+UnlockExperimentalVMOptions", "-XX:ShenandoahNoProgressThreshold=24",
```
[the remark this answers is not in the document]
kdnilsen: I'm testing now a configuration that honors the intent of this comment, to only increment gc_no_progress count if "consecutive Full GCs" fail to make progress.  This is not how it was implemented before, because our implementation has been also incrementing gc_no_progress count if degenerated fails to make progress, and we have not been resetting the count when we experience a productive concurrent GC.  My interpretation of the intent is that a concurrent GC happening between two unproductive Full GCs does not count as two "consecutive" unproductive full GCs.

If this change behaves well, we may be able to remove the ShenandoahNoProgressThreshold override on TestThreadFailure#generational.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.hpp
```diff
@@ -149,8 +149,10 @@ class ShenandoahGeneration : public CHeapObj<mtGC>, public ShenandoahSpaceInfo {
 
   // These methods change the capacity of the region by adding or subtracting the given number of bytes from the current
   // capacity.
-  void increase_capacity(size_t increment);
-  void decrease_capacity(size_t decrement);
+  size_t increase_capacity(size_t increment);
+  size_t decrease_capacity(size_t decrement);
+
+  size_t establish_capacity(size_t byte_size);
```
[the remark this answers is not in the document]
kdnilsen: Thanks.  Fixed.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.hpp
```diff
@@ -149,8 +149,10 @@ class ShenandoahGeneration : public CHeapObj<mtGC>, public ShenandoahSpaceInfo {
 
   // These methods change the capacity of the region by adding or subtracting the given number of bytes from the current
```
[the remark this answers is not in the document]
kdnilsen: Agree.  Fixed.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
```diff
@@ -754,7 +754,7 @@ void ShenandoahGeneration::prepare_regions_and_collection_set(bool concurrent) {
     size_t first_old, last_old, num_old;
     heap->free_set()->prepare_to_rebuild(young_cset_regions, old_cset_regions, first_old, last_old, num_old);
     // Free set construction uses reserve quantities, because they are known to be valid here
-    heap->free_set()->rebuild(young_cset_regions, old_cset_regions, true);
+    heap->free_set()->finish_rebuild(young_cset_regions, old_cset_regions, num_old, true);
```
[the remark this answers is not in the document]
kdnilsen: I'll add a comment.  num_old is the number of regions that currently can be allocated from within the Collector set (excluding the old-gen cset).

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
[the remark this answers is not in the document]
kdnilsen: If we don't reset gc_no_progress to zero after successful concurrent gc, it seems the use of ShenandoahNoProgressThreshold might be kind of "arbitrary".  For example, TestThreadFailure repeatedly drives the heap to exhaustion, and then it cleans up and does it again.  Each time it reaches exhaustion, we escalate to FullGC and throw OOM, which causes the mutator to release its memory.  Having released its memory, we'll run multiple successful concurrent GCs until we once again have exhausted memory, at which time we'll once again escalate to FullGC, which will be unproductive and we'll thrown OOM.

The sixth time though this loop (the sixth consecutive Full GC with many successful intervening concurrent GCs), we'll see that _gc_no_progress_count exceeds ShenandoahNoProgressThreshold and will throw OOM without even trying to do a Full GC.  I think this is not what we want.

src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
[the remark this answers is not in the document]
kdnilsen: Fixed.  Thanks.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
kdnilsen: I've taken a pass at improving descriptions of all arguments to these functions.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
kdnilsen: I'm moving a variation of this comment to the callers.  Thanks.

ysramakrishna (approved):
I think that's most of my comments. Looks mostly good otherwise, so I'm approving, but let me know if you should need a quick re-review for any substantive changes stemming from reviews (with the caveat that I might not get to rereview before Tuesday June 25 on account of PTO).

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.cpp
kdnilsen: I've factored out the loop bodies into two function calls.  Good suggestions.   Thanks.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
kdnilsen: Thanks.  I've adjusted the comments as guided.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
kdnilsen: I've adjusted this comment as suggested.  old_region_count is both input and output, so I'm leaving as is.  This also more closely mirrors the openjdk/jdk implementation.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
kdnilsen: Good catch.  That is no longer relevant.  Removing.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
kdnilsen: Revised this comment.  Thanks.

src/hotspot/share/gc/shenandoah/shenandoahFreeSet.hpp
kdnilsen: Thanks.  Have updated this comment.

test/hotspot/jtreg/gc/shenandoah/oom/TestThreadFailure.java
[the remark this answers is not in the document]
kdnilsen: In general, this new configuration works well and passes all GHA and all CI/CD pipeline tests.  However, I tried removing the ShenandoanNoProgressThreshold override from this test, and it fails.  In one execution that I carefully analyzed, the behavior is:

1. For NastyThread-0 through NastyThread-12, we perform a FullGC (which has good progress) but the good progress is not enough to satisfy the failed allocation request so we throw OOM.

2. With NastyThread-12, we do not fail fast.  GC(127) is concurrent young.  GC(128) through GC(132) are Full GCs, each with Bad Progress, but each yielding enough free memory to satisfy at least one additional allocation by NastyThread-12.

3. GC(133) is a full GC also with bad progress.  This time, the bad progress is not enough to satisfy the pending alloc request (for 4112 bytes), so we thrown OOM.

4. At this point, we have experienced 5 (Default value of ShenandoaohNoProgressThreshold) consecutive full GCs with no progress, so when the main thread attempts to allocate NastyThread-13 after joining with NastyThread-12, it does not even bother to attempt a Full GC.  It just immediately throws OOM.

5. This causes the test to fail, because main is not "supposed" to experience OOM.

Another complication is that the failure doesn't always happen with NastyThread-13.  Sometimes it happens with NastyThread-5.  GC degradation is not cumulative.  Each NastyThread is supposed to start with a clean slate (after a Full GC reclaims all previously allocated memory).

And finally, I have observed that this test will still occasionally fail even with the ShenandoahNoProgressThreshold=24 override.

So I'm puzzling a bit over why GenShen inherently needs a larger value of ShenandoahNoProgressThreshold than traditional Shenandoah in order to pass this test.  I think the explanation is that GenShen introduces more "heap fragmentation" between OLD and YOUNG generations.  Full GC can help sift out these fragments of memory so that "smaller" allocation requests can still succeed, even though Full GC reports "bad progress".

My current thought is to apply one more tweak to the GenShen behavior: If a pending allocation succeeds following a Full GC, I am inclined to count this as "good progress", regardless of what the other metrics think about progress.  That we were able to allocate following Full GC and were not able to allocate before Full GC is the ultimate measure of "good progress".  Will experiment with this.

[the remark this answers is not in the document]
kdnilsen: With this change, I got 100 consecutive successful runs of the TestThreadFailure.java test without having an override on ShenandoahNoProgressThreshold in the GenShen configuration of the test.  I believe this is preferrable to the use of the override.  It makes GenShen behave more similar to the way Shenandoah behaves.

I am now testing on CI/CD pipeline to see if this change introduces any other performance or correctness regressions.

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
```diff
@@ -326,6 +326,7 @@ void ShenandoahControlThread::service_concurrent_normal_cycle(GCCause::Cause cau
   ShenandoahConcurrentGC gc(heap->global_generation(), false);
   if (gc.collect(cause)) {
     // Cycle is complete
+    heap->notify_gc_progress();
```
ysramakrishna, line 329: I am trying to understand this. Is the following a reasonable summary of the underlying reasoning?

Any concurrent cycle that completes without degenerating is considered progress irrespective of whether it reclaimed anything. The idea is that if there were allocations during that concurrent cycle, they will have succeeded because, if not we'd have degenerated and not reached here: "collect()" would have returned "false".

src/hotspot/share/gc/shenandoah/shenandoahDegeneratedGC.cpp
```diff
@@ -304,7 +304,6 @@ void ShenandoahDegenGC::op_degenerated() {
   // Check for futility and fail. There is no reason to do several back-to-back Degenerated cycles,
   // because that probably means the heap is overloaded and/or fragmented.
   if (!metrics.is_good_progress()) {
-    heap->notify_gc_no_progress();
```
ysramakrishna, line 307: Again, let me make sure I understand the reasoning here:

We are already going to cancel gc and degenerate, so the concurrent collect will return failure, and the control thread will not record progress.

But I don't see anything that actually records lack of progress.

I assume then that a degenerated GC in and of itself isn't indicative of lack of progress, and that lack of  progress shouldn't be recorded here, but rather after having worked harder, perhaps in `op_degenerated_futile()`?

It would be good to leave some clarifying comments for when we read this code many months later and have forgotten the context. May be something like:

```
  // We haven't made progress; we'll try harder by canceling GC and upgrading to a full gc
  // to see if we make more progress.
```

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -999,13 +1001,38 @@ HeapWord* ShenandoahHeap::allocate_memory(ShenandoahAllocRequest& req) {
...
+    //           a significant amount of garbage (as guided by garbage first heuristic).  If this is not enough memory
+    //           to satisfy the pending allocation request, we are in "dire straits", and a fail-fast OOMError is probably
+    //           the better remediation than repeated attempts to allocate following repeated GC cycles.
+
+    if (result == nullptr) {
+      while ((result == nullptr) && (original_count == shenandoah_policy()->full_gc_count())) {
+        control_thread()->handle_alloc_failure(req, true);
+        result = allocate_memory_under_lock(req, in_new_region);
+      }
+      if ((result != nullptr) && mode()->is_generational()) {
```
ysramakrishna, line 1032: I am wondering why we are using a `mode()->is_generational()` filter here.

Would this not work in exactly the same way for the non-generational case as well?

If so, I'd extract this fix out more generally and upstream it separately as its own ticket for fixing the OOM'ing behavior for "no progress OOMs".

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
```diff
@@ -126,7 +126,6 @@ void ShenandoahMmuTracker::record_old_marking_increment(bool old_marking_done) {
   // No special processing for old marking
   double now = os::elapsedTime();
   double duration = now - _most_recent_timestamp;
-
```
ysramakrishna, line 129: Is this change just to remove spurious differences from upstream openjdk/jdk repo?

ysramakrishna (approved):
A few questions/comments, but (delta of) changes look good to me.

earthling-amzn:
The other side of the oom issue is test failures and timeouts caused by Shenandoah not giving up soon enough. If I recall correctly, there were three tests that would fail because they did _not_ raise an OOME when expected. Could we also run:
```
make test TEST="runtime/ClassInitErrors/TestOutOfMemoryDuringInit.java compiler/uncommontrap/TestDeoptOOM.java runtime/reflect/ReflectOutOfMemoryError.java" TEST_VM_OPTS="-Xlog:gc:file=/tmp/shen.log -XX:+UseShenandoahGC"
```
and for genshen:
```
make test TEST="runtime/ClassInitErrors/TestOutOfMemoryDuringInit.java compiler/uncommontrap/TestDeoptOOM.java runtime/reflect/ReflectOutOfMemoryError.java" TEST_VM_OPTS="-Xlog:gc:file=/tmp/shen.log -XX:+UseShenandoahGC -XX:ShenandoahGCMode=generational"
```

src/hotspot/share/gc/shenandoah/shenandoahDegeneratedGC.cpp
earthling-amzn: I think that's intentional. We only want to record no-progress after a full GC.

src/hotspot/share/gc/shenandoah/shenandoahControlThread.cpp
kdnilsen: Right.  The idea is that we completed concurrent GC without a failed allocation, and thus concurrent GC was not cancelled...

kdnilsen: I suppose we could be more "careful" about how we characterize progress...

src/hotspot/share/gc/shenandoah/shenandoahDegeneratedGC.cpp
kdnilsen: The description of ShenandoahNoProgressThreshold speaks only in terms of Full GCs.  In the case that degenerated fails to make progress, we will escalate to full GC.  If that Full GC fails to make progress, it will notify_gc_no_progress().  It's really part of the same GC cycle.  I don't think it's appropriate to count no progress twice for this situation.  If the Full GC does make progress, then it will reset the no-progress counter to zero, and there was no value in incrementing it here...

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
kdnilsen: This would work "equally well" for non-generational mode.  I was reluctant to "change the behavior" of non-gen mode, especially because the problem we're facing seems to only manifest in generational mode.

I don't think it would hurt non-gen mode at all.  But perhaps some rare actual customer of non-gen might actually be counting on the existing behavior...  (?)

src/hotspot/share/gc/shenandoah/shenandoahMmuTracker.cpp
kdnilsen: I had inserted and then removed some debugging instrumentation here.  I will restore the blank lines to "original" form.

kdnilsen:
@earthling-amzn I've tested both command lines with REPEAT_COUNT=10 and all tests pass.  (Had to add -XX:+UnlockExperimentalVMOptions to second command)

earthling-amzn (approved):

earthling-amzn:
Thanks for running additional tests.

PR #455, opened 2024-06-28
8335347: GenShen: Revert change that has adaptive heuristic ignore abbreviated cycles

earthling-amzn:
Additional performance testing has shown no benefit from this feature. In fact, latency is degraded on the extremem benchmark with this feature enabled.

---------
### Progress

### Issue
 * [JDK-8335347](https://bugs.openjdk.org/browse/JDK-8335347): GenShen: Revert change that has adaptive heuristic ignore abbreviated cycles (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/455/head:pull/455` \
`$ git checkout pull/455`

`$ git checkout pull/455` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/455/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 455`

`$ git pr show -t 455`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/455.diff">https://git.openjdk.org/shenandoah/pull/455.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/455#issuecomment-2197272250)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.cpp
@@ -132,8 +132,8 @@ void ShenandoahAdaptiveHeuristics::record_cycle_start() {
   _allocation_rate.allocation_counter_reset();
 }
 
-void ShenandoahAdaptiveHeuristics::record_success_concurrent(bool abbreviated) {
-  ShenandoahHeuristics::record_success_concurrent(abbreviated);
+void ShenandoahAdaptiveHeuristics::record_success_concurrent() {
+  ShenandoahHeuristics::record_success_concurrent();
 
   size_t available = _space_info->available();
 
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.hpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahAdaptiveHeuristics.hpp
@@ -76,7 +76,7 @@ public:
                                                      size_t actual_free);
 
   void record_cycle_start();
-  void record_success_concurrent(bool abbreviated);
+  void record_success_concurrent();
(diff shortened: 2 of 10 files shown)


kdnilsen (approved):
Thanks.

ysramakrishna:
> Additional performance testing has shown no benefit from this feature. In fact, latency is degraded on the extremem benchmark with this feature enabled.

Did we compare degenerated cycle count for more spiky or variable load benchmarks, where the original fix was supposed to help.

I also wonder if instead of the "bang, bang" control (completely ignore abbreviated as was the previous fix, now withdrawn vs treat abbreviated numbers on par with non-abbreviated cycles), I wonder if you might maintain two decaying averages for the two types separately, then toss a coin whose probability is based on the decaying ratio of the two types of events, and use the decaying average value of the corresponding type or the other based on the result of the coin toss. In effect, you are simulating and trying to predict the probability of a long or short cycle, and using the appropriate measure. Not sure how that would behave since the value seems to be continuously calculated, it would probably effectively revert to the same value as the decaying average gives without doing this more complex simulation.  :-)

Anyway, was just thinking out loud here...

earthling-amzn:
We tried every benchmark in the pipeline (dacapo, specjvm2008, specjbb2015, heapothesys, extremem, dilluvian). I think this attempt at optimization didn't work because an application is going to either have many abbreviated cycles, in which case, the heuristic lives in ignorance and runs based on the time for the rare regular length cycle (making it too aggressive). Or, the application may have very few abbreviated cycles, in which case this optimization doesn't matter. I'm sure we could craft a workload that created a sequence of abbreviated cycles, and then caused a non-abbreviated cycle, but this seems uncommon in practice. In those situations, if a degenerated cycle did occur, the late penalties would make the heuristic more aggressive. Your idea is interesting and perhaps something we could experiment with in the fullness of time.

PR #458, opened 2024-07-08
8335932: GenShen: Fix old heuristic unit test

earthling-amzn:
This isn't really a unit test and it must be run with Shenandoah enabled, but it exercises a hard to reach code path. The test has suffered some bit-rot* of late that needs to be addressed.

\* We no longer set soft-max-capacity on the old generation.

---------
### Progress

### Issue
 * [JDK-8335932](https://bugs.openjdk.org/browse/JDK-8335932): GenShen: Fix old heuristic unit test (**Bug** - P4)

### Reviewers
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [4e7e2b38](https://git.openjdk.org/shenandoah/pull/458/files/4e7e2b380b9ae99726daa9c6b70f81152c434ba4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/458/head:pull/458` \
`$ git checkout pull/458`

`$ git checkout pull/458` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/458/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 458`

`$ git pr show -t 458`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/458.diff">https://git.openjdk.org/shenandoah/pull/458.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/458#issuecomment-2215413841)

### DIFF
--- a/test/hotspot/gtest/gc/shenandoah/test_shenandoahOldHeuristic.cpp
+++ b/test/hotspot/gtest/gc/shenandoah/test_shenandoahOldHeuristic.cpp
@@ -82,12 +82,17 @@ class ShenandoahOldHeuristicTest : public ::testing::Test {
     _heap = ShenandoahHeap::heap();
     _heuristics = _heap->old_generation()->heuristics();
     _collection_set = _heap->collection_set();
-    ShenandoahHeapLocker locker(_heap->lock());
+    _heap->lock()->lock(false);
     ShenandoahResetRegions reset;
     _heap->heap_region_iterate(&reset);
-    _heap->old_generation()->set_evacuation_reserve(_heap->old_generation()->soft_max_capacity() / 4);
+    _heap->old_generation()->set_capacity(ShenandoahHeapRegion::region_size_bytes() * 10);
+    _heap->old_generation()->set_evacuation_reserve(ShenandoahHeapRegion::region_size_bytes() * 4);
     _heuristics->abandon_collection_candidates();
     _collection_set->clear();
+  }
+
+  ~ShenandoahOldHeuristicTest() override {
+    _heap->lock()->unlock();
   }
 
   ShenandoahOldGeneration::State old_generation_state() {
@@ -95,7 +100,6 @@ class ShenandoahOldHeuristicTest : public ::testing::Test {
   }
 
   size_t make_garbage(size_t region_idx, size_t garbage_bytes) {
-    ShenandoahHeapLocker locker(_heap->lock());
     ShenandoahHeapRegion* region = _heap->get_region(region_idx);
     region->set_affiliation(OLD_GENERATION);
     region->make_regular_allocation(OLD_GENERATION);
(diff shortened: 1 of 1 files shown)


test/hotspot/gtest/gc/shenandoah/test_shenandoahOldHeuristic.cpp
```diff
@@ -82,20 +82,24 @@ class ShenandoahOldHeuristicTest : public ::testing::Test {
     _heap = ShenandoahHeap::heap();
     _heuristics = _heap->old_generation()->heuristics();
     _collection_set = _heap->collection_set();
-    ShenandoahHeapLocker locker(_heap->lock());
+    _heap->lock()->lock(false);
     ShenandoahResetRegions reset;
     _heap->heap_region_iterate(&reset);
-    _heap->old_generation()->set_evacuation_reserve(_heap->old_generation()->soft_max_capacity() / 4);
+    _heap->old_generation()->set_capacity(ShenandoahHeapRegion::region_size_bytes() * 10);
+    _heap->old_generation()->set_evacuation_reserve(ShenandoahHeapRegion::region_size_bytes() * 4);
```
kdnilsen, line 89: I'll approve these changes because I assume you've tested and are satisfied with the results.

Might be good to leave some comments in this code that warn the user that failure of this test does not necessarily mean we've broken old heuristics.

What causes me some concern is that setting the capacity and reserves of old generation is very "dangerous".  There are "assumptions" scattered throughout the implementation related to reserves being sufficient to handle anticipated evacuation, and furthermore, any attempt to take control over capacity and reserves may be defeated by the GC's ergonomics...

kdnilsen (approved):

test/hotspot/gtest/gc/shenandoah/test_shenandoahOldHeuristic.cpp
kdnilsen: The fourth argument to finish_rebuild (default value false) is a flag that is supposed to be true when evacuation_reserve() is meaningful.  When this flag is true, we also assume that old_evacuation_reserve() and old_promo_reserve() are valid.  If this flag is false, the freeset rebuilder will ignore the value of evacuation_reserve.

I wonder how we make sure that the evacuation reserve we set in this "unit test" program is not ignored.

(In the share-collector-reserves development branch, I get rid of this flag and assume reserves are always valid.)

earthling-amzn: The tests are configuring the old generation with enough capacity and reserves to make sure old regions can be added to the collection set. Indeed, these tests are not true unit tests because they must be run with `-XX:+UseShenandoahGC` and if a _real_ collection happens during test execution, all bets are off. There is a lengthy comment at the top of the file explaining these risks and limitations.

PR #479, opened 2024-08-21
8338534: GenShen: Handle alloc failure differently when immediate garbage is pending

kdnilsen:
Several changes are implemented here:

1. Re-order the phases that execute immediately after final-mark so that we do concurrent-cleanup quicker (but still after concurrent weak references)
2. After immediate garbage has been reclaimed by concurrent cleanup, notify waiting allocators
3. If an allocation failure occurs while immediate garbage recycling is pending, stall the allocation but do not cancel the concurrent gc.

---------
### Progress

### Issue
 * [JDK-8338534](https://bugs.openjdk.org/browse/JDK-8338534): GenShen: Handle alloc failure differently when immediate garbage is pending (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/479/head:pull/479` \
`$ git checkout pull/479`

`$ git checkout pull/479` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/479/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 479`

`$ git pr show -t 479`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/479.diff">https://git.openjdk.org/shenandoah/pull/479.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/479#issuecomment-2321541997)


kdnilsen:
BTW, I can't figure out what jcheck whitespace is complaining about.  I think it is reporting the wrong line number.  jcheck on my branch does not report a whitespace problem.

ysramakrishna:
> BTW, I can't figure out what jcheck whitespace is complaining about. I think it is reporting the wrong line number. jcheck on my local copy does not report a whitespace problem.

Yes, for some reason `git jcheck -s` doesn't find it.

However, there is indeed whitespace (in fact, 2 as in the error message) in that file on an otherwise blank line. However, the line number is indeed incorrect as you stated. It should be line 2467.

```
diff --git a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
index cbcd22c053f..342c023d180 100644
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
@@ -2464,7 +2464,7 @@ void ShenandoahHeap::rebuild_free_set(bool concurrent) {
   _free_set->prepare_to_rebuild(young_cset_regions, old_cset_regions, first_old_region, last_old_region, old_region_count);
   size_t anticipated_immediate_garbage = (old_cset_regions + young_cset_regions) * ShenandoahHeapRegion::region_size_words();
   control_thread()->anticipate_immediate_garbage(anticipated_immediate_garbage);
-  
+
   // If there are no old regions, first_old_region will be greater than last_old_region
   assert((first_old_region > last_old_region) ||
          ((last_old_region + 1 - first_old_region >= old_region_count) &&

```

ysramakrishna:
```
% find . -type f -name "*pp" | xargs grep -n " $" 
./src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp:2467:  

```

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
```diff
@@ -753,6 +753,9 @@ void ShenandoahGeneration::prepare_regions_and_collection_set(bool concurrent) {
     // We are preparing for evacuation.  At this time, we ignore cset region tallies.
     size_t first_old, last_old, num_old;
     heap->free_set()->prepare_to_rebuild(young_cset_regions, old_cset_regions, first_old, last_old, num_old);
+    size_t anticipated_immediate_garbage = (old_cset_regions + young_cset_regions) * ShenandoahHeapRegion::region_size_words();
```
ysramakrishna, line 756: This makes it sound like old_cset_regions & young_cset_regions hold all regions that will be part of `immediate garbage` -- which indeed is the case. The name `prepare_to_rebuild()` does not give one much clue as to the fact that that's happening when we return from the call, and the API spec of the method does not explicitly specify it. One needs to read into the code of the method `find_regions_with_alloc_capacity()` to realize that this is what is happening.

In summary, firstly, I feel some of these methods are in need of tighter header file documentation of post-conditions that callers are relying on and, secondly, I feel given the extremely fat APIs (lots of reference variables that are modified by these methods) that some amount of refactoring is needed in the longer term. The refactoring should be a separate effort, but in the shorter term I think the API/spec documentation of `prepare_to_rebuild` and `find_regions_with_alloc_capacity` could be improved.

src/hotspot/share/gc/shenandoah/shenandoahOldGeneration.cpp
```diff
@@ -474,6 +474,9 @@ void ShenandoahOldGeneration::prepare_regions_and_collection_set(bool concurrent
     size_t cset_young_regions, cset_old_regions;
     size_t first_old, last_old, num_old;
     heap->free_set()->prepare_to_rebuild(cset_young_regions, cset_old_regions, first_old, last_old, num_old);
+    size_t anticipated_immediate_garbage = (cset_young_regions + cset_old_regions) * ShenandoahHeapRegion::region_size_words();
+    heap->control_thread()->anticipate_immediate_garbage(anticipated_immediate_garbage);
```
ysramakrishna, line 478: suggest `set_<foo>` for changing field value `<foo>`.

ysramakrishna (commented):
> Several changes are implemented here:
> 1. Re-order the phases that execute immediately after final-mark so that we do concurrent-cleanup quicker (but still after concurrent weak references)
> 2. After immediate garbage has been reclaimed by concurrent cleanup, notify waiting allocators
> 3. If an allocation failure occurs while immediate garbage recycling is pending, stall the allocation but do not cancel the concurrent gc.

As you suggested offline, I agree that it might make sense to do (1) separately, and then do (2+3).

Left a few comments mainly on (2+3), but I'm still missing the solution to the problem described in the ticket. I'll chat with you offline to get clarity.

src/hotspot/share/gc/shenandoah/shenandoahController.cpp
```diff
@@ -65,11 +69,12 @@ void ShenandoahController::handle_alloc_failure(ShenandoahAllocRequest& req, boo
                  req.type_string(),
                  byte_size_in_proper_unit(req.size() * HeapWordSize), proper_unit_for_byte_size(req.size() * HeapWordSize));
 
-    // Now that alloc failure GC is scheduled, we can abort everything else
-    heap->cancel_gc(GCCause::_allocation_failure);
+    if (Atomic::load(&_anticipated_immediate_garbage) < req.size()) {
```
earthling-amzn, line 72: To make sure I understand... here we are saying that if final mark anticipates this much immediate garbage (computed when it rebuilt the freeset after choosing the collection set), then we aren't going to cancel the GC if this particular request could be satisfied. Instead we will block as though the gc has already been cancelled. This thread will be notified when concurrent cleanup completes.

src/hotspot/share/gc/shenandoah/shenandoahController.cpp
```diff
@@ -92,9 +97,11 @@ void ShenandoahController::handle_alloc_failure_evac(size_t words) {
   heap->cancel_gc(GCCause::_shenandoah_allocation_failure_evac);
 }
 
-void ShenandoahController::notify_alloc_failure_waiters() {
-  _alloc_failure_gc.unset();
-  _humongous_alloc_failure_gc.unset();
+void ShenandoahController::notify_alloc_failure_waiters(bool clear_alloc_failure) {
+  if (clear_alloc_failure) {
```
earthling-amzn, line 101: Why would we not clear the alloc failure? This seems like it would confuse the control thread. Isn't this going to have the control thread attempt to notify alloc failure waiters again when the cycle is finished?

earthling-amzn (changes requested):

ysramakrishna:
@kdnilsen : Should this become a draft for now?

PR #292, opened 2023-06-29
8311602 GenShen: Decouple generational mode heuristics

earthling-amzn:
The general idea here is to straighten out code paths which fork based on the mode configuration. A `ShenandoahYoungHeuristics` class has been added to host the logic specific to the generational mode. In many cases, the entangled generational code can simply be moved behind methods which override existing virtual methods. 

A few other notable changes:
* The confusing "trigger" heuristic has been removed from `ShenandoahOldHeuristics`.
  * The old heuristic defines its own triggers now. It is no longer possible to specify the old generation's trigger heuristic.
* Selection of aged regions for in-place promotion has been moved to `ShenandoahGeneration`.
  * The method implementation does not depend on any members of `ShenandoahHeuristics`. Moving the method therefore removes an unnecessary dependency on the heuristics.
* A new `ShenandoahHeapCharacteristics` interface has been added to provide information about the heap _or generation_ to the heuristics.
  * On the development branch, this interface is implemented by `ShenandoahGeneration`.
  * The plan is to upstream this interface and have it implemented by `ShenandoahHeap` 

The taxonomy of these heuristic classes has become somewhat more complex - here is a chart showing the relationships:

```
ShenandoahHeuristics
-  ShenandoahPassiveHeuristics
-  ShenandoahCompactHeuristics
-  ShenandoahAggressiveHeuristics
-  ShenandoahStaticHeuristics
-  ShenandoahOldHeuristics
+  ShenandoahAdaptiveHeuristics
  +  ShenandoahGenerationalHeuristics
    -  ShenandoahYoungHeuristics
    -  ShenandoahGlobalHeuristics       
```

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (1 review required, with at least 1 [Committer](https://openjdk.org/bylaws#committer))

### Issue
 * [JDK-8311602](https://bugs.openjdk.org/browse/JDK-8311602): GenShen: Decouple generational mode heuristics (**Task** - P4)

### Reviewers
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [23578ee3](https://git.openjdk.org/shenandoah/pull/292/files/23578ee3e660a28e19ede481ef36e554f360bde7)
[...]

PR #484, opened 2024-08-22
8338780: GenShen: Fix up some comments

ysramakrishna:
/issue JDK-8338780

In https://github.com/openjdk/jdk/pull/20395 Aleksey provided some suggestions for improving various comments. This ticket gathers together several of those changes. Additionally, there was a resulting simplification in the collection set construction for the global gc case as a result of the review & ensuing discussion.

**Testing:** 
- [x] GHA
- [x] SPECjbb w/GenShen
- [x] jtreg:tier1
- [x] jtreg:hotspot_gc (one test fails with OOM in master as well as current branch; will investigate that separately)
- [x] Codepipeline testing

---------
### Progress

### Issue
 * [JDK-8338780](https://bugs.openjdk.org/browse/JDK-8338780): GenShen: Fix up some comments (**Task** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/484/head:pull/484` \
`$ git checkout pull/484`

`$ git checkout pull/484` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/484/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 484`

`$ git pr show -t 484`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/484.diff">https://git.openjdk.org/shenandoah/pull/484.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/484#issuecomment-2305281088)

### DIFF
--- a/src/hotspot/share/gc/shenandoah/heuristics/shenandoahGlobalHeuristics.cpp
+++ b/src/hotspot/share/gc/shenandoah/heuristics/shenandoahGlobalHeuristics.cpp
@@ -40,29 +40,14 @@ ShenandoahGlobalHeuristics::ShenandoahGlobalHeuristics(ShenandoahGlobalGeneratio
 void ShenandoahGlobalHeuristics::choose_collection_set_from_regiondata(ShenandoahCollectionSet* cset,
                                                                        RegionData* data, size_t size,
                                                                        size_t actual_free) {
-  // The logic for cset selection in adaptive is as follows:
+  // See comments in ShenandoahAdaptiveHeuristics::choose_collection_set_from_regiondata():
+  // we do the same here, but with the following adjustments for generational mode:
   //
-  //   1. We cannot get cset larger than available free space. Otherwise we guarantee OOME
-  //      during evacuation, and thus guarantee full GC. In practice, we also want to let
-  //      application to allocate something. This is why we limit CSet to some fraction of
-  //      available space. In non-overloaded heap, max_cset would contain all plausible candidates
-  //      over garbage threshold.
-  //
-  //   2. We should not get cset too low so that free threshold would not be met right
-  //      after the cycle. Otherwise we get back-to-back cycles for no reason if heap is
-  //      too fragmented. In non-overloaded non-fragmented heap min_garbage would be around zero.
-  //
-  // Therefore, we start by sorting the regions by garbage. Then we unconditionally add the best candidates
-  // before we meet min_garbage. Then we add all candidates that fit with a garbage threshold before
-  // we hit max_cset. When max_cset is hit, we terminate the cset selection. Note that in this scheme,
-  // ShenandoahGarbageThreshold is the soft threshold which would be ignored until min_garbage is hit.
-
-  // In generational mode, the sort order within the data array is not strictly descending amounts of garbage.  In
-  // particular, regions that have reached tenure age will be sorted into this array before younger regions that contain
-  // more garbage.  This represents one of the reasons why we keep looking at regions even after we decide, for example,
-  // to exclude one of the regions because it might require evacuation of too much live data.
-
-
+  // In generational mode, the sort order within the data array is not strictly descending amounts
+  // of garbage. In particular, regions that have reached tenure age will be sorted into this
+  // array before younger regions that typically contain more garbage. This is one reason why,
+  // for example, we continue examining regions even after rejecting a region that has
+  // more live data than we can evacuate.
 
   // Better select garbage-first regions
   QuickSort::sort<RegionData>(data, (int) size, compare_by_garbage);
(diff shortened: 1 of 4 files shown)


src/hotspot/share/gc/shenandoah/heuristics/shenandoahGlobalHeuristics.cpp
```diff
@@ -127,7 +112,8 @@ void ShenandoahGlobalHeuristics::choose_global_collection_set(ShenandoahCollecti
   for (size_t idx = 0; idx < size; idx++) {
     ShenandoahHeapRegion* r = data[idx].get_region();
     if (cset->is_preselected(r->index())) {
-      fatal("There should be no preselected regions during GLOBAL GC");
+      assert(false, "There should be no preselected regions during GLOBAL GC");
```
ysramakrishna, line 115: If this is a true invariant (which it is in testing to date), then it must be the case that the call at line 55 above to `add_preselected_regions_to_collection_set()` doesn't do anything, and indeed returns `0` into `cur_young_garbage` there.

Clearly, I am missing something here. @kdnilsen ?

ysramakrishna: I'll run some testing with another assert at the code at line 55 to check my (mis)understanding.

kdnilsen: You are correct.  In ShenandoahGenerationalHeuristics::choose_collection_set(), we only call prime_collection_set() if generation->is_young().  For global collections, we will not prime the collection set, which means we will not preselect any regions for promotion.

Given this reality, we can remove the call to add_preselected_regions_to_collection_SET() from ShenandoahGlobalHeuristics::choose_collection_set_from_regiondata().  Keeping the assertion in place is how we know it was safe to remove the call to add preselected regions.

We could do some archeology on the code to remind ourselves how we got here.  But maybe not necessary.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahGlobalHeuristics.cpp
```diff
@@ -40,29 +40,14 @@ ShenandoahGlobalHeuristics::ShenandoahGlobalHeuristics(ShenandoahGlobalGeneratio
...
-  // we hit max_cset. When max_cset is hit, we terminate the cset selection. Note that in this scheme,
-  // ShenandoahGarbageThreshold is the soft threshold which would be ignored until min_garbage is hit.
-
-  // In generational mode, the sort order within the data array is not strictly descending amounts of garbage.  In
-  // particular, regions that have reached tenure age will be sorted into this array before younger regions that contain
-  // more garbage.  This represents one of the reasons why we keep looking at regions even after we decide, for example,
-  // to exclude one of the regions because it might require evacuation of too much live data.
-
-
+  // In generational mode, the sort order within the data array is not strictly descending amounts
```
kdnilsen, line 46: Given discussion below, this comment is not true for Global collections.  Since global collection do not preselect aged regions.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahGlobalHeuristics.cpp
kdnilsen: Could lines 114-118 be replaced with: assert(!cset->is_preselected(r->index(), comment)

ysramakrishna: Thanks Kelvin! I'll convert this to a draft while I do a few experiments/investigations etc. along the lines of your suggestions... (Feel free to leave more comments here as necessary though; thanks!)

ysramakrishna: You are right Kelvin. I made the changes you suggested and will run them through all of the tests. Once that is done, I'll reopen this for a quick re-review. Thanks!

PS: (Archeology) The code was duplicated in https://github.com/openjdk/shenandoah/pull/292 and then the `fatal()` was added in https://github.com/openjdk/shenandoah/pull/318.
Interestingly, for the latter, I had a bunch of review feedback (of a general nature of refactoring) that I never quite got around to submitting, it seems (as I discovered when I loaded the PR in my browser; it's probably pretty stale by now).

kdnilsen (approved):

ysramakrishna:
Thanks @kdnilsen!

/integrate

PR #486, opened 2024-08-23
8338881: GenShen: Use explicit third temp register for post barrier

earthling-amzn:
* Use register given as `tmp3` argument instead of "out of the blue" `r3`
* Simplify logic for post barrier call

## Testing
GHA, internal pipelines

---------
### Progress

### Issue
 * [JDK-8338881](https://bugs.openjdk.org/browse/JDK-8338881): GenShen: Use explicit third temp register for post barrier (**Task** - P4)

### Reviewers
 * [Aleksey Shipilev](https://openjdk.org/census#shade) (@shipilev - Committer) ⚠️ Review applies to [fa7dc0c4](https://git.openjdk.org/shenandoah/pull/486/files/fa7dc0c43bcf7ec88b01b74d3016cb6262d59ebe)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/486/head:pull/486` \
`$ git checkout pull/486`

`$ git checkout pull/486` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/486/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 486`

`$ git pr show -t 486`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/486.diff">https://git.openjdk.org/shenandoah/pull/486.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/486#issuecomment-2307490312)

### DIFF
--- a/src/hotspot/cpu/aarch64/gc/shenandoah/shenandoahBarrierSetAssembler_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/gc/shenandoah/shenandoahBarrierSetAssembler_aarch64.cpp
@@ -419,7 +419,7 @@ void ShenandoahBarrierSetAssembler::store_at(MacroAssembler* masm, DecoratorSet 
   BarrierSetAssembler::store_at(masm, decorators, type, Address(tmp3, 0), val, noreg, noreg, noreg);
 
   bool in_heap = (decorators & IN_HEAP) != 0;
-  bool needs_post_barrier = val != noreg && in_heap && ShenandoahCardBarrier;
+  bool needs_post_barrier = (val != noreg) && in_heap && ShenandoahCardBarrier;
   if (needs_post_barrier) {
     store_check(masm, tmp3);
   }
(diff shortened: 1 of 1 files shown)


src/hotspot/cpu/aarch64/gc/shenandoah/shenandoahBarrierSetAssembler_aarch64.cpp
```diff
@@ -416,21 +416,13 @@ void ShenandoahBarrierSetAssembler::store_at(MacroAssembler* masm, DecoratorSet
...
-    }
-    BarrierSetAssembler::store_at(masm, decorators, type, Address(tmp3, 0), val, noreg, noreg, noreg);
-    if (ShenandoahCardBarrier) {
-      store_check(masm, r3);
-    }
-  }
+  BarrierSetAssembler::store_at(masm, decorators, type, Address(tmp3, 0), val, noreg, noreg, noreg);
 
+  bool in_heap = (decorators & IN_HEAP) != 0;
+  bool needs_post_barrier = val != noreg && in_heap && ShenandoahCardBarrier;
```
shipilev, line 422: Consider `(val != noreg)`.

shipilev (approved):
OK, so comprehension check. The deleted "Barrier needs uncompressed oop for region cross check." thing is copied from G1, which needed "original" value oop for the post-barrier to work. The call to `BSA::store_at` would compress `val`, destroying it. This is irrelevant for Shenandoah code, because we only pass the "location" for card table update, not the actual "value" we store. So this "corruption" is irrelevant. Saves some moves then, nice.

earthling-amzn:
Right, Shenandoah doesn't check that the destination for the store now holds a pointer that points into a different region before updating the card table.

PR #492, opened 2024-09-05
8339611: GenShen: Simplify ShenandoahOldHeuristics::trigger_collection_if_fragmented

kdnilsen:
Replace iterative solution with a single mathematical computation, and adjust old-gen defragmentation efforts to be a bit less aggressive.

---------
### Progress

### Issue
 * [JDK-8339611](https://bugs.openjdk.org/browse/JDK-8339611): GenShen: Simplify ShenandoahOldHeuristics::trigger_collection_if_fragmented (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/492/head:pull/492` \
`$ git checkout pull/492`

`$ git checkout pull/492` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/492/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 492`

`$ git pr show -t 492`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/492.diff">https://git.openjdk.org/shenandoah/pull/492.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/492#issuecomment-2332921834)


src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -546,40 +546,34 @@ void ShenandoahOldHeuristics::clear_triggers() {
...
+    size_t old_words_consumed = old_region_count * region_size_words - old_fragmented_available;
+    size_t old_words_spanned = old_region_span * region_size_words;
+    double old_density = ((double) old_words_consumed) / old_words_spanned;
+
+    double old_span_percent = ((double) old_region_span) / allowed_old_gen_span;
+    double old_span_percent_squared = old_span_percent * old_span_percent;
+
+    if ((old_span_percent >= 0.50) && (old_density / old_span_percent_squared < 0.75)) {
+      // We trigger old defragmentation, for example, if:
+      //  old_span_percent is 100% and old_density is below 75.0%, or
```
kdnilsen, line 570: Note that old_span_percent may be greater than 100%.  Can add comment:
  old_span_percent is 110% and old_density is below 90.8%, or
  
At higher existing utilization, there is less benefit to defragmenting, because there are diminishing returns for the larger efforts.  We're just copying nearly full regions around...

kdnilsen: But if we are at 110% old_span_percent, we have violated our intended Humongous Reserves, so I'm thinking we should try to squeeze that last 10% out of old so we don't have to STW on the next humongous allocation request...

earthling-amzn: I have a few broader questions:
* why is `allowed_old_gen_span` defined in terms of humongous reserve?
* why do we want to square `old_span_percent` and divide it into `old_density`? The comments suggest we ought to be comparing `old_density` directly?

kdnilsen: The idea is that services that anticipate a need for "lots of humongous" allocation can specify a non-zero value for ShenandoahGenerationalHumongousReserve.  If this value is non-zero, GenShen endeavors to avoid placing old-gen regions into the first N% of the heap.  The low end of the heap is reserved for humongous regions, leaving the rest of the heap eligible for Old regions.

Each time we rebuild the freeset, the top-most available regions are reserved for old collector. Next to these, we reserve for the collector (young "survivor regions").  The remaining heap is available for mutator allocations, with humongous allocations prioritized to bottom.

There's an "implicit assumption" in this code that the top-most old-gen region is aligned at top of the heap.  I'll change this by defining old_region_span as (num_regions() - first_old_region).  This makes the correlation more clear.  (I'll also add some comments).

The reason for squaring old_span_percent is because we want non-linear triggering of defragmenting old.  We don't really care about old fragmentation if old's span is small.  Doing too much defragmentation of OLD is shown to degrade performance when the defragmentation was really not "necessary".  With linear (not-squared) old_span_percent, we would be triggering almost twice as frequently at lower span values and less aggressively when old_span_percent exceeds its allowed span:
  // Trigger if old_span_percent is: 110% and density is below 82.5%
  //                                                      100%                                    75% (same at this trigger)
  //                                                       90%                                     67.5%
  //                                                       80%                                     60%
  //                                                       70%                                     56.25%
  //                                                       60%                                     45%
  //                                                       50%                                     37.5%                                      

As old span approaches the "target maximum span" it becomes increasingly important to defragment old.  On the other hand, if we totally ignore fragmentation until we have exceeded the limit, we have high probability of experiencing a STW Full GC due to humongous allocation failure.  This is because:

1. Old defragmentation generally requires many GC cycles, first to do OLD marking, and then to do the mixed evacuations.
2. Defragmenting mixed evacuations typically run slower than "traditional" mixed evacuations.  This is because a typical mixed evacuation only evacuates regions that have "abundant garbage".  But a defragmenting mixed evacuation may have to evacuate regions that are highly utilized.

One other reason to not totally ignore defragmentation until it has become "urgent" is because early investment in defragmenting can prevent the more severe forms of fragmentation from manifesting. When large numbers of regions with low utilization have been promoted in place (which s is one of the key causes of old fragmentation), it is easier to consolidate the live memory from multiple sparsely populated old-gen regions while they are still sparsely populated.  Otherwise, if we delay the defragmentation effort, it is likely that many of the sparsely populated regions become high utilized before we defragment, and then we'll have to copy a lot more data around in order to move regions out of the humongous region zone.

There is room for more extensive performance measurements and improved heuristics, I'm sure.

earthling-amzn: Thank you for the detailed explanation. It makes more sense now.

earthling-amzn (approved):

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -546,40 +546,48 @@ void ShenandoahOldHeuristics::clear_triggers() {
   _growth_trigger = false;
 }
 
-void ShenandoahOldHeuristics::trigger_collection_if_fragmented(size_t first_old_region, size_t last_old_region, size_t old_region_count, size_t num_regions) {
+// This triggers old-gen collection if the number of regions "dedicated" to old generation is much larger than
+// is required to represent the memory currently used within the old generation.  This trigger looks specifically
+// at density of the old-gen spanned region.  A different mechanism triggers old-gen GC if the total number of
+// old-gen regions (regardless of how close the regions are to one another) grows beyond an anticipated growth target.
```
ysramakrishna, line 552: I realize this is a private API / work method. Would still be good to place the documentation comment in the .hpp file.

```
// If the old generation is fragmented, set appropriate collection triggering signals.
```

ysramakrishna: In this implementation, I'd then describe how the fragmentation is computed.

ysramakrishna: Also, I'd avoid using "trigger", and instead say "set_trigger", since the actual collection will happen later, as far as I understand.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -546,40 +546,48 @@ void ShenandoahOldHeuristics::clear_triggers() {
...
+void ShenandoahOldHeuristics::trigger_collection_if_fragmented(size_t first_old_region, size_t last_old_region,
```
ysramakrishna, line 553: I'd name this method `set_fragmentation_trigger()`, with the side-effect that the fragmentation measures are recorded for logging, and associated triggers are set which are used later to start an old cycle.

I'd then rename `trigger_maybe()` to `set_triggers()`, and rename `trigger_collection_if_overgrown()` to `set_growth_trigger()`, etc.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -546,40 +546,48 @@ void ShenandoahOldHeuristics::clear_triggers() {
...
+                                                               size_t old_region_count, size_t num_regions) {
   if (ShenandoahGenerationalHumongousReserve > 0) {
-    size_t old_region_span = (first_old_region <= last_old_region)? (last_old_region + 1 - first_old_region): 0;
+    // Our intent is to pack old-gen memory into the highest-numbered regions of the heap.  Count all memory
+    // above first_old_region as the "span" of old generation.
+    size_t old_region_span = (first_old_region <= last_old_region)? (num_regions - first_old_region): 0;
```
ysramakrishna, line 558: Can it ever be the case that `first_old_region > last_old_region` ? When can that be the case?

Then, `old_region_span = 0`....

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -546,40 +546,48 @@ void ShenandoahOldHeuristics::clear_triggers() {
...
-      size_t span_threshold = eighths * allowed_old_gen_span / 8;
-      double density_threshold = (eighths - 2) / 8.0;
-      if ((old_region_span >= span_threshold) && (old_density < density_threshold)) {
-        trigger_old_is_fragmented(old_density, first_old_region, last_old_region);
-        return;
-      }
-      eighths--;
+    size_t old_words_consumed = old_region_count * region_size_words - old_fragmented_available;
+    size_t old_words_spanned = old_region_span * region_size_words;
+    double old_density = ((double) old_words_consumed) / old_words_spanned;
```
ysramakrishna, line 573: ... and if `old_region_span = 0`, then you end up with line 573 causing a divide-by-zero error.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
```diff
@@ -546,40 +546,48 @@ void ShenandoahOldHeuristics::clear_triggers() {
...
+    if ((old_span_percent >= 0.50) && (old_density / old_span_percent_squared < 0.75)) {
+      // We trigger old defragmentation, for example, if:
+      //  old_span_percent is 110% and old_density is below 90.8%, or
+      //  old_span_percent is 100% and old_density is below 75.0%, or
+      //  old_span_percent is  90% and old_density is below 60.8%, or
+      //  old_span_percent is  80% and old_density is below 48.0%, or
+      //  old_span_percent is  70% and old_density is below 36.8%, or
+      //  old_span_percent is  60% and old_density is below 27.0%, or
+      //  old_span_percent is  50% and old_density is below 18.8%.
+      trigger_old_is_fragmented(old_density, first_old_region, last_old_region);
```
ysramakrishna, line 590: This method seems a bit useless, why not inline the assignment here? I realize you may be trying to make this method span fewer lines, but I think that can be easily achieved by just moving the description of its work out of line from the code into a block comment at the start that describes how fragmentation is computed and what the trigger threshold criteria are.

ysramakrishna (commented):
Left a few clean up comments to make it more readable and more easily understandable.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
kdnilsen: It happens if there are no old regions, i.e. old-gen is empty.  I'll add comments.  (first_old_region == num_regions and last_old_region == -1 implies empty)

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
kdnilsen: Thanks.  Adding comment and changing name to set_trigger_if_old_is_fragmented().  Also, renaming trigger_collection_if_overgown() to set_trigger_if_old_is_overgrown().

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
kdnilsen: Thanks.  Renaming trigger_old_has_grown() to set_growth_trigger().  Renaming trigger_old_is_fragmented() to set_fragmentation_trigger().  Ranaming trigger_maybe() to evaluate_triggers().

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
kdnilsen: Good catch.  Thanks.  I'm restructuring the code to only perform this division if old_span_percent > 0.5.  We only compute old_span_percent if ShenandoahGeneratiionalHumongousReserve > 0.

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
kdnilsen: I'm pulled in both directions on this.  I'll go with your approach, removing both set_fragmentation_trigger() and set_growth_trigger() (renames mentioned in my responses to above comments).

src/hotspot/share/gc/shenandoah/heuristics/shenandoahOldHeuristics.cpp
ysramakrishna: Yes, that should fix the issue. Just ping me for a re-review when that is completed. Thanks!

earthling-amzn (approved):

PR #503, opened 2024-09-20
8339616: GenShen: Introduce new state to distinguish promote-in-place phase as distinct from concurrent evacuation

earthling-amzn:
The generational mode for Shenandoah is able to promote regions in place (effectively moving an entire region of live objects, without evacuating any objects). It does this during a normal evacuation phase. However, in some cases, Shenandoah may choose to skip the evacuation phase, _even when there are entire regions to promote_. Prior to this PR, Shenandoah would essentially force itself into an evacuation phase, with nothing to evacuate. Though this was expedient, it cause all manner of knock-on effects and unintended consequences with barriers and asserts and the state of the heap. Here, we have reverted that business and created a path for this special "only doing in place promotions" cycle that does not cause the heap to expect to evacuate objects or find forwarding pointers in them.


src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
```diff
@@ -510,6 +521,26 @@ void ShenandoahConcurrentGC::entry_evacuate() {
...
+  static const char* msg = "Promote in place";
+  ShenandoahConcurrentPhase gc_phase(msg, ShenandoahPhaseTimings::promote_in_place);
+  EventMark em("%s", msg);
+
+  ShenandoahWorkerScope scope(heap->workers(),
+                              ShenandoahWorkerPolicy::calc_workers_for_conc_evac(),
+                              "promote in place");
+
+  // Ultimately, the same closure handles in-place promotions for tenured regions. No objects will be moved.
+  // The threads will still participate in the oom-evac-protocol, although no memory will be allocated for
```
kdnilsen, line 539: Since there is no allocation for promote-in-place promotions, it seems there would be no need for these threads to participate in oom-evac protocol.  (But maybe they still monitor the conditions that would indicate oom-evac just so we can reuse the same code.)

kdnilsen (approved):

src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
earthling-amzn: Yes, I was being lazy for the sake of code reuse. This also only happens when it is an abbreviate cycle _and_ there are regions to promote. I'll see what it looks like to not enter the oom-evac-protocol for this case.

src/hotspot/share/gc/shenandoah/shenandoahGenerationalEvacuationTask.cpp
```diff
@@ -68,16 +69,24 @@ void ShenandoahGenerationalEvacuationTask::work(uint worker_id) {
...
     ShenandoahParallelWorkerSession worker_session(worker_id);
-    ShenandoahEvacOOMScope oom_evac_scope;
     do_work();
   }
 }
 
 void ShenandoahGenerationalEvacuationTask::do_work() {
+  if (_only_promote_regions) {
+    // No allocations will be made, do not enter oom-during-evac protocol.
+    do_evacuations();
```
shipilev, line 82: It really looks to me that promotion-in-place should be in a separate task? Or the name of the task should reflect it also does promotions? Or at least the "real evacuation" paths in `do_evacuations` be asserted with `!_only_promote_regions`?

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
```diff
@@ -1079,7 +1079,7 @@ void ShenandoahVerifier::verify_after_concmark() {
           _verify_cset_none,           // no references to cset anymore
           _verify_liveness_complete,   // liveness data must be complete here
           _verify_regions_disable,     // trash regions not yet recycled
-          _verify_size_exact,          // expect generation and heap sizes to match exactly
+          sizeness,          // expect generation and heap sizes to match exactly
```
shipilev, line 1082: So... the comment is outdated? I suggest to split `ShenandoahVerifier::verify_after_concmark` into two methods, and clearly point out the expected state in comments then.

shipilev (approved):
Looks okay, but a few cosmetics:

src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
```diff
@@ -510,6 +521,23 @@ void ShenandoahConcurrentGC::entry_evacuate() {
   op_evacuate();
 }
 
+void ShenandoahConcurrentGC::entry_promote_in_place() {
+  shenandoah_assert_generational();
+
```
ysramakrishna, line 526: Do you also want to assert that collection set is empty?

ysramakrishna: The other approach might be to _always_ have a separate pre-evac phase that promotes in place before actual evac to follow (even when collection set is non-empty).

ysramakrishna (approved):
Nice clean-up/disentaglement that makes for clearer and more maintainable code. Thank you!

Left a couple of comments/remarks.

src/hotspot/share/gc/shenandoah/shenandoahGenerationalEvacuationTask.cpp
earthling-amzn: I didn't go so far as to split them into separate tasks, but I did factor the different paths into separate methods. Also added an assert that the collection set is empty when we are only promoting regions in place.

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
earthling-amzn: Done.

src/hotspot/share/gc/shenandoah/shenandoahConcurrentGC.cpp
earthling-amzn: Yes, added this assertion in f409f645e1.

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
```diff
@@ -1069,18 +1069,32 @@ void ShenandoahVerifier::verify_before_concmark() {
...
+          _verify_marked_complete_satb_empty, // bitmaps as precise as we can get, except dangling j.l.r.Refs
+          _verify_cset_none,                  // no references to cset anymore
+          _verify_liveness_complete,          // liveness data must be complete here
+          _verify_regions_disable,            // trash regions not yet recycled
+          _verify_size_exact,                 // expect generation and heap sizes to match exactly
+          _verify_gcstate_stable_weakroots    // heap is still stable, weakroots are in progress
+  );
+}
+
+void ShenandoahVerifier::verify_before_region_promotions() {
```
shipilev, line 1086: I think the better name is `::verify_after_concmark_with_promotions`. It is still "after mark" check first and foremost.

shipilev (approved):
One final thing.

src/hotspot/share/gc/shenandoah/shenandoahVerifier.cpp
earthling-amzn: Okay

shipilev (approved):

PR #504, opened 2024-09-26
8341042: GenShen: Reset mark bitmaps for unaffiliated regions when preparing for a cycle

earthling-amzn:
A recent change to reset unaffiliated bitmaps during the concurrent cleanup phase has caused a performance regression. Likely because concurrent cleanup only uses one thread (at this time). This issue is to revert the regression causing change while we work on a preferred solution.

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
@@ -579,14 +579,9 @@ void ShenandoahHeapRegion::recycle() {
 
   set_top(bottom());
   clear_live_data();
-  heap->marking_context()->clear_bitmap(this);
-
   reset_alloc_metadata();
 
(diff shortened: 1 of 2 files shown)


src/hotspot/share/gc/shenandoah/shenandoahHeapRegion.cpp
```diff
@@ -579,14 +579,9 @@ void ShenandoahHeapRegion::recycle() {
 
   set_top(bottom());
   clear_live_data();
-  heap->marking_context()->clear_bitmap(this);
```
earthling-amzn, line 582: This is a merge error on this branch. `shenandoah-jdk21u` does not reset the bitmaps twice like this.

ysramakrishna: yikes :-)

PR #523, opened 2024-10-23
8342848: Shenandoah: Marking bitmap may not be completely cleared in generational mode

pengxiaolong:
TBC

### DIFF
--- a/src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
+++ b/src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
@@ -77,7 +77,7 @@ public:
     ShenandoahHeap* heap = ShenandoahHeap::heap();
     ShenandoahMarkingContext* const ctx = heap->marking_context();
     while (region != nullptr) {
-      bool needs_reset = _generation->contains(region) || !region->is_affiliated();
+      bool needs_reset = !region->is_affiliated() || _generation->contains(region);
       if (needs_reset && heap->is_bitmap_slice_committed(region)) {
         ctx->clear_bitmap(region);
       }


src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
```diff
@@ -77,7 +77,7 @@ class ShenandoahResetBitmapTask : public WorkerTask {
     ShenandoahHeap* heap = ShenandoahHeap::heap();
     ShenandoahMarkingContext* const ctx = heap->marking_context();
     while (region != nullptr) {
-      bool needs_reset = _generation->contains(region) || !region->is_affiliated();
+      bool needs_reset = !region->is_affiliated() || _generation->contains(region);
```
pengxiaolong, line 80: We should really read the affiliation for the whole test, but will have to create a new method to achieve that to keep the code clean and encapsulated.

pengxiaolong:
The test failure should not be caused by this change, spotted the same failure in other open PR:

```
java.lang.RuntimeException: expected testPhantom1 to be cleared
	at gc.shenandoah.TestReferenceRefersToShenandoah.fail(TestReferenceRefersToShenandoah.java:155)
	at gc.shenandoah.TestReferenceRefersToShenandoah.expectCleared(TestReferenceRefersToShenandoah.java:166)
	at gc.shenandoah.TestReferenceRefersToShenandoah.testConcurrentCollection(TestReferenceRefersToShenandoah.java:243)
	at gc.shenandoah.TestReferenceRefersToShenandoah.main(TestReferenceRefersToShenandoah.java:340)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
	at java.base/java.lang.reflect.Method.invoke(Method.java:573)
	at com.sun.javatest.regtest.agent.MainWrapper$MainTask.run(MainWrapper.java:138)
	at java.base/java.lang.Thread.run(Thread.java:1576)

JavaTest Message: Test threw exception: java.lang.RuntimeException: expected testPhantom1 to be cleared
JavaTest Message: shutting down test

` ``

File: src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp (hunk 1 above)
Comment (earthling-amzn) [reply src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp:80]: Hmm, you reckon the affiliation is changing between first testing for `_generation->contains` and then testing for `!region->is_affiliated()`? I suppose this _could_ happen, but the fix feels fragile (not sure we can guarantee the compiler won't reorder the tests). We do have `ShenandoahHeapRegion::affiliation`. What about:
```C
  auto affiliation = region->affiliation();
  bool needs_reset = (affiliation == ShenandoahAffiliation::FREE || _generation->contains(affiliation))
```
Where `ShenandoahGeneration::contains(ShenandoaAffiliation affiliation)` tests if the affiliation and the generation type match?

earthling-amzn (changes requested):

earthling-amzn:
Yes, we have a ticket for this test failure: https://bugs.openjdk.org/browse/JDK-8342734.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.cpp
pengxiaolong: I didn't notice we have `ShenandoahGeneration::contains(ShenandoaAffiliation affiliation) `, I'll update the PR, thanks.

pengxiaolong: I have added `ShenandoahGeneration::contains(ShenandoaAffiliation affiliation)` and updated the PR.

src/hotspot/share/gc/shenandoah/shenandoahGeneration.hpp
```diff
@@ -26,6 +26,7 @@
 #define SHARE_VM_GC_SHENANDOAH_SHENANDOAHGENERATION_HPP
 
 #include "memory/allocation.hpp"
+#include "gc/shenandoah/shenandoahAffiliation.hpp"
```
earthling-amzn, line 29: Nit, but `shenandoah/shenandoahAffiliation.hpp` should come between `heuristics/shenandoahSpaceInfo.hpp` and `shenandoah/shenandoahGenerationType.hpp`.

earthling-amzn (approved):
Okay with me to /integrate after fixing include order in `shenandoahGeneration.hpp`

src/hotspot/share/gc/shenandoah/shenandoahGeneration.hpp
pengxiaolong: Fixed, thanks.

PR #549, opened 2024-11-21
8344779: GenShen: Consolidate shared heap region closures into one header

earthling-amzn:
Move shared heap region closure implementations into one header, move non-shared implementations closer to their only usage.


kdnilsen (approved):

src/hotspot/share/gc/shenandoah/shenandoahHeapRegionClosures.cpp
```diff
@@ -1,5 +1,4 @@
 /*
- * Copyright (c) 2021, Red Hat, Inc. All rights reserved.
```
ysramakrishna, line 2: This had me confused, but I verified that the deletion of the Copyright line is indeed correct based on what I can tell from `git annotate`.

Also the way the rename shows up in the diffs here is pretty subtle.

ysramakrishna (approved):
LGTM

src/hotspot/share/gc/shenandoah/shenandoahHeapRegionClosures.cpp
earthling-amzn: Yes, git has a mind of its own when it comes to renames.

PR #550, opened 2024-11-21
8344797: GenShen: Update and rename confusing method

earthling-amzn:
A few riders here:
* [Remove vestigial debugging assertions](https://github.com/openjdk/shenandoah/commit/9062113458bd1909409967abb743e11982f351ab)
* [Log cancellation time at debug level](https://github.com/openjdk/shenandoah/commit/e3dd23efcb179bc61565d25f6b46f3690a50caa4)
* [Use try_set instead of set](https://github.com/openjdk/shenandoah/pull/550/commits/49cb66a0fa7d0d660f5b986eea9d71035c157310)

---------
### Progress

### Issue
 * [JDK-8344797](https://bugs.openjdk.org/browse/JDK-8344797): GenShen: Update and rename confusing method (**Task** - P4)

### Reviewers
 * [Y. Srinivas Ramakrishna](https://openjdk.org/census#ysr) (@ysramakrishna - Committer) ⚠️ Review applies to [49cb66a0](https://git.openjdk.org/shenandoah/pull/550/files/49cb66a0fa7d0d660f5b986eea9d71035c157310)
 * [Kelvin Nilsen](https://openjdk.org/census#kdnilsen) (@kdnilsen - Committer) ⚠️ Review applies to [49cb66a0](https://git.openjdk.org/shenandoah/pull/550/files/49cb66a0fa7d0d660f5b986eea9d71035c157310)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/shenandoah.git pull/550/head:pull/550` \
`$ git checkout pull/550`

`$ git checkout pull/550` \
`$ git pull https://git.openjdk.org/shenandoah.git pull/550/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 550`

`$ git pr show -t 550`

</details>

<a href="https://git.openjdk.org/shenandoah/pull/550.diff">https://git.openjdk.org/shenandoah/pull/550.diff</a>

</details>
<details><summary>Using Webrev</summary>

[Link to Webrev Comment](https://git.openjdk.org/shenandoah/pull/550#issuecomment-2492415525)
</details>


src/hotspot/share/gc/shenandoah/shenandoahMarkingContext.cpp
```diff
@@ -41,24 +41,26 @@ bool ShenandoahMarkingContext::is_bitmap_clear() const {
...
     ShenandoahHeap* heap = ShenandoahHeap::heap();
     size_t start_idx = heap->heap_region_index_containing(start);
+#ifdef ASSERT
     size_t end_idx = heap->heap_region_index_containing(end - 1);
-    while (start_idx <= end_idx) {
-      ShenandoahHeapRegion* r = heap->get_region(start_idx);
-      if (!heap->is_bitmap_slice_committed(r)) {
-        return true;
-      }
-      start_idx++;
```
ysramakrishna, line 61: Thanks for fixing this and emphasizing that the test is ever only done within a single region.

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
```diff
@@ -271,7 +271,7 @@ inline void ShenandoahHeap::clear_cancelled_gc(bool clear_oom_handler) {
   _cancelled_gc.set(CANCELLABLE);
   if (_cancel_requested_time > 0) {
     double cancel_time = os::elapsedTime() - _cancel_requested_time;
-    log_info(gc)("GC cancellation took %.3fs", cancel_time);
+    log_debug(gc)("GC cancellation took %.3fs", cancel_time);
```
kdnilsen, line 274: may not be significant, but would prefer to avoid os::elapsedTime() invocation if we're not logging debug.

ysramakrishna (approved):
LGTM

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
```diff
@@ -878,7 +878,7 @@ void ShenandoahHeap::notify_heap_changed() {
   // Update monitoring counters when we took a new region. This amortizes the
   // update costs on slow path.
   monitoring_support()->notify_heap_changed();
-  _heap_changed.set();
+  _heap_changed.try_set();
```
kdnilsen, line 881: I'm not sure I understand the significance of this change from .set() to .try_set().

kdnilsen: Would a comment help explain why it is ok for set() to fail here?

kdnilsen (approved):

src/hotspot/share/gc/shenandoah/shenandoahHeap.inline.hpp
kdnilsen: maybe write the argument to log_debug() as os::elapsedTime() - _cancel_requested_time.

earthling-amzn: That's a good catch and an easy change.

src/hotspot/share/gc/shenandoah/shenandoahHeap.cpp
earthling-amzn: It was an unnecessary change. Upstream has `try_set`. We just need one thread to set it. `try_set` avoid modifying the shared variable if it is already set.

